PR #1, opened 2020-08-18
Use relative submodules to make cross project merging seemless

tspyrou:
Updating all submodules to respective openroad/HEADs.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,10 +1,9 @@
 [submodule "tools/TritonRoute"]
 	path = tools/TritonRoute
-	url = https://github.com/The-OpenROAD-Project/TritonRoute.git
-	branch = alpha2
+	url = ../TritonRoute.git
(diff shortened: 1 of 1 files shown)


rovinski:
I don't understand this. You're moving the remote repo to be on the same machine and parallel to OpenROAD-flow?

tspyrou:
Yes, that's right. The other repo's are in the same project parallel to openroad-flow-private. I did not change them being cloned into tools, just the location of where the .git is for cloning.

rovinski:
I'm not really sure on the reasoning here, we should probably just schedule a meeting. Would require several changes to Jenkins to not break.

maliberty:
I don't get it either.

rovinski (approved):
This is actually a relative _url_ and not relative _filepath_. The change is mostly semantic and can allow for easier private-public integration. It is ok by me for now.

PR #3, opened 2020-11-06
Allow multiple whitespace characters after END of lef files

davidmetz:
I'm currently trying to use SRAMs generated by https://github.com/VLSIDA/OpenRAM in OpenROAD.
The lef files OpenRAM generates have multiple spaces after the END of MACROs.
This fix allows them to be detected and merged.


maliberty (approved):

maliberty:
I made this change in our private repo.  It will get pushed out in the next update.

PR #6, opened 2021-02-28
Update mergeLef.py

ibrahimkhairy:
Fixing issue when MACRO name is part of a PIN name


tspyrou:
This PR has already been made on the private remote. No need to make it here too.

PR #8, opened 2021-03-14
sky130hd missing met1/met2 MINENCLOSEDAREA rules

antonblanchard:


antonblanchard:
I'm getting DRC errors when building a sky130hd project. Details can be found at: https://github.com/google/skywater-pdk/issues/285

maliberty:
Is there no equivalent required for sky130hs?

antonblanchard:
> Is there no equivalent required for sky130hs?

It seems like it should. @agorararmard @RTimothyEdwards any reason we only added the `MINENCLOSEDAREA` rules to the hd and hdll libraries?

RTimothyEdwards:
It _is_ required for all of the libraries.  However, without direct access to update the library repositories, we patched them as needed.  We will be updating all the libraries shortly;  getting the repositories in shape to allow pull requests is coming along and is supposed to be ready in a few days.

maliberty:
@antonblanchard would you update this PR to include hs.

antonblanchard:
Thanks @RTimothyEdwards. @maliberty I've made the same changes to the hs library.

maliberty:
@tspyrou merge the changes again for hs

agorararmard:
@maliberty (cc: @antonblanchard ): Please note that the `sky130_fd_sc_hs` didn't have that problem in the source repo. Which implies that the version of the pdk used here is either **_very_** outdated or hand edited to remove the rule? 

@antonblanchard (cc: @RTimothyEdwards): I only created a patch for hd and hdll because the rule already exists in the techlef for all other libraries.

maliberty:
@msaligane please check if we are up to date with the current skywater pdk.

maliberty:
@msaligane is going to PR a general update to the efabless versions of the skywater pdks.  I expect that will cover this issue.

msaligane (approved):
LGTM!
The upstream from google/skywater is not updated it seems. See my comment here: https://github.com/google/skywater-pdk/issues/285#issuecomment-810705537

PR #9, opened 2021-03-14
sky130hd: Use clkbuf for clock tree and buf elsewhere

antonblanchard:


mathursanjiv:
We should still study QoR differences between both. What differences/advantages do you see in regression?

msaligane (approved):
I think this is how it should be by default. So except if we see really bad QoR. I approve this change.

maliberty:
@tspyrou to merge

antonblanchard:
I saw a small timing improvement in my design after this change. I ran a few of the tests in the repo and the results were mixed, worst slack improved a bit with aes and got a bit worse with ibex. The issue with ibex could just be a difference in where setup repair gave up  (ibex has a pretty huge wns by default).

maliberty:
@antonblanchard Ravi reorganized these files.  Would you please resolve the conflict and we can merge.

antonblanchard:
@maliberty sure, will do.

tspyrou:
@antonblanchard and @maliberty I just resolved the conflict. Lets he how the CI runs.

tspyrou:
@antonblanchard sky130 aes and gcd fail with this change.

tspyrou:
@antonblanchard the conflict was trivial and I resolved it.

maliberty:
@antonblanchard this got approved but never merged.  I reran the tests this morning but there are some failures.  When I look at the diffs I see changes unrelated to the clock buffer (like padding/routing layer).  I wonder if there is a mis-merge here.  Would you take a look again and let's try to get this finished.

flow/platforms/sky130hd/config.mk
```diff
@@ -12,6 +12,28 @@ export LIB_FILES = $(PLATFORM_DIR)/lib/sky130_fd_sc_hd__tt_025C_1v80.lib \
...
+# IO Pin fix margin
+export IO_PIN_MARGIN = 70
+
+# Layer to use for parasitics estimations
+export WIRE_RC_LAYER = met3
+
+# KLayout technology file
+export KLAYOUT_TECH_FILE = ./platforms/$(PLATFORM)/$(PLATFORM).lyt
+
+
```
maliberty, line 36: All this seems unrelated

tspyrou:
It looks like its not clear if this one should go forward. I will close for now. @antonblanchard if you want to push this forward perhaps you can merge with master and re-open.

PR #10, opened 2021-03-14
sky130hd: Fix clock tree slew violations

antonblanchard:
I'm seeing clock tree slew violations on the aes test as well as some of
my own modules. Setting CTS_CLUSTER_SIZE and CTS_CLUSTER_DIAMETER to the
default values in OpenROAD fixes this.


maliberty:
Where did these magic values come from and what issue was resolved with them?

flow/platforms/sky130hd/config.mk
```diff
@@ -140,3 +140,6 @@ export FILL_CELLS = sky130_fd_sc_hd__fill_1 sky130_fd_sc_hd__fill_2 sky130_fd_sc
 # resizer repair_long_wires -max_length
 export MAX_WIRE_LENGTH = 21000
 
+# Avoid slew violations on clock tree
+export CTS_CLUSTER_SIZE = 20
+export CTS_CLUSTER_DIAMETER = 50
```
mathursanjiv, line 126: Values in cts.tcl are 30 and 100 respectively. These may be too conservative and increase clock buffer count. What is the extent of slew violation and how many violations are there? May be try larger number for cluster diameter.

antonblanchard: I can certainly try a few different settings. I tested the OpenROAD defaults and stopped when it resolved the slew violations. Is there a reason OpenROAD-flow-scripts and OpenROAD have different default values?

I've been poking around clock tree synthesis because I have some sky130hd designs that are seeing extremely high clock skew:

```
report_clock_skew
--------------------------------------------------------------------------
Clock clk
Latency      CRPR       Skew
_111501_/GATE_N v
  17.95
_111510_/GATE_N ^
   7.80      0.00      10.15
```

@mathursanjiv I notice you have a `cts_skew` branch, wonder if I should test that.

mathursanjiv: The size and diameter values would be technology and design dependent. So its not possible to have "right" defaults.
If you do not give sink_clustering_size and sink_clustering_max_diameter, tool would work with max capacitance constraint as a default. This is technology independent.

cts_skew would be for gated clock tree in case skew is because of additional levels due to clock gating.

tspyrou: Should we proceed with this PR?

maliberty: I'll leave it to @mathursanjiv to decide.

mathursanjiv: I would suggest trying a run with higher diameter, say 80. If there are no violations, use that since right now we may be increasing clock tree buffer count. Else we can accept 50.
Thanks.

tspyrou: OK, I merged a conflict to bring this up to date with master's head.

tspyrou: I mean fixed a conflict and merged the fixed version into Anton's fork.

PR #12, opened 2021-03-26
Add LSOracle as submodule and compile in docker image.

yashton:
Adds compilation of LSOracle and includes a flag USE_LSORACLE to invoke the tool in the synthesis script.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,4 +6,4 @@
 	url = ../OpenROAD.git
 [submodule "tools/LSOracle-Plugin"]
 	path = tools/LSOracle-Plugin
-	url = git@github.com:lnis-uofu/LSOracle-Plugin.git
+	url = https://github.com/lnis-uofu/LSOracle-Plugin.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "tools/OpenROAD"]
 	path = tools/OpenROAD
 	url = ../OpenROAD.git
+[submodule "tools/LSOracle-Plugin"]
+	path = tools/LSOracle-Plugin
+	url = git@github.com:lnis-uofu/LSOracle-Plugin.git
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -112,6 +112,7 @@ SHELL = /bin/bash -o pipefail
 TIME_CMD = /usr/bin/time -f "%Eelapsed %PCPU %MmemKB"
 export NUM_CORES = $(shell grep -c ^processor /proc/cpuinfo)
 OPENROAD_CMD = openroad -no_init -exit
+LSORACLE_PLUGIN = oracle
 
 WRAPPED_LEFS = $(foreach lef,$(notdir $(WRAP_LEFS)),$(OBJECTS_DIR)/lef/$(lef:.lef=_mod.lef))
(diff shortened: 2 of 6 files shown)


maliberty:
@tspyrou I don't have much in the way of review comments but I'll leave it to you to decide if we want this dependency/capability.  @vvbandeira is updating the build structure so likely this PR will need to be updated with those changes.

yashton:
The Jenkins build information isn't available to me, clicking on details takes me to a login page. Let me know when the build structure change is done, I'll update.

vvbandeira:
Looking at the README file for this plugin I think there are better ways to integrate it with the flow repo.

They already provide a docker image on DockerHub, so we don't need to add as submodule or build. See example bellow (maybe the paths are not correct, did not check).
```
COPY --from=lnis/lsoracle-plugin /LSOracle ./tools/build/LSOracle 
```

maliberty:
For a non-docker build we would need to install it as a dependency in that
case.

vvbandeira:
Yes that is true, but from what I saw on the PR although the submodule is added, the plugin is only built if it is using docker, i.e., `./build_openroad.sh --local` will not compile the plugin

yashton:
> Yes that is true, but from what I saw on the PR although the submodule is added, the plugin is only built if it is using docker, i.e., `./build_openroad.sh --local` will not compile the plugin

@vvbandeira looks like I missed adding the local compilation. I will update the script.

flow/Makefile
```diff
@@ -199,7 +200,7 @@ SYNTH_SCRIPT ?= scripts/synth.tcl
 
 $(RESULTS_DIR)/1_1_yosys.v: $(DONT_USE_LIBS) $(WRAPPED_LIBS)
 	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)
-	($(TIME_CMD) yosys -c $(SYNTH_SCRIPT)) 2>&1 | tee $(LOG_DIR)/1_1_yosys.log
+	($(TIME_CMD) yosys -m $(LSORACLE_PLUGIN) -c $(SYNTH_SCRIPT)) 2>&1 | tee $(LOG_DIR)/1_1_yosys.log
```
tspyrou: I think we need an if { [info exists ::env(USE_LSORACLE)] } {} here so that we only run the lsoracle plug in when the switch is set.

tspyrou (changes requested):

flow/Makefile
yashton: Good catch, i will update.

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "tools/OpenROAD"]
 	path = tools/OpenROAD
 	url = ../OpenROAD.git
+[submodule "tools/LSOracle-Plugin"]
+	path = tools/LSOracle-Plugin
+	url = git@github.com:lnis-uofu/LSOracle-Plugin.git
```
vvbandeira: can you change to the https url instead?

maliberty: why is https preferred?

vvbandeira: Jenkins does not like to use ssh when it does not have r/w access to the repo. That is the reason the pipeline checks are failing now.

tspyrou:
I replaced this PR with https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/13 in order to bring it up to date with the latest changes. I fixed a merge conflict in the build script. @yashton fyi.

yashton:
@tspyrou awesome, let me know if there are anything else we need to do.

PR #13, opened 2021-04-02
Lnis uofu lsoracle integration

tspyrou:


vvbandeira:
The build is now failing due to boost installation on our CI machines. lsoracle requires `./b2 install --with-program_options` while we currently install boost with `./b2 install --with-iostreams --with-test`

tspyrou:
@yashton can LSOracle build with our version of boost? Could you modify the cmakefile?

yashton:
LSOracle depends on the program_options boost module. You'll need to update the b2 install to include the `--with-program_options` to install that module in addition to the test and iostreams you currently have.

maliberty:
What is "Lnis uofu"?

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "tools/OpenROAD"]
 	path = tools/OpenROAD
 	url = ../OpenROAD.git
+[submodule "tools/LSOracle-Plugin"]
+	path = tools/LSOracle-Plugin
+	url = https://github.com/lnis-uofu/LSOracle-Plugin.git
```
maliberty, line 9: I see this has yosys as a submodule which means we will have yosys twice.  Can we avoid that?

Dockerfile
```diff
@@ -13,6 +13,9 @@ RUN mkdir -p /OpenROAD-flow
 
 COPY --from=openroad /OpenROAD/build ./tools/build/OpenROAD
 COPY --from=openroad/yosys /build ./tools/build/yosys
+COPY --from=openroad/lsoracle /LSOracle ./tools/build/LSOracle
+COPY --from=openroad/lsoracle /LSOracle/core/test.ini /usr/local/share/lsoracle/
```
maliberty, line 17: I guess this will work in a container but why do we need to install in /usr/local?  That won't be possible in a bare metal build.

yashton: Just following the conventions for linux filesystem.

.gitmodules
yashton: It depends on how you are checking out the submodules. If the command is `git submodule init --recursive`, it will check it out in both places. The plugin just needs to be linked against the yosys header files, and the makefile for the plugin can be configured to point at any yosys include directory.

yashton:
> What is "Lnis uofu"?

Laboratory for Nano-Integrated Systems at the University of Utah.

yashton:
@tspyrou Anything I can do to help keep this moving?

tspyrou:
@yashton if you can get it to compile with our updated version of boost that would be great. I will then work on it promptly.

yashton:
@tspyrou is the compilation failing on your Jenkins CI instance?

tspyrou:
> @tspyrou is the compilation failing on your Jenkins CI instance?

Yes you can see the red x's in the PR. 
I think the need to install in /usr/local is a showstopper. We don't want end users to need root privs.

yashton:
It's not required to install the config file in /usr/local, that's just a default location. 

For your jenkins build, you will need to update the Boost b2 install command to include the `--with-program_options` argument. We are compatible with the boost version you are using, but you are only installing 2 of the boost submodules, test and iostreams. LSOracle uses the program-options submodule, which needs to be installed.

yashton:
@tspyrou Hi, I've updated LSOracle to remove the dependency on an external Boost https://github.com/lnis-uofu/OpenROAD-flow-scripts/tree/lnis-uofu-lsoracle-integration

PR #29, opened 2021-05-20
docker: bump klayout version

vvbandeira:


maliberty (approved):
Have you pushed new images to dockerhub?

vvbandeira:
> Have you pushed new images to dockerhub?

No. We currently cache only dev images for OpenROAD app (which does not depend on klayout). I have an open task on my list to use a cached image on ORFS as well, but did not had the time to make that happen yet.

PR #67, opened 2021-06-30
Add explicit script and configuration variables for LSOracle in synth script

### LINKED ISSUE #66: [Flow] LSoracle integration appears incomplete
**Describe the bug**
When using the LSORACLE, there are 2 issues.
1. lsoarcle in yosys is excepting some arguments, but the command inserted into the synth.tcl script does not include them so the it generates this message:
```
5. Executing LSOracle pass (MIG optimization using LSOracle).
checking arguments
Invalid number of arguments
```
Which does not describe the missing / required arguments what are invalid.

2. When arguments are added to the lsoracle fails with the following message (this issue may belong in the LSoracle repo, but adding it here for now):
```
ERROR: LSOracle: execution of command "lsoracle -f /tmp/yosys-abc-c3g61p/lso.script 2>&1" failed: return code 255
```

Manually running that command seems to fail on:
```
Could not load context file at: /usr/local/share/lsoracle/test.ini
```
This seems to stem from an assumption that the test.ini file has been installed in a very particular location. This appears to be hardcoded in lsoracle and probably needs to be corrected.

**Expected behavior**
After running `make synth USE_LSORACLE=1` I would except lsoracle to have be part of the synthesis.

**Environment (please complete the following information):**
 - OS: REHL7 and Ubuntu 20
 - OpenROAD-flow commit: 61afe66

yashton:
Fixes #66 
Added additional variables to the makefile to specify the lsoracle binary and config file locations, defaulting to relative on the filesystem. Additionally I made the lsoracle recipe explicit in the synth script, like how abc is invoked.

### DIFF
--- a/build_openroad.sh
+++ b/build_openroad.sh
@@ -164,12 +164,13 @@ elif [ "$build_method" == "LOCAL" ]; then
   mkdir -p tools/build/OpenROAD
   (cd tools/build/OpenROAD && cmake ../../OpenROAD && $NICE make -j$PROC)
 
-  mkdir -p tools/build/LSOracle
-  cmake -B tools/build/LSOracle tools/LSOracle -D CMAKE_BUILD_TYPE=RELEASE -D YOSYS_INCLUDE_DIR=$(pwd)/tools/yosys -D YOSYS_PLUGIN=ON
-  $NICE cmake --build tools/build/LSOracle -j$PROC
-  mkdir -p tools/build/yosys/share/yosys/plugins
-  cp tools/LSOracle/core/test.ini tools/build/LSOracle/core/test.ini
-  cp tools/build/LSOracle/yosys-plugin/oracle.so tools/build/yosys/share/yosys/plugins/
+  cmake -B tools/LSOracle/build tools/LSOracle \
+        -D CMAKE_BUILD_TYPE=RELEASE \
+        -D YOSYS_INCLUDE_DIR=$(pwd)/tools/yosys \
+        -D YOSYS_PLUGIN=ON \
+        -D YOSYS_SHARE_DIR=$(pwd)/tools/build/yosys/share \
+        -D CMAKE_INSTALL_PREFIX=$(pwd)/tools/build/LSOracle
+  $NICE cmake --build tools/LSOracle/build -j$PROC -t install
 else
   echo "ERROR: No valid build method found"
   exit 1
--- a/build_openroad.sh
+++ b/build_openroad.sh
@@ -168,7 +168,7 @@ elif [ "$build_method" == "LOCAL" ]; then
         -D CMAKE_BUILD_TYPE=RELEASE \
         -D YOSYS_INCLUDE_DIR=$(pwd)/tools/yosys \
         -D YOSYS_PLUGIN=ON \
-        -D YOSYS_SHARE_DIR=$(pwd)/tools/build/yosys/share \
+        -D YOSYS_SHARE_DIR=$(pwd)/tools/build/yosys/share/yosys \
         -D CMAKE_INSTALL_PREFIX=$(pwd)/tools/build/LSOracle
   $NICE cmake --build tools/LSOracle/build -j$PROC --target install
 else
--- a/build_openroad.sh
+++ b/build_openroad.sh
@@ -168,6 +168,7 @@ elif [ "$build_method" == "LOCAL" ]; then
   cmake -B tools/build/LSOracle tools/LSOracle -D CMAKE_BUILD_TYPE=RELEASE -D YOSYS_INCLUDE_DIR=$(pwd)/tools/yosys -D YOSYS_PLUGIN=ON
   $NICE cmake --build tools/build/LSOracle -j$PROC
   mkdir -p tools/build/yosys/share/yosys/plugins
+  cp tools/LSOracle/core/test.ini tools/build/LSOracle/core/test.ini
   cp tools/build/LSOracle/yosys-plugin/oracle.so tools/build/yosys/share/yosys/plugins/
 else
   echo "ERROR: No valid build method found"
--- a/build_openroad.sh
+++ b/build_openroad.sh
@@ -170,7 +170,7 @@ elif [ "$build_method" == "LOCAL" ]; then
         -D YOSYS_PLUGIN=ON \
         -D YOSYS_SHARE_DIR=$(pwd)/tools/build/yosys/share \
         -D CMAKE_INSTALL_PREFIX=$(pwd)/tools/build/LSOracle
-  $NICE cmake --build tools/LSOracle/build -j$PROC -t install
+  $NICE cmake --build tools/LSOracle/build -j$PROC --target install
 else
   echo "ERROR: No valid build method found"
   exit 1
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -198,9 +198,15 @@ ifeq (, $(strip $(NPROC)))
 endif
 export NUM_CORES := $(NPROC)
 
-LSORACLE_PLUGIN ?= oracle
+export LSORACLE_CMD ?= $(shell command -v lsoracle)
+ifeq ($(LSORACLE_CMD),)
+  LSORACLE_CMD = $(abspath $(FLOW_HOME)/../tools/build/LSOracle/core/lsoracle)
+endif
+
+LSORACLE_PLUGIN ?= $(abspath $(FLOW_HOME)/../tools/build/LSOracle/yosys-plugin/oracle.so)
+export LSORACLE_KAHYPAR_CONFIG ?= $(abspath $(FLOW_HOME)/../tools/build/LSOracle/core/test.ini)
 ifneq ($(USE_LSORACLE),)
-	YOSYS_FLAGS ?= -m $(LSORACLE_PLUGIN)
+  YOSYS_FLAGS ?= -m $(LSORACLE_PLUGIN)
 endif
 
 #-------------------------------------------------------------------------------
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -200,11 +200,11 @@ export NUM_CORES := $(NPROC)
 
 export LSORACLE_CMD ?= $(shell command -v lsoracle)
 ifeq ($(LSORACLE_CMD),)
-  LSORACLE_CMD = $(abspath $(FLOW_HOME)/../tools/build/LSOracle/core/lsoracle)
+  LSORACLE_CMD = $(abspath $(FLOW_HOME)/../tools/build/LSOracle/bin/lsoracle)
 endif
 
-LSORACLE_PLUGIN ?= $(abspath $(FLOW_HOME)/../tools/build/LSOracle/yosys-plugin/oracle.so)
-export LSORACLE_KAHYPAR_CONFIG ?= $(abspath $(FLOW_HOME)/../tools/build/LSOracle/core/test.ini)
+LSORACLE_PLUGIN ?= $(abspath $(FLOW_HOME)/../tools/build/yosys/share/yosys/plugin/oracle.so)
+export LSORACLE_KAHYPAR_CONFIG ?= $(abspath $(FLOW_HOME)/../tools/build/LSOracle/share/lsoracle/test.ini)
 ifneq ($(USE_LSORACLE),)
   YOSYS_FLAGS ?= -m $(LSORACLE_PLUGIN)
 endif
--- a/flow/scripts/synth.tcl
+++ b/flow/scripts/synth.tcl
@@ -67,8 +67,16 @@ if {[info exist ::env(BLOCKS)]} {
 synth  -top $::env(DESIGN_NAME) {*}$::env(SYNTH_ARGS)
 
 if { [info exists ::env(USE_LSORACLE)] } {
+    set lso_script [open $::env(OBJECTS_DIR)/lso.script w]
+    puts $lso_script "ps -a"
+    puts $lso_script "oracle --config $::env(LSORACLE_KAHYPAR_CONFIG)"
+    puts $lso_script "ps -m"
+    puts $lso_script "crit_path_stats"
+    puts $lso_script "ntk_stats"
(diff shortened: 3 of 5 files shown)


flow/scripts/synth.tcl
```diff
@@ -67,8 +67,16 @@ if {[info exist ::env(BLOCKS)]} {
 synth  -top $::env(DESIGN_NAME) {*}$::env(SYNTH_ARGS)
 
 if { [info exists ::env(USE_LSORACLE)] } {
+    set lso_script [open $::env(OBJECTS_DIR)/lso.script w]
+    puts $lso_script "ps -a"
+    puts $lso_script "oracle --config $::env(LSORACLE_KAHYPAR_CONFIG)"
+    puts $lso_script "ps -m"
+    puts $lso_script "crit_path_stats"
+    puts $lso_script "ntk_stats"
+    close $lso_script
```
gadfort, line 76: Running this on GCD I get the following error from Yosys:
```
Number of cells:                521
     $lut                          382
     BUF_X16                         2
     BUF_X32                         2
     BUF_X8                          2
     DFF_X1                         35
     INV_X16                         2
     NAND2_X1                       34
     NAND2_X2                        2
     NAND2_X4                       60
```
It's importing as $lut, which do not get techmapped.

flow/scripts/synth.tcl
```diff
@@ -67,8 +67,16 @@ if {[info exist ::env(BLOCKS)]} {
...
+
     # LSOracle synthesis
-    lsoracle
+    lsoracle -script $::env(OBJECTS_DIR)/lso.script -lso_exe $::env(LSORACLE_CMD)
```
gadfort, line 79: This might require an update to LSoracle, but I get a file "hypergraph.txt" in my directory. The executable shouldn't be generating random files, if this file is not needed, then it should not get printed. If it's needed, it should probably go to the tmp directory and get cleaned up on exit.

flow/Makefile
```diff
@@ -198,9 +198,15 @@ ifeq (, $(strip $(NPROC)))
...
 export NUM_CORES := $(NPROC)
 
-LSORACLE_PLUGIN ?= oracle
+export LSORACLE_CMD ?= $(shell command -v lsoracle)
+ifeq ($(LSORACLE_CMD),)
+  LSORACLE_CMD = $(abspath $(FLOW_HOME)/../tools/build/LSOracle/core/lsoracle)
+endif
+
+LSORACLE_PLUGIN ?= $(abspath $(FLOW_HOME)/../tools/build/LSOracle/yosys-plugin/oracle.so)
+export LSORACLE_KAHYPAR_CONFIG ?= $(abspath $(FLOW_HOME)/../tools/build/LSOracle/core/test.ini)
```
gadfort: Maybe only set these if USE_LSORACLE is set.

gadfort (changes requested):
LSoracle now runs, but doesn't generate any valid result

build_openroad.sh
```diff
@@ -168,6 +168,7 @@ elif [ "$build_method" == "LOCAL" ]; then
   cmake -B tools/build/LSOracle tools/LSOracle -D CMAKE_BUILD_TYPE=RELEASE -D YOSYS_INCLUDE_DIR=$(pwd)/tools/yosys -D YOSYS_PLUGIN=ON
   $NICE cmake --build tools/build/LSOracle -j$PROC
   mkdir -p tools/build/yosys/share/yosys/plugins
+  cp tools/LSOracle/core/test.ini tools/build/LSOracle/core/test.ini
```
rovinski: This is not the right way to fix this issue - if the file is required to run, then it should be fixed in the CMake build/install scripts.

In fact, when I try to run `make install PREFIX=../../build/LSOracle` I find that it is trying to install to the hardcoded path of `/usr/local/bin/lsoracle`. Please fix in the LSOracle repo so we can downstream it to the fork.

rovinski (changes requested):

flow/scripts/synth.tcl
yashton: I'll need to investigate why that is being generated. I've filed an issue https://github.com/lnis-uofu/LSOracle/issues/51

build_openroad.sh
yashton: I have already made the change upstream https://github.com/lnis-uofu/LSOracle/commit/9568a703d4aa017f55655cb863922184726a0605
The turnaround for change up stream/fork update/PR is incredibly cumbersome, so I was hoping this stopgap would resolve the issue more quickly.
As far as installation, the installation path is set using cmake's install directory variables.
```
install(TARGETS lsoracle CONFIGURATIONS RELEASE RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
install(FILES core/test.ini CONFIGURATIONS RELEASE DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/lsoracle)
```
I don't know if PREFIX is what cmake is expecting, I've filed an issue to investigate https://github.com/lnis-uofu/LSOracle/issues/52

flow/scripts/synth.tcl
yashton: I've added an explicit techmap step after the lsoracle call, in case the techmapping on lines 85-101 is not run due to the conditionals.

flow/Makefile
yashton: Right now the vars are only used if that flag is set, but that may not always be the case. My preference is to populate the variables in the main scope rather than relying on them being dynamically populated. This was previously discussed in the initial PR.

build_openroad.sh
gadfort: it's not clear what copying the test.ini file into the build directory accomplishes, other than just patch a problem.  I think the yosys-plugin should be able accept the $(LSORACLE_KAHYPAR_CONFIG) as a flag rather than rewrite the script, since the plugin has other options that are not available to us now, through the plugin.
If the issue here is that lsoracle is not installed correctly, maybe a fix in the build.sh script to install it into a tools/install/LSOracle would be the better solution since cmake should handle copying of the file itself. In lsoracle it looks like it attempts to find this file relative to the executable, so that might also help eliminate the LSORACLE_KAHYPAR_CONFIG.

flow/scripts/synth.tcl
gadfort: that does work, but based on the plugin code, it looks like this should not be needed and it can map directly to the right cells. So maybe something isn't getting setup correctly to cause it to use the builtin library to re-integration.

build_openroad.sh
yashton: I think we are getting a little off track - the config file just needs to be _somewhere_. In the build script, neither LSOracle, nor yosys, nor OpenROAD uses cmake or any other build tool to do the installation to tools/build. Everything is just a copy operation. tools/build/LSOracle was just a simple location to put it.

The lsoracle script I am generating in synth.tcl does exactly what the plugin does for the default script. I had hoped it would be a make it more obvious what the tool was doing, just like how the abc script is done below. There's no significance to the environment variable, the relevant line for calling optimization has a command argument for specifying the config file. All the path strings you see in the lsoracle code base are just defaults, you can always specify as an option to the command. The environment variable is a good idea, but outside the scope of fixing this.

flow/scripts/synth.tcl
srtemp: MAJ3 isn't available in the built-in library, so they're treated as LUTs, necessitating a call to techmap. Adding MAJ as a Yosys primitive is on our wishlist to avoid the extra step.  In the interim, manually catching other types of cell is useful internally if we're running in interactive mode, using custom scripts, etc. It's not very useful in this context.

build_openroad.sh
rovinski: > I have already made the change upstream [lnis-uofu/LSOracle@9568a70](https://github.com/lnis-uofu/LSOracle/commit/9568a703d4aa017f55655cb863922184726a0605)

Well this repository points to a fork of [lnis-uofu/LSOracle-Plugin](https://github.com/lnis-uofu/LSOracle-Plugin), which has LSOracle as a submodule which hasn't been updated with that change. Our fork is up to date.

But in fact, the repository now looks archived saying that it is integrated into the main LSOracle repo... so there should probably be a PR to change that...

> The turnaround for change up stream/fork update/PR is incredibly cumbersome, so I was hoping this stopgap would resolve the issue more quickly.

Not really... the whole thing would probably take less than a day. There are some people on this project that are okay with stopgaps but I abhor them because this is an ongoing open-source project. If it's worth doing, then it's worth doing the right way IMO.

> As far as installation, the installation path is set using cmake's install directory variables.
> 
> ```
> install(TARGETS lsoracle CONFIGURATIONS RELEASE RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
> install(FILES core/test.ini CONFIGURATIONS RELEASE DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/lsoracle)
> ```
> 
> I don't know if PREFIX is what cmake is expecting, I've filed an issue to investigate [lnis-uofu/LSOracle#52](https://github.com/lnis-uofu/LSOracle/issues/52)

I was confusing `DESTDIR` and `PREFIX` (which is what yosys uses because it isn't CMake-based). `DESTDIR` works fine. You can close lnis-uofu/LSOracle#52

rovinski: > In the build script, neither LSOracle, nor yosys, nor OpenROAD uses cmake or any other build tool to do the installation to tools/build. Everything is just a copy operation. tools/build/LSOracle was just a simple location to put it.

This is not true: yosys is installed with `make install`. OpenROAD wasn't installed purely because of legacy cruft but #56 will fix that when I get around to it. We should move to everything being installed.

yashton: > Well this repository points to a fork of [lnis-uofu/LSOracle-Plugin](https://github.com/lnis-uofu/LSOracle-Plugin), which has LSOracle as a submodule which hasn't been updated with that change. Our fork is up to date.
> 
Where are you seeing references to the plugin still? Everything should be completely pointed at LSOracle, the plugin was integrated into that repository specifically during the PR to get it into here in the first place..

rovinski: Not sure, I must have a very stale copy or something. In that case, you can simply update the submodule hash in this PR then. We also have a fork of [lnis-uofu/LSOracle](https://github.com/lnis-uofu/LSOracle) which is also up to date. I am going to delete the fork of LSOracle-Plugin if it is unused.

flow/scripts/synth.tcl
gadfort: Ahh I see, well the techmap step appears to remap the luts, so I guess that will do until yosys is updated.

build_openroad.sh
yashton: I've updated the build to do an install rather than a copy.

rovinski (approved):
I am ok with the changes but make sure @arlpetergadfort approves first.

flow/Makefile
```diff
@@ -198,9 +198,15 @@ ifeq (, $(strip $(NPROC)))
 endif
 export NUM_CORES := $(NPROC)
 
-LSORACLE_PLUGIN ?= oracle
+export LSORACLE_CMD ?= $(shell command -v lsoracle)
+ifeq ($(LSORACLE_CMD),)
+  LSORACLE_CMD = $(abspath $(FLOW_HOME)/../tools/build/LSOracle/bin/lsoracle)
+endif
+
+LSORACLE_PLUGIN ?= $(abspath $(FLOW_HOME)/../tools/build/yosys/share/yosys/plugin/oracle.so)
```
gadfort, line 206: This path should be: `$(FLOW_HOME)/../tools/build/yosys/share/plugin/oracle.so`

gadfort (changes requested):
Looks good, the plugin path just needs to be corrected

flow/Makefile
yashton:
```
% find tools/build/yosys
tools/build/yosys
tools/build/yosys/bin
tools/build/yosys/bin/yosys
tools/build/yosys/bin/yosys-config
tools/build/yosys/bin/yosys-filterlib
tools/build/yosys/bin/yosys-smtbmc
tools/build/yosys/bin/yosys-abc
tools/build/yosys/share
tools/build/yosys/share/yosys
tools/build/yosys/share/yosys/greenpak4
tools/build/yosys/share/yosys/greenpak4/cells_latch.v
...
```
tools/build/yosys/share/yosys is correct, though it looks weird -> [/usr/local]/share/yosys -> [tools/build/yosys]/share/yosys

yashton: it would actually make things easier for installation if we just did tools/build as the prefix, so you had tools/build/bin, tools/build/share, etc, but that's a question for another day.

yashton:
Rerunning the tests, there was some http network error.

flow/Makefile
gadfort: Yes, you are correct, but that is not where the module gets installed:
```
-- Install configuration: "RELEASE"
-- Installing: /OpenROAD-flow-scripts/tools/build/LSOracle/bin/lsoracle
-- Up-to-date: /OpenROAD-flow-scripts/tools/build/LSOracle/share/lsoracle/test.ini
-- Installing: /OpenROAD-flow-scripts/tools/build/yosys/share/plugin/oracle.so
```
Notice the path for the oracle.so, that might be an issue in the LSOracle repo that needs to be corrected.

yashton: My mistake, you were talking about the makefile variable rather than the installation location, which i assumed was correct. I fixed the install location.

gadfort (approved):

flow/Makefile
gadfort: looks good, I was fixated on the Makefile and didn't notice the install location. Thanks for fixing

rovinski: > it would actually make things easier for installation if we just did tools/build as the prefix, so you had tools/build/bin, tools/build/share, etc, but that's a question for another day.

I think the reasoning is that if someone wants to use different versions of the tools then they can do so. e.g. use OpenROAD that is installed with the scripts, but use a system-installed yosys.

PR #68, opened 2021-07-02
Option to enable hierarchical synthesis and run restructure command

mathursanjiv:
if SYNTH_HIER_AREA_RECOVER is set do following:
1. Enable hierarchical synthesis instead of flat
2. Run restructure command in floorplan.tcl


flow/scripts/floorplan.tcl
```diff
@@ -87,6 +87,37 @@ remove_buffers
 
 source $::env(SCRIPTS_DIR)/report_metrics.tcl
 report_metrics "floorplan final"
+set num_instances [llength [get_cells -hier *]]
+puts "number instances before restructure is $num_instances"
+puts "Design Area before restructure"
+report_design_area
```
gadfort: wouldn't it make more sense to put this inside the if statement? It message about restructuring doesn't make sense if that is not enabled.

flow/scripts/floorplan.tcl
```diff
@@ -87,6 +87,37 @@ remove_buffers
...
+  # remove buffers inserted by abc
+  remove_buffers
+
+  write_verilog $::env(RESULTS_DIR)/post_abc.v
+}
+
+set num_instances [llength [get_cells -hier *]]
+puts "number instances after restructure is $num_instances"
+puts "Design Area after restructure"
+report_design_area
```
gadfort: same as above

flow/scripts/floorplan.tcl
```diff
@@ -87,6 +87,37 @@ remove_buffers
...
 source $::env(SCRIPTS_DIR)/report_metrics.tcl
 report_metrics "floorplan final"
+set num_instances [llength [get_cells -hier *]]
+puts "number instances before restructure is $num_instances"
+puts "Design Area before restructure"
+report_design_area
+
+if { [info exist ::env(SYNTH_HIER_AREA_RECOVER)] && $::env(SYNTH_HIER_AREA_RECOVER) == 1 } {
+
+  write_verilog $::env(RESULTS_DIR)/pre_abc.v
```
gadfort: $::env(RESULTS_DIR)/pre_abc.v -> $::env(RESULTS_DIR)/2_pre_abc.v

flow/scripts/floorplan.tcl
```diff
@@ -87,6 +87,37 @@ remove_buffers
...
+  set tiehi_port $tiehi_lib_name/$tiehi_cell_name/[lindex $env(TIEHI_CELL_AND_PORT) 1]
+
+  restructure -liberty_file $::env(DONT_USE_SC_LIB) -target "area" \
+        -tiehi_port $tiehi_port \
+        -tielo_port $tielo_port
+
+  # remove buffers inserted by abc
+  remove_buffers
+
+  write_verilog $::env(RESULTS_DIR)/post_abc.v
```
gadfort: $::env(RESULTS_DIR)/post_abc.v -> $::env(RESULTS_DIR)/2_post_abc.v

gadfort (changes requested):

flow/scripts/floorplan.tcl
mathursanjiv: These numbers were required for non-restructure run as well so that when comparison is done, all numbers are there. We can re-word this. Let me think more about it.

gadfort: if the numbers are needed then they are needed. The only change that thought would be needed really, is just to keep the naming consistent for the verilog files that are getting written out.

flow/scripts/floorplan.tcl
mathursanjiv: Yes, good point. Done.

flow/scripts/floorplan.tcl
mathursanjiv: Yes, good point. Done.

flow/scripts/floorplan.tcl
mathursanjiv: Updated naming convention.

gadfort (commented):
Thanks for the updates.
Don't forget to update the Openroad submodule

mathursanjiv:
> Thanks for the updates.
> Don't forget to update the Openroad submodule

The OpenROAD submodule changes are being tracked in separate PR. Once that is merged and scripts point to that version, I'll merge this PR.

flow/scripts/synth.tcl
```diff
@@ -63,6 +63,55 @@ if {[info exist ::env(BLOCKS)]} {
...
+        if {[expr $area > $ungroupThreshold]} {
+           select -module $moduleName
+           setattr -mod -set keep_hierarchy 1
+           select -clear
+           puts "Preserving hierarchical module: $moduleName"
+        }
+    }
+  }
+}
+
```
gadfort: This just seems to double the synthesis time. What is the purpose/intent of this? If it's just to get an estimate of the module sizes, you could always run yosys normally without the flatten option (that way we only have to maintain a single core synthesis flow), and then call yosys again setting the `PRESERVE_CELLS` option to the cells exceeding your threshold.

flow/scripts/floorplan.tcl
```diff
@@ -87,6 +87,38 @@ remove_buffers
 
 source $::env(SCRIPTS_DIR)/report_metrics.tcl
 report_metrics "floorplan final"
+set num_instances [llength [get_cells -hier *]]
+puts "number instances before restructure is $num_instances"
+puts "Design Area before restructure"
+report_design_area
```
gadfort: Have you considered just including the number of instances in the report_metics.tcl file? Then you would only have to call `report_metics "floorplan post-restructure"` to get the stats you are looking for, `report_design_area` is already collected as part of that call?

flow/scripts/floorplan.tcl
```diff
@@ -87,6 +87,38 @@ remove_buffers
...
+  # remove buffers inserted by abc
+  remove_buffers
+
+  write_verilog $::env(RESULTS_DIR)/2_post_abc.v
+}
+
+set num_instances [llength [get_cells -hier *]]
+puts "number instances after restructure is $num_instances"
+puts "Design Area after restructure"
+report_design_area
```
gadfort: This could become: `report_metics "floorplan post-restructure"` or whatever the title you need it to be.

flow/util/genMetrics.py
```diff
@@ -252,6 +252,30 @@ def extract_metrics(cwd, platform, design, flow_variant, output, hier_json):
...
+                       logPath+"/2_1_floorplan.log", defaultNotFound=0)
+
+    extractTagFromFile("floorplan_restruct_pre_area", metrics_dict,
+                       "^Design area (\S+) u\^2",
+                       logPath+"/2_1_floorplan.log", occurrence=-2, defaultNotFound=0)
+
+    extractTagFromFile("floorplan_restruct_post_area", metrics_dict,
+                       "^Design area (\S+) u\^2",
+                       logPath+"/2_1_floorplan.log", defaultNotFound=0)
+
```
gadfort: If you chose to change to report_metrics, you would obviously need to modify these.

flow/scripts/floorplan.tcl
```diff
@@ -87,6 +87,38 @@ remove_buffers
 
 source $::env(SCRIPTS_DIR)/report_metrics.tcl
 report_metrics "floorplan final"
+set num_instances [llength [get_cells -hier *]]
+puts "number instances before restructure is $num_instances"
+puts "Design Area before restructure"
+report_design_area
+
+if { [info exist ::env(SYNTH_HIER_AREA_RECOVER)] && $::env(SYNTH_HIER_AREA_RECOVER) == 1 } {
+
```
gadfort: Would it be possible to use a different tag/variable here? It's possible one would not want to run the parts from synth.tcl, but would want to invoke it in OpenROAD.

gadfort (changes requested):
@rovinski should also take a look.

flow/scripts/synth.tcl
mathursanjiv: All these changes are to enable hierarchical synthesis instead of flat synthesis. This would allow for larger designs processing and also in line with standard synthesis practice. Now, these changes are for automated ungroup support in hierarchical synthesis flow. User would set ungroup size in Makefile or design specific config file. Then first synthesis is to get the size of modules. The modules which exceed size threshold, would be preserved and rest flattened. By default, if hierarchical synthesis is enabled (using SYNTH_HIER_AREA_RECOVER), all modules would be preserved. I was avoiding to mix with PRESERVE_CELL setting since that is user controlled and this is auto-internally generated list.

flow/scripts/floorplan.tcl
mathursanjiv: Ok, I can add another variable for this section.

flow/scripts/floorplan.tcl
mathursanjiv: Ok, I'll try this and see if all metrics comes out as desired.

flow/scripts/floorplan.tcl
mathursanjiv: It looks like quite a baggage to include. Since this is optional feature, don't want to bombard with too many metrics. As feature matures and becomes default, we can include full metrics.

flow/scripts/floorplan.tcl
mathursanjiv: It looks like quite a baggage to include. Since this is optional feature, don't want to bombard with too many metrics. As feature matures and becomes default, we can include full metrics.

flow/scripts/floorplan.tcl
maliberty: There is no point to print this if you aren't going to restructure.

flow/scripts/floorplan.tcl
maliberty: same - move inside the if condition

flow/scripts/floorplan.tcl
mathursanjiv: Ok, I can put it inside if statement. Then when we generate variant comparison, we'll get zero for non-restructure variant.

flow/scripts/synth.tcl
gadfort: I don't think this ends up doing what you want it to do. Once the design is mapped with abc, yosys cannot continue to optimize (it just sees blackboxes everywhere). So only the first synthesis will do something, everything else just falls through more or less.
I think you would need to either find a way to merge with the rest of the flow and run abc twice, or after this has run you need to remove it from memory and replace with new read in of the verilog.

mathursanjiv: I've now put hierarchy generation in separate yosys run.

flow/Makefile
```diff
@@ -346,6 +347,9 @@ SYNTH_SCRIPT ?= $(SCRIPTS_DIR)/synth.tcl
 
 $(RESULTS_DIR)/1_1_yosys.v: $(DONT_USE_LIBS) $(WRAPPED_LIBS) $(DONT_USE_SC_LIB) $(DFF_LIB_FILE) $(VERILOG_FILES) $(CACHED_NETLIST) $(LATCH_MAP_FILE) $(ADDER_MAP_FILE)
 	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)
+ifeq ($(SYNTH_HIERARCHICAL), 1)
+	($(TIME_CMD) $(YOSYS_CMD) $(YOSYS_FLAGS) -c $(HIER_REPORT_SCRIPT)) 2>&1 | tee $(LOG_DIR)/1_1_yosys_hier_report.log
+endif
```
gadfort: make this a separate target with the correct dependencies and then you can just add the 1_1_yosys_hier_report.log dependency to the 1_1_yosys.v target.

```
$(LOG_DIR)/1_1_yosys_hier_report.log: ....
   ($(TIME_CMD) $(YOSYS_CMD) $(YOSYS_FLAGS) -c $(HIER_REPORT_SCRIPT)) 2>&1 | tee $(LOG_DIR)/1_1_yosys_hier_report.log

ifeq ($(SYNTH_HIERARCHICAL), 1)
$(RESULTS_DIR)/1_1_yosys.v: $(LOG_DIR)/1_1_yosys_hier_report.log
endif
```

flow/scripts/synth.tcl
```diff
@@ -63,51 +63,28 @@ if {[info exist ::env(BLOCKS)]} {
...
+           puts "Found hierarchical module: $moduleName"
+           if { [catch { select -module $moduleName
            setattr -mod -set keep_hierarchy 1
-           select -clear
-           puts "Preserving hierarchical module: $moduleName"
+           select -clear } result] } {
+              puts "Cound not find module $moduleName"
+           } else {
+             puts "Preserving hierarchical module: $moduleName"
+           }
```
gadfort: You could move this into the other synth and generate file there that outputs what modules to preserve. Then you can just source that here, keeping this file relatively simple.
```
if { [[file isfile $::env(OBJECTS_DIR)/synth_hier_yosys.tcl] } {
  source $::env(OBJECTS_DIR)/synth_hier_yosys.tcl
}
```

flow/scripts/synth_hier_report.tcl
```diff
@@ -0,0 +1,88 @@
...
+}
+abc -liberty $::env(DONT_USE_SC_LIB) \
+    -constr $::env(OBJECTS_DIR)/abc.constr
+
+# Create argument list for stat
+set stat_libs ""
+foreach lib $::env(DONT_USE_LIBS) {
+  append stat_libs "-liberty $lib "
+}
+tee -o $::env(REPORTS_DIR)/synth_hier_stat.txt stat {*}$stat_libs
```
gadfort, line 88: Do the selection of modules here and write that to a file.

flow/scripts/synth.tcl
```diff
@@ -63,51 +63,28 @@ if {[info exist ::env(BLOCKS)]} {
   }
 }
 
-set constr [open $::env(OBJECTS_DIR)/abc.constr w]
-puts $constr "set_driving_cell $::env(ABC_DRIVER_CELL)"
-puts $constr "set_load $::env(ABC_LOAD_IN_FF)"
-close $constr
```
gadfort: This is still needed later in the flow.

flow/Makefile
```diff
@@ -323,6 +324,7 @@ $(WRAPPED_LIBS):
 # |____/ |_| |_| \_| |_| |_| |_|_____|____/___|____/
 #
 synth: versions.txt \
+       $(REPORTS_DIR)/synth_hier_stat.txt \
```
gadfort: this isn't needed

flow/Makefile
```diff
@@ -332,11 +334,13 @@ synth: versions.txt \
 #-------------------------------------------------------------------------------
 SYNTH_SCRIPT ?= $(SCRIPTS_DIR)/synth.tcl
 
-$(RESULTS_DIR)/1_1_yosys.v: $(DONT_USE_LIBS) $(WRAPPED_LIBS) $(DONT_USE_SC_LIB) $(DFF_LIB_FILE) $(VERILOG_FILES) $(CACHED_NETLIST) $(LATCH_MAP_FILE) $(ADDER_MAP_FILE)
-	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)
+$(SYNTH_HIER_REPORT):
```
gadfort: the target should be: `$(OBJECTS_DIR)/mark_hier_stop_modules.tcl` that is the file you are generating and need for the next step

flow/Makefile
```diff
@@ -76,6 +76,7 @@ DESIGN_CONFIG ?= ./designs/nangate45/gcd/config.mk
 
 # Enables hierarchical yosys
 export SYNTH_HIERARCHICAL ?= 0
+export SYNTH_HIER_REPORT = $(REPORTS_DIR)/synth_hier_stat.txt
```
gadfort: this probably is not needed

flow/Makefile
```diff
@@ -332,11 +334,13 @@ synth: versions.txt \
 #-------------------------------------------------------------------------------
 SYNTH_SCRIPT ?= $(SCRIPTS_DIR)/synth.tcl
 
-$(RESULTS_DIR)/1_1_yosys.v: $(DONT_USE_LIBS) $(WRAPPED_LIBS) $(DONT_USE_SC_LIB) $(DFF_LIB_FILE) $(VERILOG_FILES) $(CACHED_NETLIST) $(LATCH_MAP_FILE) $(ADDER_MAP_FILE)
-	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)
+$(SYNTH_HIER_REPORT):
 ifeq ($(SYNTH_HIERARCHICAL), 1)
 	($(TIME_CMD) $(YOSYS_CMD) $(YOSYS_FLAGS) -c $(HIER_REPORT_SCRIPT)) 2>&1 | tee $(LOG_DIR)/1_1_yosys_hier_report.log
 endif
```
gadfort: no need to wrap with `if` here. I think it would be better to have this be a normal target and just to modify the dependencies for `1_1_yosys.v`
```
ifeq ($(SYNTH_HIERARCHICAL), 1)
$(RESULTS_DIR)/1_1_yosys.v: $(OBJECTS_DIR)/mark_hier_stop_modules.tcl
endif
```
This will add the dependency when it's needed/requested, but otherwise will not cause this step to execute.

flow/Makefile
```diff
@@ -332,11 +334,13 @@ synth: versions.txt \
 #-------------------------------------------------------------------------------
 SYNTH_SCRIPT ?= $(SCRIPTS_DIR)/synth.tcl
 
-$(RESULTS_DIR)/1_1_yosys.v: $(DONT_USE_LIBS) $(WRAPPED_LIBS) $(DONT_USE_SC_LIB) $(DFF_LIB_FILE) $(VERILOG_FILES) $(CACHED_NETLIST) $(LATCH_MAP_FILE) $(ADDER_MAP_FILE)
-	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)
+$(SYNTH_HIER_REPORT):
 ifeq ($(SYNTH_HIERARCHICAL), 1)
 	($(TIME_CMD) $(YOSYS_CMD) $(YOSYS_FLAGS) -c $(HIER_REPORT_SCRIPT)) 2>&1 | tee $(LOG_DIR)/1_1_yosys_hier_report.log
```
gadfort, line 338: you'll need: `mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)`

flow/Makefile
```diff
@@ -332,11 +334,13 @@ synth: versions.txt \
...
 endif
+
+$(RESULTS_DIR)/1_1_yosys.v: $(DONT_USE_LIBS) $(WRAPPED_LIBS) $(DONT_USE_SC_LIB) $(DFF_LIB_FILE) $(VERILOG_FILES) $(CACHED_NETLIST) $(LATCH_MAP_FILE) $(ADDER_MAP_FILE) $(SYNTH_HIER_REPORT)
```
gadfort: remove: `$(SYNTH_HIER_REPORT)`

gadfort (changes requested):

mathursanjiv:
All comments should be addressed now. Please check. Thanks.

gadfort (commented):
Looks good to me.

mathursanjiv:
@vvbandeira can you please help why jenkins is failing? I have just moved tools/OpenROAD to newer version. All script changes do not impact runs since they are all off by default.
Since some failures are in GR, adding @eder-matheus as well.

vvbandeira:
> @vvbandeira can you please help why jenkins is failing? I have just moved tools/OpenROAD to newer version. All script changes do not impact runs since they are all off by default.
> Since some failures are in GR, adding @eder-matheus as well.

@mathursanjiv the `tools/OpenROAD` submodule has not been updated for quite a while (24 days), this brings a lot of changes. CI failure is not an anomaly.

I had already reported the errors to Eder, hopefully there will be a fix soon.

flow/util/genMetrics.py
vvbandeira: The metrics name are not correct. There's a standard we are following with two underscores, not one. @ravi-varadarajan could you advise on other changes needed?

ravi-varadarajan: Yes Vitor is right. We will need to ensure that what goes into metrics collection is following the correct format the two underscores is needed but also we need to ensure the naming follows the convention

mathursanjiv: Fixed naming convention after discussion with Ravi

flow/util/genMetrics.py
```diff
@@ -68,8 +68,12 @@ def extractTagFromFile(jsonTag, jsonFile, pattern, file, count=False, occurrence
       content = f.read()
 
     m = re.findall(pattern, content, re.M)
+    useDefault = False
+    if occurrence != -1 and abs(occurrence) > len(m):
+      # Not that many occurrences, use default
+      useDefault = True
 
-    if m:
+    if m and not useDefault :
```
vvbandeira: What is the goal here?

mathursanjiv: In case metrics is optional, may not always occur. In that case, when occurrence (when !=-1) is not found, this code leads to error. The newly added metrics in this PR are optional and not on-by-defalut. One of the metrics uses occurrence = -2. So when the metrics is not present, instead of using default, it errors out.

vvbandeira: This is not clear in the code. For the case of an optional metric my suggestion is to add a kwarg to extractTagFromFile, e.g., `required=True`, and use that to implement your logic.

Setting `occurrence = -2` says to the parser to get the second to last element, thus requiring that the parser finds at least two times the regex. It would be a good idea to add a test to avoid unwanted errors.

mathursanjiv: So you mean to add another argument required=True? Check on occurrence would still be required since we do not know when required=False beforehand as it is dependent on settings in Makefile. Can we have dependency on Makefile variables here in genMetrics.py? I dont think would be right thing to do.

vvbandeira: > So you mean to add another argument required=True?

yes

> Check on occurrence would still be required since we do not know when required=False beforehand as it is dependent on settings in Makefile.

Does the code below solve the issue?

```
patternNotFound = (len(m) == 0)
if patternNotFound and not required:
     # handle here the case for optional metric not found
     # probably just a return to skip the metric (?)
```

>  Can we have dependency on Makefile variables here in genMetrics.py? I dont think would be right thing to do.

We could, but I agree that is not the best path.

mathursanjiv: We are looking for second last occurrence (with occurrence = -2). So len(m) would not be zero. But len(m) < abs(occurence) could be true. "not required" check can be added, but would there be any else for that?

vvbandeira: > We are looking for second last occurrence (with occurrence = -2).

This works with the version on master already, so no new changes needed for this feature.

> So len(m) would not be zero. But len(m) < abs(occurence) could be true. 

The test you added to the `if` statement is the one that I think is not needed (and the source of my concern with code legibility).

> "not required" check can be added, but would there be any else for that?

I don't think an `else` would be needed. The `if` statement that checks if the metric is required should `return` when true or continue with current behavior when false.

mathursanjiv: > > We are looking for second last occurrence (with occurrence = -2).
> 
> This works with the version on master already, so no new changes needed for this feature.
It works when 2 occurrences are there. If the file does not have 2 occurrences it errors. This is only reason of changing code here. We dont want to error out when optional metric is missing.
> 
> > So len(m) would not be zero. But len(m) < abs(occurence) could be true.
> 
> The test you added to the `if` statement is the one that I think is not needed (and the source of my concern with code legibility).
That test helps skip lookup resulting in error and lets code take default value in else part.
> 
> > "not required" check can be added, but would there be any else for that?
> 
> I don't think an `else` would be needed. The `if` statement that checks if the metric is required should `return` when true or continue with current behavior when false.

I think you have understood the intent behind code. Please suggest alternate code considering all these points.

vvbandeira: Add the argument:
```
def extractTagFromFile(jsonTag, jsonFile, pattern, file, count=False, occurrence=-1, defaultNotFound="N/A", t=str, required=True):
```

Replace the `if` statement you added in lines 71-74 with:
```
patternNotFound = (len(m) == 0)
if patternNotFound and not required:
    return
```

Revert the change you made to line 76
`if m and not useDefault :` -> `if m:`

Add `required=False` to the `extractTagFromFile(...)` calls as necessary.

mathursanjiv: This approach can work, but need to check for len(m) < abs(occurrence) otherwise it would give error since m is 1 when call is made with required=False.
Also we are not adding default value in jsonFile, so number of rows in different variants (one with required=True and another required=False) would be different. Thus comparison would miss optional rows.

I'll make these 2 changes in addition and checkin.

vvbandeira:
> We have used pre and post whereever applicable as the last modifier (in cts prerepair and postrepair) Can we do the same for restruct__stdcell__count and restruc__stdcell__area? Change it to
> 
> floorplan__design__restruct__stdcell__count__pre and
> floorplan__design__resturct__stdcell_count__post -- similar for area

@mathursanjiv I am ok with the code changes to `genMetrics.py`. Just review with Ravi the naming for the metrics.
@ravi-varadarajan I saw one with `design__blob` is that ok?

mathursanjiv:
> We have used pre and post whereever applicable as the last modifier (in cts prerepair and postrepair) Can we do the same for restruct__stdcell__count and restruc__stdcell__area? Change it to
> 
> floorplan__design__restruct__stdcell__count__pre and
> floorplan__design__resturct__stdcell_count__post -- similar for area

Done.

mathursanjiv:
> > We have used pre and post whereever applicable as the last modifier (in cts prerepair and postrepair) Can we do the same for restruct__stdcell__count and restruc__stdcell__area? Change it to
> > floorplan__design__restruct__stdcell__count__pre and
> > floorplan__design__resturct__stdcell_count__post -- similar for area
> 
> @mathursanjiv I am ok with the code changes to `genMetrics.py`. Just review with Ravi the naming for the metrics.
> @ravi-varadarajan I saw one with `design__blob` is that ok?

I have removed blob metrics. I think they are details and not essential in metrics report.

mathursanjiv:
I believe everyone is ok with the PR now.
There are 2 Jenkins failures listed below. @maliberty Please merge this PR if you are ok. Thanks!

1. jpeg/sky130hs:
[2021-07-21T04:53:07.549Z] [ERROR] cts__timing__setup__ws__prerepair fail test: -0.39 >= -0.32
[2021-07-21T04:53:07.549Z] [ERROR] cts__timing__setup__ws__postrepair fail test: -0.39 >= -0.32

2. swerv/nangate45
[2021-07-21T05:03:03.900Z] [INFO GRT-0101] Running extra iterations to remove overflow.
[2021-07-21T05:03:05.396Z] [ERROR GRT-0170] Invalid index for position (490, 369).
[2021-07-21T05:03:05.842Z] Error: global_route.tcl, 42 GRT-0170

maliberty (approved):

maliberty:
@arlpetergadfort @ravi-varadarajan please approve if you are ok

gadfort (approved):

PR #87, opened 2021-07-20
Upgrade yosys and OpenROAD to check abc remap flow

mathursanjiv:


rovinski:
I tried this a while ago and the newer version caused a decline in several metrics, so if you can convince the team to accept the update then go ahead.

PR #110, opened 2021-07-29
Improvements to build script

vvbandeira:
- Messages match OpenROAD style.
- Add `--threads N` option.
- Refactor help/usage to include all flags and remove unused options.

### DIFF
--- a/build_openroad.sh
+++ b/build_openroad.sh
@@ -74,7 +74,7 @@ while (( "$#" )); do
       shift
       ;;
     -*|--*=) # unsupported flags
-      echo "[ERROR][FLOW-0004] Unsupported flag $1" >&2
+      echo "[ERROR FLW-0004] Unsupported flag $1." >&2
       exit 1
       ;;
     *) # preserve positional arguments
@@ -86,10 +86,10 @@ done
 
 # Choose install method
 if [ -z ${BUILD_METHOD+x} ] && which docker &> /dev/null; then
-  echo "[INFO FLOW-0000] Using docker build method. This will create a docker image tagged '${DOCKER_TAG}'"
+  echo "[INFO FLW-0000] Using docker build method. This will create a docker image tagged '${DOCKER_TAG}'."
   build_method="DOCKER"
 else
-  echo "[INFO FLOW-0000] Using local build method. This will create binaries at tools/install/"
+  echo "[INFO FLW-0000] Using local build method. This will create binaries at 'tools/install/..'."
   build_method="LOCAL"
 fi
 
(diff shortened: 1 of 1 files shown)


build_openroad.sh
```diff
@@ -44,37 +55,29 @@ while (( "$#" )); do
...
       shift
       ;;
-    -p|--platform)
-      UPDATE_PLATFORM=1
-      shift
-      ;;
-    --no_init)
-      OPENROAD_FLOW_NO_INIT=1
-      shift
-      ;;
```
rovinski, line 71: Why remove these? I'd at least like to keep `--no_init` so that I can rebuild OR without having git complain about untracked files. I can understand removing platforms as we no longer keep private platforms inside `flow/platforms` but then you might as well remove the code for updating as well.

rovinski (changes requested):

build_openroad.sh
vvbandeira: These options are not used anywhere (at least according to running `grep -r` on the master branch):

```
grep -r OPENROAD_FLOW_NO_INIT
build_openroad.sh:      OPENROAD_FLOW_NO_INIT=1     -> this is line 69 that I am proposing removing
```

```
grep -r UPDATE_PLATFORM
build_openroad.sh:      UPDATE_PLATFORM=1      -> line 65, proposing to remove
build_openroad.sh:if [ ! -z ${UPDATE_PLATFORM+x} ]; then   -> 135 to 145, proposing to remove
```

If they are used in an external script (although improbable since they are not exported) let me know and I will keep.

rovinski: Oh that's because line 69 is supposed to be `OPENROAD_FLOW_NO_GIT_INIT=1` to correspond with line 100 https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/f6fdc33bdae9f3213621c8504c5ba739fd8ab60d/build_openroad.sh#L100

I thought that was fixed in a PR but maybe not?

vvbandeira: Fixed on 1af0dd0

maliberty: --local is useful to people who want to build OR without docker afaik.  Do we mention it in the docs?

vvbandeira: yes, on the current [docs here](https://openroad.readthedocs.io/en/latest/user/GettingStarted.html#option-2-build-from-sources-locally), and on the [new here](https://openroad-test.readthedocs.io/en/latest/user/BuildLocally.html#clone-and-build)

rovinski (approved):

PR #112, opened 2021-07-31
Docker use more cached images

vvbandeira:


rovinski (approved):
If it works with the CI then it's fine by me.

PR #126, opened 2021-08-19
bump metrics

eder-matheus:


flow/designs/gf12/coyote/rules.json
```diff
@@ -52,12 +52,12 @@
...
         {
             "field": "detailedroute__route__wirelength",
-            "value": 5726255,
+            "value": 5635046,
             "compare": "<="
         },
         {
             "field": "detailedroute__route__drc_errors__count",
-            "value": 0,
+            "value": 2,
```
maliberty, line 60: I thought the only degradations were to slew/cap.  Is this intentional?

eder-matheus: Not really. I just noticed that grt results differ when running in the secure-ci compared to a run in ucsd servers. I'll investigate it and close this PR for now.

PR #141, opened 2021-08-30
Use berkeley ABC in yosys build

mathursanjiv:
Fix for issue: https://openroad.atlassian.net/browse/OR-1094

### DIFF
--- a/build_openroad.sh
+++ b/build_openroad.sh
@@ -145,7 +145,7 @@ if [ "$build_method" == "DOCKER" ]; then
 
 # Local build
 elif [ "$build_method" == "LOCAL" ]; then
-  $NICE make install -C tools/yosys -j${PROC} PREFIX=$(pwd)/tools/install/yosys CONFIG=gcc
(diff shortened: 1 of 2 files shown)


maliberty:
This only updates the local build and not the docker build.

You should run secure-ci on this change.

rovinski:
Seems weird that we're only updating ABC and not yosys?

maliberty:
> Seems weird that we're only updating ABC and not yosys?
@rovinski yosys hasn't updated to the latest abc which we require.

mathursanjiv:
> This only updates the local build and not the docker build.
> 
> You should run secure-ci on this change.

@maliberty Updated docker build. I cannot do secure CI run from public fork. Would do in a separate branch and share.

maliberty:
> @maliberty Updated docker build. I cannot do secure CI run from public fork. Would do in a separate branch and share.

Approval pending on your result.

mathursanjiv:
> > @maliberty Updated docker build. I cannot do secure CI run from public fork. Would do in a separate branch and share.
> 
> Approval pending on your result.

Yes, have been having IT problems. Hopefully would get it resolved soon and fire runs.

maliberty (approved):
secure-ci passed

PR #160, opened 2021-09-20
Add option for timing based remap

mathursanjiv:
Following changes:
1. Add option for timing based remap in floorplan.tcl and Makefile
2. Fix issue with module area estimate for ungroup in synth_hier_report.tcl
3. Add one more file to delete in Makefile for remap
4. Fix Minor bug in synth_hier_report.tcl to check for existence of ADDER_MAP_FILE script before using it.
No changes to default flow.


ravi-varadarajan:
Validated "UNGROUP size   settings and TIMING_REMAP

flow/scripts/floorplan.tcl
```diff
@@ -87,6 +87,34 @@ if {[info exists ::env(MACRO_WRAPPERS)]} {
   }
 }
 
+##### Restructure for timing #########
+if { [info exist ::env(RESYNTH_TIMING_RECOVER)] && $::env(RESYNTH_TIMING_RECOVER) == 1 } {
+  repair_timing
```
maliberty, line 96: Should remove_buffers be above this block?

flow/scripts/floorplan.tcl
```diff
@@ -87,6 +87,34 @@ if {[info exists ::env(MACRO_WRAPPERS)]} {
...
+
+  # post restructure area/timing report (ideal clocks)
+  repair_timing
+  puts "Post restructure-opt wns"
+  report_worst_slack -max -digits 3
+  puts "Post restructure-opt tns"
+  report_tns -digits 3
+
+}
+
```
maliberty, line 127: Why is "report_metrics "floorplan final" false" before the area recover block below?

flow/scripts/floorplan.tcl
```diff
@@ -87,6 +87,34 @@ if {[info exists ::env(MACRO_WRAPPERS)]} {
...
+  write_verilog $::env(RESULTS_DIR)/2_pre_abc_timing.v
+  restructure -target timing -liberty_file $::env(DONT_USE_SC_LIB) \
+              -abc_logfile $::env(RESULTS_DIR)/$::env(DESIGN_NICKNAME)_$::env(PLATFORM)_abc.log \
+              -work_dir $::env(RESULTS_DIR)
+
+  write_verilog $::env(RESULTS_DIR)/2_post_abc_timing.v
+
+  # post restructure area/timing report (ideal clocks)
+  repair_timing
```
maliberty, line 115: Why do this here and not leave it for later where it naturally happens in the flow?

flow/scripts/floorplan.tcl
```diff
@@ -87,6 +87,34 @@ if {[info exists ::env(MACRO_WRAPPERS)]} {
   }
 }
 
+##### Restructure for timing #########
+if { [info exist ::env(RESYNTH_TIMING_RECOVER)] && $::env(RESYNTH_TIMING_RECOVER) == 1 } {
```
maliberty, line 94: Its a bit odd to have both RESYNTH_TIMING_RECOVER and RESYNTH_AREA_RECOVER since they seem mutually exclusive.  Would it be better to have RESYNTH_MODE with two values?

flow/scripts/synth_hier_report.tcl
```diff
@@ -98,16 +100,42 @@ if { [info exist ::env(REPORTS_DIR)] && [file isfile $::env(REPORTS_DIR)/synth_h
...
+      continue
+    }
+    if {[regexp {   Number of wires.*} $line ]} {
+      set hierarchy_section 0
+      continue
+    }
+    if { $hierarchy_section == 1 } {
+      if {[regexp { +(\S+) +.*} $line -> module_name]} {
+        lappend module_list $module_name
+        puts "Found module $module_name"
```
maliberty, line 54: Is this just for debug?  You will print the module later.

maliberty (changes requested):

flow/scripts/floorplan.tcl
mathursanjiv: Wanted to re-use the buffers inserted by yosys and do timing opt on top of that. remove_buffers before this would increase run time also. Since these results are only to get estimate of timing, this way would be a bit faster.

flow/scripts/floorplan.tcl
mathursanjiv: Technically you can do both but I have not tried it. Single variable would not allow anyone to try both together if one wants to. This code is to allow experimentation of these modes. Ideally using this infrastructure a optimization recipe need to be arrived at which would be a separate project. For example, timing remap could be done post area remap as well.
All this code can be removed as well since this is primarily to show how this features can be used. Eventually, this could all go in timing driven placement mega command.

flow/scripts/synth_hier_report.tcl
mathursanjiv: With this we can see modules in log file for which area is computed. We can remove this as well.

flow/scripts/floorplan.tcl
mathursanjiv: This helps get post re-map timing numbers just to see if re-map helped improve timing. This not final optimization, thus placed before remove_buffers. So all buffers inserted by this optimization are removed.

flow/scripts/floorplan.tcl
mathursanjiv: I am not sure about the purpose of this here as I have not put it. Not sure what makes it true. I can put if after area recover block if that makes sense.

maliberty (approved):

PR #188, opened 2021-10-18
Secure halomacro

msaligane:
@ravi-varadarajan I have started this based on our discussion -- Sorry for taking so long.
I do not have access to the private PDKs at UCSD. So I will let you add those locally and propagate as usual.

Updated sky130hd and sky130hs


tspyrou (changes requested):
https://github.com/efabless/sky130_sram_macros should be used to add a testcase so that we can verify that this continues to work in the CI. LEF units in the efabless ram not compatible with the tech lef. We can fix this and make a local copy in out unit tests and open an issue with efabless to fix it in the upstream version.

ravi-varadarajan:
@msaligane   We can close this PR?  as we have merged in the changes for the halo/channel updates for both public and private PDKS

PR #189, opened 2021-10-18
Update timing rules padding strategy

vvbandeira:


flow/util/genRuleFile.py
```diff
@@ -168,12 +171,8 @@
             customThreshold = 0
             customValue = option['customThreshold']
         if option['usePeriod']:
-            if value >= 0:
-                # value is positive, allow degradation up to period with pad
-                value = - period * option['padding'] / 100
-            else:
-                # value is negative, use current value - pad
-                value -= period * option['padding'] / 100
+            value = value - period * option['padding'] / 100
```
maliberty: -=

flow/designs/tsmc65lp/tinyRocket/rules.json
```diff
@@ -58,7 +58,7 @@
         {
             "field": "detailedroute__route__drc_errors__count",
             "value": 0,
-            "compare": "=="
+            "compare": "<="
```
maliberty, line 61: Were some files just stale - this isn't a change in the code.  Similarly with negative slack limits which should be unchanged.

rovinski:
LGTM

flow/util/genRuleFile.py
vvbandeira: Fixed on 850da6b

flow/designs/tsmc65lp/tinyRocket/rules.json
vvbandeira: They were stale, some still are actually. AES sky130hs was not updated since July. I will start another PR to update all the files using the latest committed version of tools/OpenROAD.

maliberty (approved):

PR #200, opened 2021-10-27
New time report

vvbandeira:


ravi-varadarajan (approved):

flow/util/genMetrics.py
```diff
@@ -132,19 +130,28 @@ def extractGnuTime(prefix, jsonFile, file):
...
-        '^Elapsed time: \S+\[h:\]min:sec. Average CPU: (\S+). Peak memory: \S+KB.',
+        '^Elapsed time:.*CPU time:.*\((\S+)%\).*',
         file)
+    elapsed = jsonFile[prefix + '__runtime__total']
+    cpuPerc = jsonFile[prefix + '__cpu__total']
+    # convert from time string to seconds
+    elapsed = reversed(elapsed.split(":"))
+    cpuTime = 0
+    for time, multFactor in zip(elapsed, [1, 60, 3600]):
+        cpuTime += float(time) * multFactor
```
maliberty: Why is the cpu time being computed from the elapsed time?  That doesn't seem right.

vvbandeira: From: https://man7.org/linux/man-pages/man1/time.1.html
```
       %E     Elapsed real time (in [hours:]minutes:seconds).

       %S     Total number of CPU-seconds that the process spent in
              kernel mode.

       %U     Total number of CPU-seconds that the process spent in user
              mode.

       %P     Percentage of the CPU that this job got, computed as (%U +
              %S) / %E.
```

Parsing both %S and %U then adding them would make the code more complex. The genMetrics script could use some refactoring, but that I think is a separated PR/task.

maliberty: cpu_time != elapsed_time which is the whole point in requesting it be added.  If you want to just use user mode time that would ok.

maliberty (approved):

PR #203, opened 2021-10-27
Patch make issue

vvbandeira:
This is more a quick fix than a solution.
For some designs, when private_tool_script is present, make issue fails with:
execvp: /bin/bash: Argument list too long

### DIFF
--- a/flow/util/utils.mk
+++ b/flow/util/utils.mk
@@ -80,9 +80,17 @@ $(foreach script,$(ISSUE_SCRIPTS),$(script)_issue): %_issue : versions.txt
 	-@rm -f $(VARS_BASENAME).sh $(VARS_BASENAME).tcl $(VARS_BASENAME).gdb
 	@$(foreach V, $(.VARIABLES), \
 	    $(if $(filter-out environment% default automatic, $(origin $V)), \
-	      echo export $V=\""$($V)\""     >> $(VARS_BASENAME).sh ; \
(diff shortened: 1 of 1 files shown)


flow/util/utils.mk
```diff
@@ -118,22 +126,6 @@ endif
...
-	        echo set env\($V\) \""$($V)\"" >> $(VARS_BASENAME).tcl ; \
-	        echo set env $V "$($V)"        >> $(VARS_BASENAME).gdb ; \
-	    ) \
-	)
-	@sed -i '/export \./d' $(VARS_BASENAME).sh
-	@sed -i -e 's/ \// /g' -e 's/"\//"/' $(VARS_BASENAME).sh
-	@sed -i '/set env(\./d' $(VARS_BASENAME).tcl
-	@sed -i -e 's/ \// /g' -e 's/"\//"/' $(VARS_BASENAME).tcl
-	@sed -i '/set env \./d' $(VARS_BASENAME).gdb
-	@sed -i -e 's/ \// /g' -e 's/"\//"/' $(VARS_BASENAME).gdb
```
maliberty, line 135: Is this no longer needed?

maliberty (approved):

flow/util/utils.mk
vvbandeira: This is not reachable, the target `$(VARS_BASENAME).tcl` is not in any make rule. We still do the `sed` step on line 97.

maliberty (approved):

PR #209, opened 2021-11-03
Build script allow user set compilation arguments

vvbandeira:
Allow users to set compilation flags for each tool. The goal is to make life easier for those who do not have a different setup than the one used internally by The OpenROAD Project.

Users can now use a combination of `--<TOOL>-args-overwrite` and `--<TOOL>-args "STRING"` to correctly build ORFS tools in their custom setup.


vvbandeira:
@Colin-Holehouse, @rovinski, @mgwoo -- this is a first stab at improving the experience of building ORFS at Arm, UMich, ABKGroup machines (i.e., custom cron job).

build_openroad.sh
```diff
@@ -18,146 +52,306 @@ Options:
...
 fi
 
-# Clone repositories
-if [ -z ${OPENROAD_FLOW_NO_GIT_INIT+x} ]; then
-  git submodule update --init --recursive
+YOSYS_ARGS="\
+CONFIG=gcc \
+PREFIX=${INSTALL_PATH}/yosys \
+ABCREV=bafd2a7 ABCURL=https://github.com/berkeley-abc/abc \
+"
```
rovinski: Would it make sense to move this and the other ARGS to the top of the file so that they're all in one place?

build_openroad.sh
```diff
@@ -18,146 +52,306 @@ Options:
...
+if [[ "${YOSYS_OVERWIRTE_ARGS}" == "YES" ]]; then
+        echo "[INFO FLW-0013] Overwriting Yosys compilation flags."
+        YOSYS_ARGS="${YOSYS_USER_ARGS}"
+else
+        YOSYS_ARGS="${YOSYS_ARGS} ${YOSYS_USER_ARGS}"
 fi
 
+OPENROAD_APP_USER_ARGS=""
```
rovinski: I think this will overwrite any user args set on line 162?

build_openroad.sh
```diff
@@ -18,146 +52,306 @@ Options:
...
 
-# Clone repositories
-if [ -z ${OPENROAD_FLOW_NO_GIT_INIT+x} ]; then
-  git submodule update --init --recursive
+YOSYS_ARGS="\
+CONFIG=gcc \
+PREFIX=${INSTALL_PATH}/yosys \
+ABCREV=bafd2a7 ABCURL=https://github.com/berkeley-abc/abc \
+"
+if [[ "${YOSYS_OVERWIRTE_ARGS}" == "YES" ]]; then
```
rovinski: Is there an advantage of checking for string equality over using `-z`?

build_openroad.sh
vvbandeira: These variables depend on INSTALL_PATH.
INSTALL_PATH needs to be defined before parsing arguments so it can be updated.
If we define these variables before parsing the arguments they will not get the possibly updated INSTALL_PATH.

I guess we can define before and then just append the INSTALL_PATH dependent after, but I thought this was better. What do you think: define before and append or define after?

build_openroad.sh
vvbandeira: Not really, mostly preference. To a lesser extent, people that are not very familiar with shell scripting might find it easier to read this way.

build_openroad.sh
vvbandeira: Yes, good catch. I moved by mistake from the top. Will fix.

vvbandeira (commented):
@rovinski thanks for the comments. Do you think this is going to make things easier? Any other suggestion or corner case not covered here?

build_openroad.sh
vvbandeira: Fixed on e35afed

build_openroad.sh
vvbandeira: Fixed on f8863c1

build_openroad.sh
vvbandeira: Fixed on f8863c1

rovinski (approved):
Looks good to me.

Issue #202, opened 2021-10-27
DFFLEAGALIZE failed with ASAP7 technology


vijayank88:
**Describe the bug**
A given verilog code passing DFFLEAGLIZE with sky130A/Nangate45 technology, but fail if I use ASAP7 technology. 
* [Toolname] for runtime issues yosys synthesis
* [Design] for example design issues DFFLEGALIZE failed
Error follows:
```
11. Executing TECHMAP pass (map to technology primitives).
11.1. Executing Verilog-2005 frontend: ./platforms/asap7/yoSys/cells_latch.v
11.2. Continuing TECHMAP pass.
12. Executing DFFLIBMAP pass (mapping DFF cells to sequential cells from liberty file).
12.1. Executing DFFLEGALIZE pass (convert FFs to types supported by the target).
ERROR: FF eth_shiftreg.$auto$simplemap.cc:559:simplemap_adffe_sdffe_sdffce$136 (type $_DFFE_PP0P_) cannot be legalized: dffs with async set or reset are not supported
Command exited with non-zero status 1
Elapsed time: 0:02.41[h:]min:sec. Average CPU: 98%. Peak memory: 135204KB.
make: *** [Makefile:366: results/asap7/eth_shiftreg/base/1_1_yosys.v] Error 1

```

**Expected behavior**
Need to pass synthesis stage with ASAP7

**Environment (please complete the following information):**
Please only report issues for supported OSs.
 - OS: Ubuntu 20.04
 - OpenROAD-flow commit : openroad v2.0-1465-gc0801e316

**File Uploads**
Attached eth_shiftreg with src and config file.  
Update Makefile with below line:
DESIGN_CONFIG=./designs/asap7/eth_shiftreg/config.mk

[eth_shiftreg.zip](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/files/7426362/eth_shiftreg.zip)

mi-and-core:
dffs with async set or reset are not supported. To overcome with this add  

`export DFF_LIB_FILE = $($(CORNER)_DFF_LIB_FILE)`

to your config.mk .

vijayank88:
@mi-and-core thanks for the update. Its passed synthesis stage.
Normally user will copy config.mk from other projects and edit die size and design name only.
[...]

PR #210, opened 2021-11-05
Add ASAP7 / Ethernet IP / UART / sha3 designs

vijayank88:


rovinski:
@vijayank88 Please add DCO to your commits. Click on "Details" next to the DCO test and follow the instructions there.

vijayank88:
@rovinski last week through browser i added few test cases and removed. Those commits reflecting here. How do I ignore those things and want to create only today updates to pull request

vijayank88:
With DCO command:
[vijayan@chipware-vijayan flow]$ git rebase HEAD~9 --signoff
First, rewinding head to replay your work on top of it...
Applying: Fix Calibre rule name and only run after packaging testcase
Applying: Add files via upload
Applying: Add files via upload
Applying: Update Makefile
Applying: Update Makefile
Applying: Delete flow/designs/asap7/uart directory
Applying: Delete flow/designs/src/uart directory
Applying: New time format to include CPU seconds
Applying: Remove dead code
Applying: Double quotes break make issue
Applying: Use user-cpu time only
Applying: Remove unused rule
Applying: Split foreach loops to avoid crash
Applying: Bump gf12/bp_single metrics
Applying: Update rules.json
Applying: Update CTS script with optional SETUP_SLACK_MARGIN and HOLD_SLACK_MARGIN user input
Applying: Add setup and hold slack margin
Applying: Add slack margin for repair_timing to address WC corner
Applying: Allow user to set slack margin from make command line
Applying: Reorder timing fix, setup before hold
Applying: Remove short circuit for 2-steps SDC
Applying: update OR
Applying: reset OR to version on master
Applying: Provide easy way to run valgrind
Applying: eth_top,uart,sha3 test cases for ASAP7 technology contributed
[vijayan@chipware-vijayan flow]$ git push --force-with-lease origin master
To github.com:vijayank88/OpenROAD-flow-scripts.git
 ! [rejected]          master -> master (stale info)
error: failed to push some refs to 'git@github.com:vijayank88/OpenROAD-flow-scripts.git'

vvbandeira:
Thanks, @vijayank88 tests are green now.
@rovinski could you help review by assigning it to yourself or someone else?
I will add to CI after this is merged. (@dralabeing FYI)

dralabeing:
@vijayank88 The commits refer to UART- title says ethernet.

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,50 @@
+$(info [INFO-FLOW] Ethernet Design)
+
+DESIGN_DIR                   := $(realpath $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
+DESIGN_PDK_HOME              := $(realpath $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
+
+export DESIGN_NAME            = eth_top
+export DESIGN_NICKNAME        = eth_top
```
maliberty: DESIGN_NICKNAME already defaults to DESIGN_NAME so there is no need to set it here.

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,50 @@
...
+export CORNER                ?= BC
+
+export LIB_FILES             += $($(CORNER)_LIB_FILES)
+export LIB_DIRS              += $($(CORNER)_LIB_DIRS)
+export DB_FILES              += $($(CORNER)_DB_FILES)
+export DB_DIRS               += $($(CORNER)_DB_DIRS)
+export WRAP_LIBS             += $(WRAP_$(CORNER)_LIBS)
+export WRAP_LEFS             += $(WRAP_$(CORNER)_LEFS)
+export TEMPERATURE            = $($(CORNER)_TEMPERATURE)
+export DFF_LIB_FILE = $($(CORNER)_DFF_LIB_FILE)
```
maliberty: Why does any of this have to be specified here - it should be part of the platform setup.

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,50 @@
...
+
+export ABC_CLOCK_PERIOD_IN_PS = 400
+
+export DESIGN_POWER           = VDD
+export DESIGN_GROUND          = VSS
```
maliberty: Does this do anything?

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,50 @@
...
+
+export DESIGN_DIR
```
maliberty: What is this for?

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,50 @@
...
+
+
+export PLATFORM               = asap7
+export DESIGN_PDK_HOME
+
+export CORNER                ?= BC
```
maliberty: duplicate

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,50 @@
...
+export ABC_CLOCK_PERIOD_IN_PS = 400
+
+export DESIGN_POWER           = VDD
+export DESIGN_GROUND          = VSS
+
+export DESIGN_DIR
+
+
+export PLATFORM               = asap7
+export DESIGN_PDK_HOME
```
maliberty: purpose?

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,50 @@
...
+export CORE_UTILIZATION       = 0.05
+export CORE_ASPECT_RATIO      = 1
+export CORE_MARGIN            = 2
+export PLACE_DENSITY          = 0.60
+
+export DONT_USE_SC_LIB         = $(OBJECTS_DIR)/lib/merged.lib
+
+
+export DIE_AREA    = 0 0 800 800
+export CORE_AREA   = 4 4 796 796
```
maliberty: It makes no sense to set both CORE_UTILIZATION / CORE_ASPECT_RATIO  and DIE_AREA / CORE_AREA (the later will just be ignored).

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,50 @@
...
+export CORNER                ?= BC
+
+export PDN_CFG                = $(FOUNDRY_DIR)/openRoad/pdn/grid_strategy-M2-M5-M7.cfg
+
+export CORE_UTILIZATION       = 0.05
+export CORE_ASPECT_RATIO      = 1
+export CORE_MARGIN            = 2
+export PLACE_DENSITY          = 0.60
+
+export DONT_USE_SC_LIB         = $(OBJECTS_DIR)/lib/merged.lib
```
maliberty: why is this needed?

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,50 @@
+$(info [INFO-FLOW] Ethernet Design)
+
+DESIGN_DIR                   := $(realpath $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
+DESIGN_PDK_HOME              := $(realpath $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
+
+export DESIGN_NAME            = eth_top
+export DESIGN_NICKNAME        = eth_top
+export DESIGN                 = eth_top
```
maliberty: Why create another variable?

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,50 @@
...
+
+export VERILOG_FILES          = $(sort $(wildcard $(abspath $(DESIGN_DIR)/../../src/$(DESIGN))/*.v))
```
maliberty: Most other designs don't use relative paths, eg
export VERILOG_FILES          = $(sort $(wildcard $(abspath $(DESIGN_HOME)/src/$(DESIGN))/*.v))

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,50 @@
...
+export LIB_FILES             += $($(CORNER)_LIB_FILES)
+export LIB_DIRS              += $($(CORNER)_LIB_DIRS)
+export DB_FILES              += $($(CORNER)_DB_FILES)
+export DB_DIRS               += $($(CORNER)_DB_DIRS)
+export WRAP_LIBS             += $(WRAP_$(CORNER)_LIBS)
+export WRAP_LEFS             += $(WRAP_$(CORNER)_LEFS)
+export TEMPERATURE            = $($(CORNER)_TEMPERATURE)
+export DFF_LIB_FILE = $($(CORNER)_DFF_LIB_FILE)
+
+export ABC_CLOCK_PERIOD_IN_PS = 400
```
maliberty: Your sdc has -period 1000

maliberty (changes requested):
I've just commented on the first config.mk as there is plenty to sort out

maliberty:
These designs should be added to jenkins/public_tests_all.Jenkinsfile.  As it stands they don't run in the CI.

flow/designs/asap7/eth_top/config.mk
vijayank88: @maliberty copied from ASAP7 built-in examples and just modified as per my design requirement.
Only changed DESIGN NAME and DIE AREA

flow/designs/asap7/eth_top/config.mk
vijayank88: @maliberty Then we need to fix ASAP7 config.mk in other built-in designs.

flow/designs/asap7/eth_top/config.mk
vijayank88: This is opensource design. We don't have design constraints, just set clock to  meet Positive slack. 
ABC_CLOCK_PERIOD used only at synthesis stage.

flow/designs/asap7/eth_top/config.mk
vijayank88: Default CORE_UTILIZATION is 50. I don't have any hard macros, just allowing tool to use full core for placement. Otherwise congestion error will be there.

flow/designs/asap7/eth_top/config.mk
vijayank88: @maliberty If we leave empty default tool will calculate DIE_SIZE and 50% core utilization. Here manual value will over-ride all settings. Consider like its my required DIE_SIZE

flow/designs/asap7/eth_top/config.mk
rovinski: That is because the ASAP7 platform is kind of a mess right now. We should catch these problems where we can.

flow/designs/asap7/eth_top/config.mk
rovinski: The options are mutually exclusive. `DIE_AREA` / `CORE_AREA` are completely ignored in the flow if `CORE_UTILIZATION` exists:
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/abca824a44ebfa6dd1dc079831c738f7ec215477/flow/scripts/floorplan.tcl#L45-L59

flow/designs/asap7/eth_top/config.mk
rovinski: @vijayank88 I think he's asking about `DONT_USE_SC_LIB`. @maliberty I believe it's a bad hack that's persisted because the ASAP7 standard cells are divided up into multiple `.lib`s so it needs to be merged into one `.lib` because of the outstanding issue with ABC being unable to use multiple `.lib`s.

There are multiple problems here.
1) This variable shouldn't be in the design config; it should be in the platform config. But that should get fixed in a different PR probably.
2) ABC should be able to handle multiple libs so we don't need to handle merging them together at all.
3) ASAP7 should honestly not be releasing fragmented libs like that. It is non-standard and makes no sense. I complained a while ago about it, but I don't think the complaint was forwarded to anyone able to address it.

flow/designs/asap7/eth_top/config.mk
rovinski: @vijayank88 this is true for all of the ASAP7 designs. Honestly, nangate45, tsmc65, and gf12 are all better kits to look at for the moment as an example.

rovinski (changes requested):
@vijayank88 you still need to address nearly all of @maliberty's issues. I agree on all points unless specified.

flow/designs/asap7/eth_top/config.mk
rovinski: I don't understand. Then why set this value at all? The flow will extract the value from the SDC. If there isn't a good reason for setting this to 400 then it should be removed.

flow/designs/asap7/eth_top/config.mk
vijayank88: @maliberty  DESIGN_NAME used inside scripts/*.tcl for reference. DESIGN_NICKNAME or DESIGN variable used to point src directory path.

flow/designs/asap7/eth_top/config.mk
vijayank88: @rovinski Just explored the floorplan.tcl.
1st priority goes to FOOTPRINT, then CORE_UTILIZATION else DIE_AREA. Its may be wrong. We should give priority to user DIE_AREA as well. We can't force user to use default values calculated by tool. Maybe # Initialize floorplan using DIE_AREA/CORE_AREA also fall in elseif condition only. 
Experts take decision on this.

vijayank88: I think OpenLane flow uses "relative" variable to decide tool automatically for DIE_AREA. "absolute" will be user input DIE_AREA

flow/designs/asap7/eth_top/config.mk
vijayank88: line 4: DESIGN_PDK_HOME              := $(realpath $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
line 34: export DESIGN_PDK_HOME
I think it can be merged in single line.
While doing script clean/ Removing unwanted variables its should be taken care @maliberty

flow/designs/asap7/eth_top/config.mk
vijayank88: @maliberty  
line 3: variable defined 
line 30: variable exported.
Need to be merged in single line

flow/designs/asap7/eth_top/config.mk
vijayank88: @maliberty  not used inside scripts anywhere.

flow/designs/asap7/eth_top/config.mk
vijayank88: Need to check with who created the initial setup make file.  (built-in example designs and setup)

flow/designs/asap7/eth_top/config.mk
vijayank88: @rovinski is ABC_CLOCK_PERIOD and .sdc clock_period should be same to leverage tool to pick standard cells from library?
Also as per my understand synthesis stage we're not doing any optimization.

flow/designs/asap7/eth_top/config.mk
vijayank88: not found where CORNER variable used in the flow

vijayank88:
#202 blocking the CI, as meta data changed due to update of export DFF_LIB_FILE = $($(CORNER)_DFF_LIB_FILE).
We need to updata metadata file or need to revert my update in config.mk for aes and gcd

vvbandeira:
> These designs should be added to jenkins/public_tests_all.Jenkinsfile. As it stands they don't run in the CI.

@maliberty, I will handle adding the new designs to CI after the PR is merged.

flow/designs/asap7/eth_top/config.mk
rovinski: `DESIGN` is not used anywhere in the flow. And `DESIGN_NICKNAME` is set to `DESIGN_NAME` if it doesn't exist. Therefore, only setting `DESIGN_NAME` is required here.

flow/designs/asap7/eth_top/config.mk
rovinski: By default `ABC_CLOCK_PERIOD_IN_PS` is set to the clock from the `.sdc`. Unless there is reasoning to set `ABC_CLOCK_PERIOD_IN_PS` to something different, such as to meet a specific slack target, then it should not be set. It doesn't sound to me like there is a reason.

flow/designs/asap7/eth_top/config.mk
rovinski: It doesn't really matter because they are mutually exclusive. Only 1 of them is supposed to be set. We should probably be returning a flow error. But that also means that having more than 1 defined in the design config is unnecessary. Please only use the one that you intend.

vijayank88:
> #202 blocking the CI, as meta data changed due to update of export DFF_LIB_FILE = $($(CORNER)_DFF_LIB_FILE). We need to updata metadata file or need to revert my update in config.mk for aes and gcd

I've reverted back my changes to gcd and aes design config.mk. i.e., removed export DFF_LIB_FILE = $($(CORNER)_DFF_LIB_FILE) as these designs don't have Asynchronous signal. Now CI passed. 

NOTE: only if your design has Asynchronous signal include "export DFF_LIB_FILE = $($(CORNER)_DFF_LIB_FILE)" in config.mk otherwise don't use this.

vvbandeira:
> > #202 blocking the CI, as meta data changed due to update of export DFF_LIB_FILE = $($(CORNER)_DFF_LIB_FILE). We need to updata metadata file or need to revert my update in config.mk for aes and gcd
> 
> I've reverted back my changes to gcd and aes design config.mk. i.e., removed export DFF_LIB_FILE = $($(CORNER)_DFF_LIB_FILE) as these designs don't have Asynchronous signal. Now CI passed.
> 
> NOTE: only if your design has Asynchronous signal include "export DFF_LIB_FILE = $($(CORNER)_DFF_LIB_FILE)" in config.mk otherwise don't use this.

Thanks @vijayank88, I think now you only need to address the concerns on your config.mk (mostly clean up) and we can merge this PR.

vijayank88:
@rovinski  @maliberty  ASAP7 platform config.mk and designs config.mk duplicate variable removed and updated incorrect file path. review and update me if any other details required.

flow/designs/asap7/aes/config.mk
```diff
@@ -1,52 +1,38 @@
...
+export LIB_FILES             += $($(CORNER)_LIB_FILES)
+export LIB_DIRS              += $($(CORNER)_LIB_DIRS)
+#export DB_FILES              += $($(CORNER)_DB_FILES)
+#export DB_DIRS               += $($(CORNER)_DB_DIRS)
+##export WRAP_LIBS             += $(WRAP_$(CORNER)_LIBS)
+##export WRAP_LEFS             += $(WRAP_$(CORNER)_LEFS)
+#export TEMPERATURE            = $($(CORNER)_TEMPERATURE)
 
+export DONT_USE_SC_LIB         = $(OBJECTS_DIR)/lib/merged.lib
+#####
```
maliberty: Why is this needed?  It still seems like platform setup.

flow/designs/asap7/eth_top/rules.json
```diff
@@ -0,0 +1,74 @@
...
+        },
+        {
+            "field": "detailedroute__route__drc_errors__count",
+            "value": 0,
+            "compare": "<="
+        },
+        {
+            "field": "finish__timing__setup__ws",
+            "value": 0.0,
+            "compare": ">="
```
maliberty: How much positive slack is there?  I'm wondering if we should up the clock frequency and make this more challenging?

flow/designs/asap7/gcd/config.mk
```diff
@@ -1,37 +1,13 @@
...
-export DESIGN                 = gcd
-
 export PLATFORM               = asap7
 
-export VERILOG_FILES          = $(sort $(wildcard $(abspath $(DESIGN_DIR)/../../src/$(DESIGN))/*.v))
-export SDC_FILE               = $(DESIGN_DIR)/constraint.sdc
-
 export CORNER                ?= BC
+export DESIGN_NAME            = gcd
+export DESIGN_NICKNAME        = gcd
```
maliberty: the default is 
export DESIGN_NICKNAME ?= $(DESIGN_NAME)

so there is no need to specify nickname when it matches name

flow/designs/asap7/ibex/config.mk
```diff
@@ -1,43 +1,29 @@
 $(info [INFO-FLOW] IBEX Design)
 
-DESIGN_DIR                   := $(realpath $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
-DESIGN_PDK_HOME              := $(realpath $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
+export PLATFORM               = asap7
+export CORNER                ?= BC
```
maliberty: does asap7 have corners?

maliberty (changes requested):
my comments apply to many files that are the same.

rovinski:
> my comments apply to many files that are the same.

Same here. There also should not be commented out code.

flow/designs/asap7/eth_top/rules.json
vijayank88: 699 worst slack I've. @maliberty 
I've reduce clock to 1350 from 2000ps. Now 
```
 "field": "finish__timing__setup__ws",
             "value": -27.95,
             "compare": ">="
```
Is that ok now?

vijayank88:
@rovinski  @maliberty 
I've some queries regarding optional library setup. https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/202#issuecomment-962442228
Please let me know your availability will have quick call to discuss and close this PR.

flow/designs/asap7/sha3/config.mk
```diff
@@ -0,0 +1,20 @@
+export PLATFORM               = asap7
+export CORNER                ?= BC
```
maliberty: Do we need this?

flow/designs/asap7/sha3/config.mk
```diff
@@ -0,0 +1,20 @@
...
+
+ifdef ($(ASAP7_USE4X)) 
+  export DIE_AREA               = 0 0 30 30
+  export CORE_AREA              = 0.5 0.5 29.5 29.5
+else
+  export DIE_AREA               = 0 0 90 90
+  export CORE_AREA              = 5.08 5.08 80 80
+endif
+
+export DFF_LIB_FILE           = $($(CORNER)_DFF_LIB_FILE)
```
maliberty, line 13: is this need here or is it part of the platform?  Why is it corner based? It doesn't seem like it would vary by corner.

flow/designs/asap7/jpeg/config.mk
```diff
@@ -1,44 +1,17 @@
-$(info [INFO-FLOW] JPEG Design)
-
-DESIGN_DIR                   := $(realpath $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
-DESIGN_PDK_HOME              := $(realpath $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
+export PLATFORM               = asap7
+export CORNER                ?= BC
```
maliberty: is this needed?

flow/designs/asap7/jpeg/config.mk
```diff
@@ -1,44 +1,17 @@
...
-# These values must be multiples of placement site
-export DONT_USE_SC_LIB        = $(OBJECTS_DIR)/lib/merged.lib
-
 export CORE_UTILIZATION       = 30
 export CORE_ASPECT_RATIO      = 1
 export CORE_MARGIN            = 2
 export PLACE_DENSITY          = 0.60
 
-export DESIGN_DIR DESIGN_PDK_HOME
+export DFF_LIB_FILE           = $($(CORNER)_DFF_LIB_FILE)
```
maliberty, line 16: is this needed?

flow/designs/asap7/ibex/config.mk
maliberty: how was this resolved?

flow/designs/asap7/ibex/config.mk
```diff
@@ -1,43 +1,15 @@
...
+export VERILOG_FILES         = $(sort $(wildcard ./designs/src/$(DESIGN_NICKNAME)/*.v))
+export SDC_FILE              = ./designs/$(PLATFORM)/$(DESIGN_NICKNAME)/constraint.sdc
 
 export CORE_UTILIZATION       = 25 
 export CORE_ASPECT_RATIO      = 1
 export CORE_MARGIN            = 2
 export PLACE_DENSITY          = 0.60
 
-export DONT_USE_SC_LIB         = $(OBJECTS_DIR)/lib/merged.lib
+export DFF_LIB_FILE           = $($(CORNER)_DFF_LIB_FILE)
```
maliberty, line 14: same

flow/designs/asap7/gcd/config.mk
```diff
@@ -1,37 +1,11 @@
...
-export WRAP_LIBS             += $(WRAP_$(CORNER)_LIBS)
-export WRAP_LEFS             += $(WRAP_$(CORNER)_LEFS)
-export TEMPERATURE            = $($(CORNER)_TEMPERATURE)
-
-export ABC_CLOCK_PERIOD_IN_PS = 400
-
-export DESIGN_POWER           = VDD
-export DESIGN_GROUND          = VSS
-
 export CORNER                ?= BC
```
maliberty: same

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,17 @@
+export PLATFORM               = asap7
+export CORNER                ?= BC
```
maliberty: same

flow/designs/asap7/eth_top/config.mk
```diff
@@ -0,0 +1,17 @@
...
+export VERILOG_FILES         = $(sort $(wildcard ./designs/src/$(DESIGN_NICKNAME)/*.v))
+export SDC_FILE              = ./designs/$(PLATFORM)/$(DESIGN_NICKNAME)/constraint.sdc
+export ABC_AREA               = 1
+
+export CORE_UTILIZATION       = 40
+export CORE_ASPECT_RATIO      = 1
+export CORE_MARGIN            = 2
+export PLACE_DENSITY          = 0.60
+
+export DFF_LIB_FILE           = $($(CORNER)_DFF_LIB_FILE)
```
maliberty: same

flow/designs/asap7/aes/config.mk
```diff
@@ -1,52 +1,25 @@
-$(info [INFO-FLOW] AES Design)
+export PLATFORM               = asap7
 
-DESIGN_DIR                   := $(realpath $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
-DESIGN_PDK_HOME              := $(realpath $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
+#User can modify below variables as per design requirement
+export CORNER                ?= BC
```
maliberty: same

flow/designs/asap7/aes/config.mk
```diff
@@ -1,52 +1,25 @@
...
-
+#Floorplaning and placement variables for relative die size
 export CORE_UTILIZATION       = 30
 export CORE_ASPECT_RATIO      = 1
 export CORE_MARGIN            = 2
 export PLACE_DENSITY          = 0.60
 
-export DONT_USE_SC_LIB         = $(OBJECTS_DIR)/lib/merged.lib
-
 ifeq ($(USE_4X),1)
```
maliberty: what is this for?

flow/designs/asap7/sha3/config.mk
vijayank88: @maliberty as platform reference we keep this variable. For ASAP7 all libraries defined specific to CORNER. so without this right libraries can't be chosen.

flow/designs/asap7/jpeg/config.mk
vijayank88: If design has async reset clock flow will fail at synthesis stage. So we need this variable only for specific designs. refer #202  this workaround should be used in design config.mk.

flow/designs/asap7/aes/config.mk
vijayank88: are you pointing USE_4X?
If yes, in ASAP7 there is 1X and 4X cells. We need to check with ASAP7 about in details for this usage.

flow/designs/asap7/ibex/config.mk
vijayank88: please check platforms/asap7/config.mk for CORNER library selection

flow/designs/asap7/eth_top/rules.json
maliberty: yes

flow/designs/asap7/ibex/config.mk
maliberty: The default corner is already defined in platforms/asap7/config.mk so why are you repeating it in each design?  This line should be removed.

flow/designs/asap7/aes/config.mk
dralabeing: @vijayank88 This question is for DONT_USE_SC_LIB

flow/designs/asap7/sha3/config.mk
maliberty: This should be in the platform config as well.

flow/designs/asap7/aes/config.mk
maliberty: Why is this relevant only for a single design?  This won't do anything as you have already defined CORE_UTLIZATION which has higher precedence.

vijayank88: @dralabeing DONT_USE_SC_LIB moved to platform config.mk. highlight in red is removed.

@maliberty aes,ibex,jpeg,gcd are built-in design. I have cleaned up duplicate variables only, if I modify configuration parameter CI will not pass as its affects metadata checks.

I thinks ASAP7 contributor intention is user has to know 7nm library has 1x, 4x cells. Example format shown part of design contribution. So new user can utilize this as well.

If you want to remove this confirm me I can remove it.

flow/designs/asap7/ibex/config.mk
vijayank88: @maliberty Example design should keep this variable will help user to select required CORNER. If we remove user will use Fast corner library only as per platform config.mk. Is that ok?

flow/designs/asap7/aes/config.mk
rovinski: I don't think the 4x is useful. This was only part of the PDK to work around commercial tool license restrictions. OpenROAD has no such restrictions.

flow/designs/asap7/eth_top/constraint.sdc
```diff
@@ -0,0 +1,4 @@
+# Set the current design
+current_design ethmac
```
rovinski: Why is setting the current design necessary? This should already be done in the script before the SDC is loaded

flow/designs/asap7/gcd/config.mk
```diff
@@ -1,37 +1,11 @@
...
+export DESIGN_NAME            = gcd
+export DESIGN_NICKNAME        = $(DESIGN_NAME)
```
rovinski: `DESIGN_NICKNAME` is unnecessary.

flow/designs/src/eth_top/xilinx_dist_ram_16x32.v
```diff
@@ -0,0 +1,50 @@
...
+    RAM16X1D ram27 (.DPO(data_out[27]), .SPO(), .A0(waddr[0]), .A1(waddr[1]), .A2(waddr[2]), .A3(waddr[3]), .D(data_in[27]), .DPRA0(raddr[0]), .DPRA1(raddr[1]), .DPRA2(raddr[2]), .DPRA3(raddr[3]), .WCLK(wclk), .WE(we));
+    RAM16X1D ram28 (.DPO(data_out[28]), .SPO(), .A0(waddr[0]), .A1(waddr[1]), .A2(waddr[2]), .A3(waddr[3]), .D(data_in[28]), .DPRA0(raddr[0]), .DPRA1(raddr[1]), .DPRA2(raddr[2]), .DPRA3(raddr[3]), .WCLK(wclk), .WE(we));
+    RAM16X1D ram29 (.DPO(data_out[29]), .SPO(), .A0(waddr[0]), .A1(waddr[1]), .A2(waddr[2]), .A3(waddr[3]), .D(data_in[29]), .DPRA0(raddr[0]), .DPRA1(raddr[1]), .DPRA2(raddr[2]), .DPRA3(raddr[3]), .WCLK(wclk), .WE(we));
+    RAM16X1D ram30 (.DPO(data_out[30]), .SPO(), .A0(waddr[0]), .A1(waddr[1]), .A2(waddr[2]), .A3(waddr[3]), .D(data_in[30]), .DPRA0(raddr[0]), .DPRA1(raddr[1]), .DPRA2(raddr[2]), .DPRA3(raddr[3]), .WCLK(wclk), .WE(we));
+    RAM16X1D ram31 (.DPO(data_out[31]), .SPO(), .A0(waddr[0]), .A1(waddr[1]), .A2(waddr[2]), .A3(waddr[3]), .D(data_in[31]), .DPRA0(raddr[0]), .DPRA1(raddr[1]), .DPRA2(raddr[2]), .DPRA3(raddr[3]), .WCLK(wclk), .WE(we));
+endmodule
```
rovinski: Why is this file in here? We are not supporting FPGAs.

flow/designs/asap7/eth_top/config.mk
rovinski: Why are `DESIGN_NAME` and `DESIGN_NICKNAME` different here? It is honestly more worth renaming the source folder to `ethmac` and removing `DESIGN_NICKNAME`

rovinski:
It looks like the ethmac design is clearly supposed to include srams but I don't see any macros added?

flow/designs/asap7/eth_top/config.mk
vijayank88: @rovinski as per @maliberty  suggestion if module and directory name same. Will define DESIGN_NAME only, DESIGN_NICKNAME already defined in Makefile. I will update this all design config.mk

flow/designs/asap7/aes/config.mk
vijayank88: @rovinski then I will clean up all 4X die area definition in config.mk

flow/designs/asap7/eth_top/constraint.sdc
vijayank88: @rovinski yes its part of Makefile. I will clean up this as well

flow/designs/src/eth_top/xilinx_dist_ram_16x32.v
vijayank88: @rovinski  its opensource code. xilinx_dis_ram_16x32.v not used in the flow. If ETH_FIFO_XILINX its defined, then it will look for this file

vijayank88:
> It looks like the ethmac design is clearly supposed to include srams but I don't see any macros added?

Using verilog its generating GDSII. How do find myself its need sram macros? 
How to create sram macros for my design? is openRAM i've to use for macro creation?

Do we have any reference designs in OpenROAD flow to handle such memory modules?

flow/designs/asap7/gcd/config.mk
vijayank88: updated

flow/designs/asap7/aes/config.mk
vijayank88: removed 4X usage from config.mk

flow/designs/asap7/gcd/config.mk
vijayank88: updated

rovinski:
> Using verilog its generating GDSII. How do find myself its need sram macros? How to create sram macros for my design? is openRAM i've to use for macro creation?
> 
> Do we have any reference designs in OpenROAD flow to handle such memory modules?

Look at `eth_spram_256x32.v`. It doesn't have anything defined so it will instantiate a 256x32 bit flop array. This of course "works" but is bad design practice and in my opinion very bad for an example design.

We don't currently have any asap7 designs that use macros. For nangate45 we used [bsg_fakeram](https://github.com/bespoke-silicon-group/bsg_fakeram). But that has it's own set of problems (#224).

@maliberty what are your thoughts?

dralabeing:
@maliberty Please share your thoughts on the last review comment on SRAMs. This is the only pending change request. This is blocking the merging of the other 2 designs. I suggest we merge this and then fix based on the suggestion by @rovinski

maliberty:
I agree with @dralabeing about merging and fixing.

maliberty (approved):

maliberty:
@rovinski you need to approve as well since you requested changes.

maliberty:
I've opened https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/235 to track the ram issue.

rovinski (approved):

PR #212, opened 2021-11-07
Add image gallery to Jenkins

vvbandeira:
image
image

### DIFF
--- a/flow/util/genReportTable.py
+++ b/flow/util/genReportTable.py
@@ -237,6 +237,23 @@ table th {
 .main-table td:nth-child(3n+1) {
   border-right: solid 2px;
 }
+div.gallery {
+  margin: 5px;
+  border: 1px solid #ccc;
+  float: left;
+  width: 180px;
+}
+div.gallery:hover {
+  border: 1px solid #777;
+}
+div.gallery img {
+  width: 100%;
+  height: auto;
+}
+div.desc {
+  padding: 15px;
+  text-align: center;
+}
 '''
 
 with open(cssOutput, 'w') as f:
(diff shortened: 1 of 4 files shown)


rovinski (commented):
Looks good. Is there a way to standardize the resolution? Or is it being auto-sized on purpose? Also the instance names are visible on some macros, not sure if that's desired or not.

rovinski:
Also with this, there is no longer a need for the other KLayout-based gallery scripts. You can look at #213 after this is merged.

vvbandeira:
> Looks good. Is there a way to standardize the resolution? Or is it being auto-sized on purpose? Also the instance names are visible on some macros, not sure if that's desired or not.

The instance names are controlled by the GUI, I did not touch that code. @maliberty would know.

The size/style I am working on. I did not want to run all designs on my PC, I was waiting for the CI to do the heavy lift ;)

vvbandeira:
@rovinski I will fix the height instead of width so we have nice rows. The resolution is controlled in the `final_report.tcl` script. Maybe a fixed resolution would make larger designs lose details, perhaps a minimum resolution? But we can address this in a later PR -- the generation of the HTML is independent of the resolution.

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/0dd34eab65336abae4ea48dd9305390858b520c4/flow/scripts/final_report.tcl#L88-L93


image

rovinski:
@vvbandeira I feel like if someone wanted a higher resolution image, then it could be just generated manually. What's the main goal here? Is it to have a nice gallery to put into presentations? Or is it to aid in debugging in some way? If it's the former, then capturing the fine details doesn't really matter.

rovinski (approved):

maliberty:
The images are all 1000 pixels tall which seems like plenty of resolution for this purpose to me.  I do think we should generate some more images (eg placement only, clock tree only) but that can wait for another PR.  

It would be nice if it could be designs in columns and pdks in rows for easier comparison.

maliberty:
is there a reason to use png rather than webp?

vvbandeira:
@maliberty please check the two versions I generated. One is as requested (designs in columns and pdks in rows) and the other is with them switched. Let me know which you think is best.
[reports.zip](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/files/7495197/reports.zip)

vvbandeira:
> is there a reason to use png rather than webp?

The Qt version installed on Jenkins apparently does not support webp, it only saves in png.

maliberty:
> The Qt version installed on Jenkins apparently does not support webp, it only saves in png.

Is it different than what is installed in the new cluster?

maliberty:
@vvbandeira report-gallery-col-platform-row-design.html

vvbandeira:
> > The Qt version installed on Jenkins apparently does not support webp, it only saves in png.
> 
> Is it different than what is installed in the new cluster?

No, same across the board. Qt 5.9.7
Secure CI also generates png.

maliberty:
> No, same across the board. Qt 5.9.7 Secure CI also generates png.

So why does 'save_image foo.webp' on or-3 work for me and generate a webp?

maliberty:
I'm just trying to save space on Jenkins:
-rw-rw-r-- 1 mliberty mliberty 1.3M Nov  8 06:54 foo.png
-rw-rw-r-- 1 mliberty mliberty 157K Nov  8 06:54 foo.webp

maliberty:
We can sort out the webp thing later as it is a minor optimization.

maliberty (approved):

PR #251, opened 2021-12-07
cleanup a cryptic use of tcl

tspyrou:
Signed-off-by: Tom Spyrou <46945466+tspyrou@users.noreply.github.com>


louiic (commented):
Tom ... this looks suspicious ...

It is supposed to be repairing setup but the code change to repair hold.

tspyrou:
@louiic thanks for catching that, fixed.

maliberty (commented):
I find it cryptic both ways but then I always think TCL is cryptic.  There are many uses of {*} through the flow scripts.

tspyrou:
@maliberty I guess it was just cryptic to me then.

louiic:
Matt, … yes it is … this is the quick way to get around “eval”

louiic:
I am ok with “eval” only that … the original line is to do setup repair but the modification is with hold repair… or am I missing something?

tspyrou:
@louiic if the latest update is OK can you approve?

louiic (approved):

louiic:
Tom,

Done.

Louii

PR #252, opened 2021-12-07
Only set FLOW_VARIANT if the argument is given

vvbandeira:

### DIFF
--- a/flow/test/test_helper.sh
+++ b/flow/test/test_helper.sh
@@ -8,11 +8,17 @@ cd "$(dirname $(readlink -f $0))/../"
 DESIGN_NAME=${1:-gcd}
 PLATFORM=${2:-nangate45}
 CONFIG_MK=${3:-config.mk}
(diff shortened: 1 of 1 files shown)


flow/test/test_helper.sh
```diff
@@ -8,11 +8,17 @@ cd "$(dirname $(readlink -f $0))/../"
 DESIGN_NAME=${1:-gcd}
 PLATFORM=${2:-nangate45}
 CONFIG_MK=${3:-config.mk}
-FLOW_VARIANT=${4:-base}
+if [ $# -eq 4 ]; then
+  FLOW_VARIANT=${4:-base}
```
maliberty: :-base is no longer needed

maliberty (approved):

flow/test/test_helper.sh
vvbandeira: Fixed on 12d6f1a

PR #275, opened 2021-12-22
Update padding and grt adjustment for nangate45 platform

luis201420:


maliberty:
You have missed the DCO - please see the details link to fix it.  In general you need to git commit -s to sign your commits.

Do you need to update the OR submodule here?

eder-matheus:
> You have missed the DCO - please see the details link to fix it. In general you need to git commit -s to sign your commits.
> 
> Do you need to update the OR submodule here?

I think he updated the OR submodule to have the metrics with the latest OR.

maliberty:
Ok so just DCO remains

PR #280, opened 2021-12-24
Update padding and grt adjustments for sky130 platforms

luis201420:


eder-matheus (commented):
I think the new values are good. We have a few designs with degradation, but these degradations are pretty small. I'll let @maliberty review and approve it.

maliberty (approved):

PR #282, opened 2021-12-25
Make OpenRoad more compatible with non-linux systems

yurivict:
* use /usr/bin/env to look for bash
* use pre-defined SHELL variable


maliberty:
You are missing DCO - please see the details link to correct

maliberty (approved):

yurivict:
I corrected the DCO.

PR #286, opened 2021-12-28
fix arguments for ppl on gf12/ca53 design

eder-matheus:
Signed-off-by: Eder Monteiro <eder.matheus.monteiro@gmail.com>


ravi-varadarajan (approved):

ravi-varadarajan:
This change will affect the results so will probably require update of rule and OK files?

eder-matheus:
> This change will affect the results so will probably require update of rule and OK files?

True. I'll update it.

maliberty:
@eder-matheus please open a new PR for the metrics updates.

PR #292, opened 2021-12-29
Chameleon hier metrics

vvbandeira:


flow/designs/sky130hd/chameleon_hier/rules-base.json
```diff
@@ -62,12 +62,12 @@
         },
         {
             "field": "finish__timing__setup__ws",
-            "value": -3.05,
+            "value": -1.59,
```
maliberty, line 65: The CI failure didn't include finish__timing__setup__ws so why did it change here?

maliberty (approved):

flow/designs/sky130hd/chameleon_hier/rules-base.json
vvbandeira: We generate the rules file based on the latest metadata file. There was an improvement of that metric, see line 105. It went from -2.65 to -1.19.

PR #299, opened 2022-01-02
Update sky130hd klayout technology file

antonblanchard:
This fixes a number of issues in the sky130hd klayout technology file.

### DIFF
--- a/flow/platforms/sky130hd/sky130hd.lyt
+++ b/flow/platforms/sky130hd/sky130hd.lyt
@@ -27,30 +27,49 @@
    <net-property-name>#1</net-property-name>
    <produce-inst-names>true</produce-inst-names>
    <inst-property-name>#1</inst-property-name>
+   <produce-pin-names>false</produce-pin-names>
+   <pin-property-name>#1</pin-property-name>
    <produce-cell-outlines>true</produce-cell-outlines>
-   <cell-outline-layer>OUTLINE</cell-outline-layer>
+   <cell-outline-layer>prBoundary.boundary</cell-outline-layer>
    <produce-placement-blockages>true</produce-placement-blockages>
    <placement-blockage-layer>PLACEMENT_BLK</placement-blockage-layer>
    <produce-regions>true</produce-regions>
    <region-layer>REGIONS</region-layer>
    <produce-via-geometry>true</produce-via-geometry>
-   <via-geometry-suffix>.drawing</via-geometry-suffix>
-   <via-geometry-datatype>0</via-geometry-datatype>
+   <special-via_geometry-suffix-string>.drawing</special-via_geometry-suffix-string>
+   <special-via_geometry-datatype-string>44</special-via_geometry-datatype-string>
    <produce-pins>true</produce-pins>
-   <pins-suffix>.pin</pins-suffix>
-   <pins-datatype>2</pins-datatype>
-   <produce-obstructions>true</produce-obstructions>
+   <special-pins-suffix-string>.pin</special-pins-suffix-string>
+   <special-pins-datatype-string>16</special-pins-datatype-string>
+   <produce-lef-pins>true</produce-lef-pins>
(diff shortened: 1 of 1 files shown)


antonblanchard:
An example of the issue:

bad

And with this applied:

good

flow/platforms/sky130hd/sky130hd.lyt
```diff
@@ -74,6 +94,16 @@
...
+  <mag>
+   <lambda>1</lambda>
+   <dbu>0.001</dbu>
+   <layer-map>layer_map()</layer-map>
+   <create-other-layers>true</create-other-layers>
+   <keep-layer-names>false</keep-layer-names>
+   <merge>true</merge>
+   <lib-paths>
+   </lib-paths>
+  </mag>
```
rovinski, line 106: We don't use Magic in ORFS but I'll assume this is fine.

flow/platforms/sky130hd/sky130hd.lyt
```diff
@@ -101,7 +132,72 @@
...
+# Example:
+#   var thick_m1 = true
+#   if thickm1
+#     1: 0.5 1.5
+#   else
+#     1: 0.5 1.2
+#   end
+
+</src>
+ </d25>
```
rovinski: No need for this; please remove.

flow/platforms/sky130hd/sky130hd.lyt
```diff
@@ -101,7 +132,72 @@
...
  <connectivity>
+  <connection>li1,67/44,met1</connection>
+  <connection>met1,68/44,met2</connection>
+  <connection>met2,69/44,met3</connection>
+  <connection>met3,70/44,met4</connection>
+  <connection>met4,71/44,met5</connection>
```
rovinski: Can you please add vias as symbols instead of GDS layer/data?

rovinski: Also do you know if via pin layers count towards connectivity? If so, please add them.

rovinski (changes requested):

flow/platforms/sky130hd/sky130hd.lyt
antonblanchard: Ok, removed.

flow/platforms/sky130hd/sky130hd.lyt
antonblanchard: Added the vias as symbols. I'm not sure about the via pin layers, I can't find any reference to the layer/data for them in either the magic tech file or sky130 PDK documentation. Also somewhat suspicious is the datatype is not consistent between layers:

```
via.pin - 68/58
via2.pin - 69/58
via3.pin - 70/48
via4.pin - 71/48
```

antonblanchard:
```
<symbols>li1='67/20+67/5+67/16'</symbols>
<symbols>met1='68/20+68/5+68/16'</symbols>
<symbols>met2='69/20+69/5+69/16'</symbols>
<symbols>met3='70/20+70/5+70/16'</symbols>
<symbols>met4='71/20+71/5+71/16'</symbols>
<symbols>met5='72/20-72/15+72/5+72/16'</symbols>
```

I removed the labels (datatype 5), seems wrong to be including them here. Another thing - why do we subtract datatype 5 (short) only from met5? From reading the PDK these are some sort of resistor for metal layers, and if so shouldn't we subtract it from all metal layers?

rovinski: > I'm not sure about the via pin layers, I can't find any reference to the layer/data for them in either the magic tech file or sky130 PDK documentation.

Yes, that's strange. Let's ignore it for now.

> Another thing - why do we subtract datatype 5 (short) only from met5

I'm not sure I follow - there was no connectivity in the master branch version. Your PR is adding it. I'm guessing you're copying this from elsewhere, in which case I don't know what it's supposed to mean or if it's a mistake.

rovinski (approved):

PR #312, opened 2022-01-12
Add USEMINSPACING OBS OFF to generated LEF files

antonblanchard:
USEMINSPACING OBS defaults to ON, which means we always use minimum wire
spacing rules.

### DIFF
--- a/flow/scripts/generate_lef.tcl
+++ b/flow/scripts/generate_lef.tcl
@@ -11,8 +11,8 @@ set llx [expr [$BBOX  xMin] / $defUnits]
 set width [expr [$BBOX getDX] / $defUnits]
 set height [expr [$BBOX getDY] / $defUnits]
 
-set data "VERSION 5.7 ;\n  NOWIREEXTENSIONATPIN ON ;\n  DIVIDERCHAR \"/\" ;\n  BUSBITCHARS \"\[\]\" ;\nMACRO $BName \
-\n  CLASS BLOCK ; \n  FOREIGN $BName ; \n  ORIGIN $llx $lly ; \n  SIZE  $width BY $height ; "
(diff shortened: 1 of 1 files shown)


maliberty:
The real problem is that we don't have a LEF abstract generator that can give proper DESIGNRULEWIDTH to OBS.  This trades over optimism for over pessimism which is safer.

maliberty (approved):

rovinski:
Isn't this script obsoleted by `generate_abstract_lef` in OpenROAD?

maliberty:
Good point though I think it has the same limitation.

rovinski:
It would be good to fix code there and replace this script with a call to `generate_abstract_lef`
@antonblanchard would you be willing to open a PR in OpenROAD to do so?

antonblanchard:
Thanks, I had no idea `write_abstract_lef` existed. Will work on a PR.

antonblanchard:
@maliberty @rovinski I just took a look at `write_abstract_lef`, and the layer obstructions are macro sized, and don't have any clearance for the pins. I'm guessing that will cause issues for detailed routing. ORFS does the right thing.

Before I have a go at fixing it, I wanted to know if should use `Polygon90Set`, similar to how `bloatSet` etc works, or if there are OpenROAD APIs for this sort of thing.

maliberty:
In general if the pins are at the macro edge you don't need a clearance for them.  Using Polygon90Set is fine (its just a thin wrapper on boost polygon to expose it to swig; you can also just directly use the boost libs in c++).

From LEF:  Normally, obstructions block routing, except for when a pin port overlaps an obstruction (a
port geometry overrules an obstruction).

maliberty:
I think this change is fine so I'll merge it but once you verify that write_abstract_lef meets your needs we should remove it.

PR #329, opened 2022-01-27
Reactivate chameleon test

tspyrou:


flow/platforms/sky130hd/chameleon/chameleon.macro_placment.cfg
```diff
@@ -1,6 +1,6 @@
-_4672_ R0 200 150
```
maliberty, line 1: If you added flow/designs/sky130hd/chameleon/chameleon.macro_placment.cfg why do we still need this?

maliberty (commented):
Can we now remove flow/platforms/sky130hd/chameleon/ ?  It looks like you copied over the files to this design.

flow/platforms/sky130hd/chameleon/chameleon.macro_placment.cfg
tspyrou: I rm'd the platform copy.

maliberty (approved):

PR #353, opened 2022-02-17
Update synth.tcl to call rename -wire and make register names predictable

tspyrou:


maliberty:
The manual says
image

which seems more about wire names than register names.

PR #428, opened 2022-04-17
Remove design from small list pending OR-1412

vvbandeira:
Testcase is still tested on other pipelines, [example](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/jenkins/public_tests_all.Jenkinsfile#L277-L282).


maliberty (approved):
Are you adding it to another pipeline?

vvbandeira:
> Are you adding it to another pipeline?

No, it was already there.

PR #435, opened 2022-04-19
editing flow script for cts to take buffer list 

habibayassin:


flow/platforms/asap7/config.mk
```diff
@@ -99,7 +99,7 @@ export TAPCELL_TCL             = $(PLATFORM_DIR)/openRoad/tapcell.tcl
 
 # TritonCTS options
 export CTS_BUF_CELL            = BUFx4_ASAP7_75t_R
```
maliberty, line 101: You can remove CTS_BUF_CELL as CTS_BUF_LIST replaces it.

maliberty: never mind - this is needed for clock root.

maliberty:
The changes look fine but the CI issues need investigating.

maliberty:
How long does the asap7 tech char take with this list of buffers?

habibayassin:
> How long does the asap7 tech char take with this list of buffers?

it takes very long time to run locally, honestly, I leave it to run in the background then check the logs/reports. However, for ethmac design in ASAP7 cts creates patterns starting from 5000 to 12250000 in increments of 5000. Creation of patterns takes most of the time while running the flow.

maliberty:
> > How long does the asap7 tech char take with this list of buffers?
> 
> it takes very long time to run locally, honestly, I leave it to run in the background then check the logs/reports. However, for ethmac design in ASAP7 cts creates patterns starting from 5000 to 12250000 in increments of 5000. Creation of patterns takes most of the time while running the flow.

Try cutting it down to 2, 4, 8, 12, 24 x buffers and see what the runtime is like.  We want pattern generation to be quick and not dominating.

jjcherry56:
I think there is a serious problem with cts support for multiple buffers that needs to be addressed before they can be deployed to the flow. The "characterization" phase assumes there is a single value for the max cap/slew of the buffers, which is just wrong. Each buffer has a different max cap; it is not a global value. Doing the characterization beyond the max cap of a buffer means that the table data is extrapolated and cts generates max cap violations (which is a current problem with even one buffer being used). The recent change to TechChar::initCharacterization() in 2a387163a1 on 2022-04-19 uses the max of the liberty output pin max_capacitance across the buffers for charMaxCap_, which pretty much guarantees there are going to be even more max slew/max cap violations by cts.

maliberty:
@jjcherry56 just because the characterization range includes extrapolated values doesn't mean those buffers will be selected by CTS.  They are unlikely to be the best delay.  In any case there are bigger problem with the CTS code that has stalled this PR.

PR #466, opened 2022-05-26
Update Microwatt macro liberty files

antonblanchard:

### DIFF
--- a/flow/designs/sky130hd/microwatt/lib/Microwatt_FP_DFFRFile.lib
+++ b/flow/designs/sky130hd/microwatt/lib/Microwatt_FP_DFFRFile.lib
@@ -1,3 +1,4 @@
+
 library (Microwatt_FP_DFFRFile_lib) {
     delay_model : "table_lookup";
     time_unit : "1ns";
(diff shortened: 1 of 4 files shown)


antonblanchard:
There's a few changes here, now I understand more about how liberty file constraints work, as well as how the DFFRAM works. @jjcherry56 @maliberty does this look reasonable?

Firstly setup constraints are measured to the next clock cycle. I'm not sure what I was thinking previously (negative values???), but I now use the clock period of the macro minus the input delay (in this case 8ns clock period, with 0.25*8ns input delay):

```
             timing() {
                 timing_type : setup_rising;
                 related_pin : "CLK";
                 rise_constraint(scalar) {
-                    values("-1.8");
+                    values("6.4000");
                 }
                 fall_constraint(scalar) {
-                    values("-1.8");
+                    values("6.4000");
                 }
             }
```

The other confusion involves the DFFRAMs. Hierarchical STA was showing a max timing path between the positive edge of the clock and the negative edge. That had me confused, until I worked out that the DFFRAM inverts the clock and feeds that into a clock gating cell:

```
            sky130_fd_sc_hd__inv_1 CLKINV(.Y(CLK_B), .A(CLK));
            sky130_fd_sc_hd__dlclkp_1 CG( .CLK(CLK_B), .GCLK(GCLK), .GATE(WE0_WIRE) );
```

This means the inputs to the clock gating cell need to be ready before the negative edge of the clock (such that the output of the clock gating cell is high for the second half of the cycle, and the latch can latch the data). As such, our liberty setup timing constraints need to be measured backwards from the negative edge of the clock, eg:


```
             timing() {
-                timing_type : setup_rising;
+                timing_type : setup_falling;
                 related_pin : "CLK";
                 rise_constraint(scalar) {
-                    values("-3.5");
+                    values("2.5200");
                 }
                 fall_constraint(scalar) {
-                    values("-3.5");
+                    values("2.5200");
                 }
             }
```

jjcherry56:
You haven't explained the structure of your ram for me to really understand what you are modeling, but I don't think your setup constraints are correct. Somewhere inside the ram is the equivalent of a DFF. There is a path from the ram block pin to the DFF clock with delay clk_delay, and a path from the data input (pin not shown in your snippet) with delay data_delay. The setup constraint on the data pin should be data_delay-clk_delay. It doesn't have anything to do with the clock period. But then you talk about a gated negative clock, and if THAT is the clock used to capture the data then the constraint has to be to the negative edge of the clock (timing_type setup_falliing).

Your gated clock constraint looks reasonable, provided that 2.4ns is the difference in the delay between the input to the clock gate and the clock path to the clock gate.

maliberty:
@antonblanchard is this still active?

antonblanchard:
@maliberty I'm working on an update to this now.

maliberty:
> @maliberty I'm working on an update to this now.
still relevant?

antonblanchard:
> > @maliberty I'm working on an update to this now.
> > still relevant?

Thanks for the reminder - I need to update this to use OpenSTA generated liberty files.

maliberty:
Please reopen if you do work on this.

PR #485, opened 2022-06-14
Update sky130hd config.mk - Synthesis

mousaq92:
Updated DONT_USE_CELLS to include the contents of drc_exclude.cells and no_synth.cells in OL. Updated MIN_BUF_CELL_AND_PORTS, ABC_DRIVER_CELL, and ABC_LOAD_IN_FF to match OL.

### DIFF
--- a/flow/platforms/sky130hd/config.mk
+++ b/flow/platforms/sky130hd/config.mk
@@ -19,7 +19,13 @@ export GDS_FILES = $(wildcard $(PLATFORM_DIR)/gds/*.gds) \
 # on all layers.
 # *lpflow* cells are for multi-power domains
 export DONT_USE_CELLS += \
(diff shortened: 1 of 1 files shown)


maliberty:
The commit needs to be signed (see the Details link).
There is timing degradation in many test cases.  Since there are so many changes in one it is hard to know the cause.  This needs to be justified before this can be approved.

antonblanchard:
We shouldn't add a bunch of cells to the don't use list without a clear explanation of why. Some of them appear to have been added to Openlane because of pin access issues that are fixed now.

Openlane also adds every *_1 strength cell to the dont use list, the reason for which was never clear to me. It feels like it was done to help detailed placement (ie start with larger *_2 cells that may get downsized), but the root cause is likely https://github.com/The-OpenROAD-Project/OpenLane/pull/1172

maliberty:
Closed for lack of activity.  Please reopen if still relevant and update.

PR #505, opened 2022-07-01
Add export command

ObiWanRohan:
Added an export command to allow exporting the reports, logs, and results of a complete run

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,13 @@ flow/platforms/*
 !flow/platforms/sky130ram
 flow/private
 
+# Exports
+exports
+
+# Other editor specific config
+.vscode
+.metals
+
 # network
 .nfs*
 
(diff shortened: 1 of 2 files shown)


maliberty:
What is the motivation for this change?  
Why doesn't it copy objects/ ?

ObiWanRohan:
This command is meant to output an export of everything related to one run.
This is a preliminary version, and I wanted to start a discussion on which files would be beneficial to copy over.

Also, I'm fairly new to the OpenROAD flow, and I'm not sure what the `objects` directory contains

maliberty:
objects/ contains some preprocessed inputs and should be considered part of the run.

maliberty:
What are you exporting to?  If you just want to make multiple runs FLOW_VARIANT is more useful as you can still access the results (eg bring up the GUI).

ObiWanRohan:
Thanks for the `FLOW_VARIANT` suggestion. It looks quite useful.

The export feature is supposed to allow creating a single folder with everything. The use case is to share the files and to keep a record.

maliberty:
Sorry I lost track of this PR.   I'm not clear on how this is beneficial over make issue.

ObiWanRohan:
What does `make issue` do?

maliberty:
It packages a tar.gz with everything needed to reproduce a given run

maliberty:
See https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/f2b56bf1a4e9aa5733552faec9584e5f5173dac8/flow/util/utils.mk#L81

vvbandeira:
@ObiWanRohan setting `COPY_ISSUE` might also prove helpful for your use case. See:
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/f2b56bf1a4e9aa5733552faec9584e5f5173dac8/flow/util/makeIssue.sh#L116-L119

maliberty:
The appears to be stale.  You an reopen if there is something to progress on here.

PR #513, opened 2022-07-14
Remove ASAP7 4x library

antonblanchard:
The 4x sized ASAP7 library is unused and appears to have been created to
work around issues with proprietary tools. Remove it.


maliberty:
I think it may still have relevance for academic users with proprietary tools (which was its origin).  I'm fine to park 4X on a branch and remove it from master so long as there is a note in the master README so that people can find them if desired.

antonblanchard:
> I think it may still have relevance for academic users with proprietary tools (which was its origin). I'm fine to park 4X on a branch and remove it from master so long as there is a note in the master README so that people can find them if desired.

Makes sense. That, or an easily found explanation of what the 4x library is for - it took me some searching around to realise why it was created.

maliberty:
You could update the README to make it clear

maliberty (approved):

PR #520, opened 2022-07-18
Fix ASAP7 VIA resistances

antonblanchard:
While it is an academic PDK, having the same via resistance for all
layers is not realistic. The updated values come from one of the
papers on the ASAP7 PDK.


maliberty:
What paper did you take the values from?

You can update the metrics with the update_ok target for the failing design.

antonblanchard:
> What paper did you take the values from?

I found them in here: https://pages.hmc.edu/harris/research/asap7.pdf

maliberty:
The paper reference makes sense.  You have conflicts to resolve

antonblanchard:
@maliberty conflicts resolved, and the CI passes.

maliberty (approved):

PR #544, opened 2022-08-02
nangate45: fix `RCX_RULES` path

joennlae:
Signed-off-by: Jannis Schönleber <joennlae@gmail.com>

### DIFF
--- a/flow/platforms/nangate45/config.mk
+++ b/flow/platforms/nangate45/config.mk
@@ -106,7 +106,7 @@ export CDL_FILE = $(PLATFORM_DIR)/cdl/NangateOpenCellLibrary.cdl
 export TEMPLATE_PGA_CFG ?= $(PLATFORM_DIR)/template_pga.cfg
 
 # OpenRCX extRules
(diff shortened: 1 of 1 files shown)


maliberty:
I don't understand the purpose of this PR.  Why aren't you just setting PLATFORM_DIR?

joennlae:
First of all, very cool project 💯 . I honestly just forgot to explain the purpose of this PR:

The problem is the one that you are describing. I am setting the `PLATFORM_DIR`, but it is not in the `RCX_RULES` path of the `nangate45/config.mk` file.

```mk
export RCX_RULES                           = ./platforms/$(PLATFORM)/rcx_patterns.rules
```

All other `flow/designs/$tech/config.mk` files have the path set like this:
```mk
export RCX_RULES     = $(PLATFORM_DIR)/rcx_patterns.rules
```

That breaks the script when I am not running it from the default position (e.g `flow/`).

This PR addresses this issue.

Working implementation with the fix can be found here:
https://github.com/joennlae/halutmatmul

maliberty (approved):
Thanks.  The change makes sense as it improves the uniform platform setup.

PR #579, opened 2022-09-12
Asap7 update and flow fixes without `.lib`

joennlae:
To make it reviewable.

GIthub cannot render changes due to the swapped library files :-)

I worked heavily with ASAP7 for a project and thought I would upstream my changes.
I will try to reason why every commit makes sense but I am happy to discuss and also eager to see the results of the CI. We probably have to update things after CI again.

- floorplan: add `set_dont_use` before `repair_design`

If one uses `RESYNTH_TIMING_RECOVER` it used cells that we do not want to use during `repair_design` and introduced them to the design.

- cts: move `set_dont_use` before `repair_clock_inverters`

Same issue as before `repair_clock_inverters` cloud introduce cells we dont want to use.

- asap7: update to `asap7` `v28` + adding super low threshold option

Updating `.lib`, `.lef`, `.gds` for asap7 and `config.mk` to newest version.
GIthub cannot render this commit.

Should fix all these issues:

> ### There are some fixes in the newer version:
> 1. ICG lib with "clock gating function and state table" has been fixed. [ICG cells throw an error in library compiler The-OpenROAD-Project/asap7#3](https://github.com/The-OpenROAD-Project/asap7/issues/3)
>    Edit: ICG*DC* cells with incorrect width has been fixed. [ICG*DC* cells have incorrect width The-OpenROAD-Project/asap7#12](https://github.com/The-OpenROAD-Project/asap7/issues/12)
> 2. Large ASNYC FF delay has been fixed. [Wrong timing values (1e31) for ASYNC_DFFHx1_ASAP7_75t_R in the sc7p5t_SEQ_RVT_SS_nldm_201020 library The-OpenROAD-Project/asap7#9](https://github.com/The-OpenROAD-Project/asap7/issues/9)
> 3. Multi-VT GDS files missing has been fixed.

- global_route: remove deprecated option `-report_violating_nets`

- asap7: remove unused `yosys` blackbox files

Cleanup of the `.blackbox.v` files and references for `asap7` as we use `.lib` files

- asap7: allow with `v28` working clock gating cells `ICG*`

Allow ICG* cells as they have been fixed for version 28.

- asap7: make `CTS_BUF_DISTANCE` editable via parent config

Just to allow for more flexibility same as in other techs.

Now we come to the more opinionated commits. Which I am also very happy to discuss.

- global_route: add `MIN_CLOCK_ROUTING_LAYER` + to `asap7` config

[...]

PR #580, opened 2022-09-12
asap7 update and flow fixes lib only

joennlae:
`.lib` files for ASAP7 upgrade to `v28`


maliberty:
Why is this an update of the .lib but not of the lef/gds/cdl?  This puts the setup in a v27 / v28 mix.

joennlae:
> Why is this an update of the .lib but not of the lef/gds/cdl? This puts the setup in a v27 / v28 mix.

These files were small enough to fit into the other PR #579

maliberty:
Ok but this PR can't be merged until the other is ready as well.

joennlae:
yes completely agree

maliberty:
I don't really understand how you intend to merge 579 & then this one.  Is this a two at once?  Will you have to update the ok files again?

joennlae:
@maliberty Yes, it is two at once :-). The problem is that GitHub can not show the diff when the lib files are in the change commit. That is why I build out the `.lib` files of the other commit.

rovinski:
You have some flow errors it seems:
```
make: *** No rule to make target 'platforms/asap7/lib/asap7sc7p5t_AO_RVT_FF_nldm_201020.lib', needed by 'objects/asap7/jpeg/base/lib/asap7sc7p5t_AO_RVT_FF_nldm_201020.lib'.  Stop.
```

joennlae:
> You have some flow errors it seems:
> 
> ```
> make: *** No rule to make target 'platforms/asap7/lib/asap7sc7p5t_AO_RVT_FF_nldm_201020.lib', needed by 'objects/asap7/jpeg/base/lib/asap7sc7p5t_AO_RVT_FF_nldm_201020.lib'.  Stop.
> ```

Yes, that is expected :-) GitHub can not show the diff when adding these library files. So here only the library files are added.

But I see how this is confusing and I will make a pull request with the library files and the changes such that the CI will pass. Forgot that this would be an option (because I can not open a merge request on GitHub when GitHub is not capable of showing the diff, I have to do it via the GitHub command line) :-)

rovinski:
I would suggest the following:
1. Make one PR with any general flow updates that affect all processes and get the CI to pass, then merge.
2. Make another PR with the ASAP7 .lib updates and the minimum number of changes to get the CI to pass. Then merge that PR.
3. Update any flow related settings related specifically to ASAP7, get the CI to pass, then merge.

I think the key thing here is that each PR needs to be atomic, i.e. can stand on it's own between the time that the PRs are merged.

PR #591, opened 2022-09-21
metrics: Correctly report the number of setup and hold violations

antonblanchard:

### DIFF
--- a/flow/scripts/report_metrics.tcl
+++ b/flow/scripts/report_metrics.tcl
@@ -1,3 +1,16 @@
+proc count_violating_endpoints { minmax } {
+    set count 0
+    foreach pin [sta::endpoints] {
+        foreach vertex [$pin vertices] {
+            set slack [ $vertex slack $minmax ]
+            if [ expr $slack < 0 ] {
+                set count [expr $count + 1]
+            }
+        }
+    }
+    return $count;
+}
+
 proc report_metrics { when {include_erc true} {include_clock_skew true} } {
   puts "\n=========================================================================="
   puts "$when check_setup"
@@ -121,12 +134,12 @@ proc report_metrics { when {include_erc true} {include_clock_skew true} } {
     puts "\n=========================================================================="
     puts "$when setup_violation_count"
     puts "--------------------------------------------------------------------------"
-    puts "setup violation count [llength [find_timing_paths -path_delay max -slack_max 0]]"
+    puts "setup violation count [count_violating_endpoints max]"
 
     puts "\n=========================================================================="
     puts "$when hold_violation_count"
     puts "--------------------------------------------------------------------------"
-    puts "hold violation count [llength [find_timing_paths -path_delay min -slack_max 0]]"
+    puts "hold violation count [count_violating_endpoints min]"
 
     set critical_path [lindex [find_timing_paths -sort_by_slack] 0]
     if {$critical_path != ""} {


maliberty:
Thanks for catching this.  This will have to wait for @vvbandeira to be back next week as all the metrics will need updating.

antonblanchard:
Note also that these metrics claim to be a count of violations, but always return 0 or 1. This is because `find_timing_paths` only returns 1 timing path unless you specify `-endpoint_count` or `-group_count`. 

Just guessing here, but is this what the metric should be?

```
-    puts "setup violation count [llength [find_timing_paths -path_delay max -slack_max 0]]"
+    puts "setup violation count [llength [find_timing_paths -path_delay max -slack_max 0 -group_count $sta::group_count_max]]"
```

antonblanchard:
I added `-group_count  $sta::group_count_max`. If this is not the intention of these metrics I can remove it.

flow/scripts/report_metrics.tcl
```diff
@@ -116,12 +116,12 @@ proc report_metrics { when {include_erc true} {include_clock_skew true} } {
     puts "\n=========================================================================="
     puts "$when setup_violation_count"
     puts "--------------------------------------------------------------------------"
-    puts "setup violation count [llength [find_timing_paths -path_delay max -slack_max 0]]"
+    puts "setup violation count [llength [find_timing_paths -path_delay max -slack_max 0 -group_count $sta::group_count_max]]"
```
maliberty: @jjcherry56 is this the best way to do this?  It seems like we want the number of violating endpoints rather than the number of paths but I'm not sure how to get that.

maliberty: @tspyrou do you know how best to do this?

tspyrou: @maliberty iterate the endpoint pins and count the ones with slack < 0. I think this is exposed to tcl.

maliberty: @tspyrou what API is that?

tspyrou: https://github.com/The-OpenROAD-Project/OpenSTA/blob/333b91e72d8b87fe442f0608cc5c73c741e18dda/tcl/StaTcl.i#L277 and
https://github.com/The-OpenROAD-Project/OpenSTA/blob/333b91e72d8b87fe442f0608cc5c73c741e18dda/tcl/StaTcl.i#L4667

maliberty: @antonblanchard can you make this change?

antonblanchard: Thanks @tspyrou. This is what I've come up with:

```
proc count_violating_endpoints { minmax } {
    set count 0
    foreach pin [sta::endpoints] {
        foreach vertex [$pin vertices] {
            set slack [ $vertex slack $minmax ]
            if [ expr $slack < 0 ] {
                set count [expr $count + 1]
            }
        }
    }
    return $count;
}
```

jjcherry56: That's right, although you can use incf instead of expr to bump the count.
Coming to an STA near you is the function sta::endpoint_violation_count which does this is c++ instead of tcl.
Just have to get through the CI hurdles.

antonblanchard: Thanks @jjcherry56!

maliberty:
When @vvbandeira is back we'll need to rebuild all the metrics with this fix.

maliberty:
@vvbandeira please take a look

vvbandeira:
@maliberty @antonblanchard if we update the OpenROAD submodule, can we use [this new API](https://github.com/The-OpenROAD-Project/OpenROAD/blob/54d7611d5cf5caaa6812165d146cd8581768c2eb/src/Metrics.tcl#L113-L114) instead of the Tcl function or are they doing different things?

maliberty:
@vvbandeira I think they are the same - Cherry only recently added that as an sta api

antonblanchard:
The new API came together as we were working this issue, and we should now use it. Want me to update the PR?

vvbandeira:
> The new API came together as we were working this issue, and we should now use it. Want me to update the PR?

No, that's not necessary. I am working on a PR with some metrics-related changes. I can include the new API in that PR. Thanks for catching the problem and fixing it.

vvbandeira:
Closing as stale, after PR #1239 we use `sta` API to get the number of endpoints with violations. Please reopen if this PR should be reconsidered.

PR #593, opened 2022-09-22
Improve support for MacOS

rovinski:
- build_openroad.sh
  - Use OS-specific mechanism to get # CPUs
  - Export Mac-specific paths
- setup_env.sh
  - Export Mac-specific paths
- Makefile
  - Use stdbuf only if it exists

Signed-off-by: Austin Rovinski <rovinski@umich.edu>

### DIFF
--- a/build_openroad.sh
+++ b/build_openroad.sh
@@ -9,7 +9,17 @@ cd "$(dirname $(readlink -f $0))"
 
 # Defaults variable values
 NICE=""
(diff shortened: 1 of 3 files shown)


maliberty (approved):
I have no way to test this so I'll have to assume you have

rovinski:
Yes, tested on MacOS Monterey 12.5.1 and also a quick test on CentOS 7 to make sure Linux still works.

PR #598, opened 2022-09-27
Add Makefile shell test and fallback to bash

rovinski:
This should help fix the longstanding issue as mentioned in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/discussions/330.

The root of the issue is:

1. Ubuntu uses `dash` as its system default shell rather than `bash` which does not support `-o pipefail -c`
2. GNU Make [does not inherit `SHELL` from the environment](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html). It chooses the non-interactive system default shell (`dash`), which is different from the interactive system default shell (`bash`).
3. The system default non-interactive shell counts as setting the `SHELL` variable so it will not be set here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/55469ea7dc713be75fa5caf2007d82aa3dc7c300/flow/Makefile#L102

This results in Make silently choosing `dash` and failing. This PR fixes that by first testing the inherited shell (using `echo`) and then testing the shell return status. On failure, the Makefile forces the shell to bash and tries again, leaving a warning.

The proper solution for Ubuntu and WSL users is to use `sudo dpkg-reconfigure dash` and select `No`, but this works around the issue for a smoother user experience.

This PR does have the limitation that `.SHELLSTATUS` was only added in Make 4.2 (circa 2016). If a user uses Make < 4.2, `.SHELLSTATUS` will be empty and the fallback condition will be falsely triggered. 1) The only harm is the user is forced to use bash instead of their default shell; 2) All our supported OSs use Make >= 4.2; 3) It seems an unlikely scenario this is a real problem.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -99,8 +99,8 @@ endif
 
 #-------------------------------------------------------------------------------
 # Proper way to initiate SHELL for make
-SHELL          ?= /usr/bin/env bash
-.SHELLFLAGS    = -o pipefail -c
+SHELL          := /usr/bin/env bash
+.SHELLFLAGS    := -o pipefail -c
 
 #-------------------------------------------------------------------------------
 # Setup variables to point to root / head of the OpenROAD directory
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -102,6 +102,24 @@ endif
 SHELL          ?= /bin/bash
 .SHELLFLAGS    = -o pipefail -c
 
+# Test that the shell works. Otherwise, warn user and fall back to bash. Note
+# that Make does not inherit $SHELL from the parent environment (See 5.3.2 of
+# GNU Make documentation).
+$(shell echo "" &> /dev/null)
+ifneq ($(.SHELLSTATUS),0)
+  define shell_warn_msg
+[WARNING][FLOW] Shell '$(SHELL)' returned an error on a simple test. Attempting\
+to fall back to '/bin/bash'.
+To fix this warning, try:
+  1) Ubuntu/WSL users: run 'sudo dpkg-reconfigure dash' and select 'No'
+  2) Run 'chsh' and change your shell path to '/bin/bash' or another shell
+  3) Invoke make with 'make SHELL=/bin/bash'
+
+endef
+  $(warning $(shell_warn_msg))
+  SHELL := /bin/bash
+endif
+
 #-------------------------------------------------------------------------------
 # Setup variables to point to root / head of the OpenROAD directory
 # - the following settings allowed user to point OpenROAD binaries to different
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -99,26 +99,8 @@ endif
 
 #-------------------------------------------------------------------------------
 # Proper way to initiate SHELL for make
-SHELL          ?= /bin/bash
+SHELL          ?= /usr/bin/env bash
 .SHELLFLAGS    = -o pipefail -c
-
-# Test that the shell works. Otherwise, warn user and fall back to bash. Note
-# that Make does not inherit $SHELL from the parent environment (See 5.3.2 of
-# GNU Make documentation).
-$(shell echo "" &> /dev/null)
-ifneq ($(.SHELLSTATUS),0)
-  define shell_warn_msg
-[WARNING][FLOW] Shell '$(SHELL)' returned an error on a simple test. Attempting\
-to fall back to '/bin/bash'.
-To fix this warning, try:
-  1) Ubuntu/WSL users: run 'sudo dpkg-reconfigure dash' and select 'No'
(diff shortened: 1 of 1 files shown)


rovinski:
FYI what this looks like for the end user:

## Functional shell:
```
$ make
[INFO FLOW] Using platform directory ./platforms/nangate45
make: Nothing to be done for 'finish'
```

## Non-functional shell with this PR:
```
$ make
/bin/sh: 0: Illegal option -o pipefail
[WARNING][FLOW] Shell '/bin/sh' returned an error on a simple test. Attempting to fall back to '/bin/bash'.
To fix this warning, try:
  1) Ubuntu/WSL users: run 'sudo dpkg-reconfigure dash' and select 'No'
  2) Run 'chsh' and change your shell path to '/bin/bash' or another shell
  3) Invoke make with 'make SHELL=/bin/bash'
[INFO FLOW] Using platform directory ./platforms/nangate45
make: Nothing to be done for 'finish'
```
Note how it still runs to completion.

## Non-functional shell without this PR:
```
$ make
[INFO FLOW] Using platform directory ./platforms/nangate45
/bin/sh: 0: Illegal option -o pipefail
/bin/sh: 0: Illegal option -o pipefail
/bin/sh: 0: Illegal option -o pipefail
/bin/sh: 0: Illegal option -o pipefail
/bin/sh: 0: Illegal option -o pipefail
/bin/sh: 0: Illegal option -o pipefail
/bin/sh: 0: Illegal option -o pipefail
/bin/sh: 0: Illegal option -o pipefail
/bin/sh: 0: Illegal option -o pipefail
/bin/sh: 0: Illegal option -o pipefail
Makefile:666: *** KLayout not found in PATH.  Stop.
```

Interestingly it will always fail on KLayout because it relies on: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/55469ea7dc713be75fa5caf2007d82aa3dc7c300/flow/Makefile#L259-L261

Where `KLAYOUT_CMD` will be an empty string because the shell errored.

flow/Makefile
```diff
@@ -102,6 +102,24 @@ endif
...
+# that Make does not inherit $SHELL from the parent environment (See 5.3.2 of
+# GNU Make documentation).
+$(shell echo "" &> /dev/null)
+ifneq ($(.SHELLSTATUS),0)
+  define shell_warn_msg
+[WARNING][FLOW] Shell '$(SHELL)' returned an error on a simple test. Attempting\
+to fall back to '/bin/bash'.
+To fix this warning, try:
+  1) Ubuntu/WSL users: run 'sudo dpkg-reconfigure dash' and select 'No'
+  2) Run 'chsh' and change your shell path to '/bin/bash' or another shell
```
vvbandeira: Just a small improvement: give the full command to `chsh` similar to `dpkg`.

```
chsh -s /bin/bash
```
?

vvbandeira (approved):

maliberty:
Note that the recommendation has a negative impact on boot time per https://wiki.ubuntu.com/DashAsBinSh

If we are dependent on bash then we should do ```SHELL := /bin/bash``` as no other shell will necessarily work.

rovinski:
> Note that the recommendation has a negative impact on boot time per https://wiki.ubuntu.com/DashAsBinSh
> 
> If we are dependent on bash then we should do `SHELL := /bin/bash` as no other shell will necessarily work.

I don't know what the cleanest solution is here then. The shell was unset from bash because of 545c8d5 which I assume was for compatibility. But the Ubuntu recommendation is the exact opposite. The current solution works, but leaves annoying unhideable shell errors. Maybe I should incorporate a variable `OPENROAD_USE_BASH` which Ubuntu users can set to get rid of the warning. Or maybe I should do the opposite and make `OPENROAD_USE_SYSTEM_SHELL` and default to bash.

maliberty:
In that commit it looks like the issue was "use /usr/bin/env to look for bash".  It was where to find bash that was the problem rather than using bash itself.  Could we use "/usr/bin/env bash" from the Makefile?

rovinski:
Maybe - it'd be a lot simpler if that's the case. It makes more sense that it's a bad commit because based on my reading, `SHELL` is effectively _always_ set when Make is invoked and therefore `SHELL ?= /bin/bash` does nothing. I'll do some testing.

maliberty:
> Maybe - it'd be a lot simpler if that's the case. It makes more sense that it's a bad commit because based on my reading, `SHELL` is effectively _always_ set when Make is invoked and therefore `SHELL ?= /bin/bash` does nothing. I'll do some testing.

Would you explore that path?  If it works I would prefer it to this change.

rovinski:
It does work... that was annoyingly simple.

maliberty (approved):

PR #610, opened 2022-10-12
exclude .cfg from issue files if doesn't exist

### LINKED ISSUE #561: Unable to create reproducible w/ platform files
**Describe the bug**

As discussed in https://github.com/The-OpenROAD-Project/OpenROAD/issues/2163#issuecomment-1218925961 creating an issue using `make *_issue` fails with the following error:
```
[INFO][FLOW] Using platform directory ./platforms/asap7
[INFO-FLOW] ASU ASAP7 - version 2
Default PVT selection: BC
# Creating run-me-noop-asap7-base.sh script
# Creating vars-noop-asap7-base.sh/tcl script
# remove variables starting with a dot
# remove non portable commands
# remove QT variable
# Archiving issue to global_place_noop_asap7_base_2022-08-18_02-29.tar.gz
tar: ./platforms/asap7/*.cfg: Warning: Cannot stat: No such file or directory
````
Passing `EXCLUDE_PLATFORM=1` does allow the reproducible creation to complete:
```
[INFO][FLOW] Using platform directory ./platforms/asap7
[INFO-FLOW] ASU ASAP7 - version 2
Default PVT selection: BC
# Creating run-me-noop-asap7-base.sh script
# Creating vars-noop-asap7-base.sh/tcl script
# remove variables starting with a dot
# remove non portable commands
# remove QT variable
# Archiving issue to global_place_noop_asap7_base_2022-08-18_02-30.tar.gz
# Remove liberty and lef files from tar file
```

[Flow]

**Expected behavior**

`make *_issue` create a tarball with the reproducible including the platform files.

**Environment (please complete the following information):**
Please only report issues for supported OSs.
 - OS: `Linux 4e1cb256e122 4.19.0-21-cloud-amd64 #1 SMP Debian 4.19.249-2 (2022-06-30) x86_64 x86_64 x86_64 GNU/Linux`
 - OpenROAD-flow commit: `2f572bcc84a401f64ce9ff8b1a66337253e4936e`

### PR DESCRIPTION
**Pull request by habibayassin**

### CODE REVIEW

File: flow/util/utils.mk [hunk 1]
```diff
@@ -109,8 +109,13 @@ ISSUE_CP_PLATFORM_FILE_VARS = LIB_FILES \
 ISSUE_CP_FILE_VARS = $(ISSUE_CP_DESIGN_FILE_VARS)
 ifndef EXCLUDE_PLATFORM
 ISSUE_CP_FILE_VARS += $(ISSUE_CP_PLATFORM_FILE_VARS)
+COUNT= $(shell ls -1 $(PLATFORM_DIR)/*.cfg 2>/dev/null | wc -l)
+ifeq ($(COUNT), 0)
+ISSUE_CP_FILES_PLATFORM = $(PLATFORM_DIR)/*.tcl
+else
```
rovinski: I think you could do
```
ISSUE_CP_FILES = $(wildcard $(PLATFORM_DIR)/*.tcl) $(wildcard $(PLATFORM_DIR)/*.cfg)
```
Which would be cleaner and pick up only files that exist

habibayassin: I moved the logic in bash script as Vitor and I agreed, can you take a look?

rovinski: Whatever works. I think the only thing to be careful is that every variable needs to be `export`ed now, otherwise it won't get picked up by the script (I think).

rovinski:
I think you need to modify all `+=` if you are moving to pure bash because bash doesn't support appending strings that way. You have to do `var = "$var new_item1 new_item2"`

habibayassin:
> I think you need to modify all `+=` if you are moving to pure bash because bash doesn't support appending strings that way. You have to do `var = "$var new_item1 new_item2"`

I echoed the value and it seems to be updating correctly

rovinski:
hmm, nevermind then I thought it worked differently.

vvbandeira (approved):

PR #623, opened 2022-10-23
asap7: M6 and M7 SPACINGTABLE does not match WIDTH/SPACING

antonblanchard:
M6 and M7 has:

  WIDTH 0.032 ;
  SPACING 0.032 ;

Which doesn't seem to match the SPACINGTABLE:

  SPACINGTABLE
    PARALLELRUNLENGTH 0.0
    WIDTH 0.0 0.024
    WIDTH 0.025 0.072 ;


antonblanchard:
I'm not sure what the large spacing rule is supposed to be, so I kept it at 0.072. `SPACINGTABLE` appears to be copy/pasted from the smaller metal layers.

maliberty (approved):
This seems like a bug.  I have no idea about .072 either.  It would be good to do a similar to PR to the asap7 repo.

antonblanchard:
> This seems like a bug. I have no idea about .072 either. It would be good to do a similar to PR to the asap7 repo.

I'll submit a bug. The values in the Pad layer look strange too.

maliberty:
I did update asap7 from v27 to v28.  I don't know if it affects this issue or not.  We should probably fix this upstream in asap7 as well

antonblanchard:
> I did update asap7 from v27 to v28. I don't know if it affects this issue or not.

I've rebased on top of this. Having said that, the 1x tech lef only exists in ORFS so it didn't change.

> We should probably fix this upstream in asap7 as well

We did fix it upstream: https://github.com/The-OpenROAD-Project/asap7/commit/b86c7e9336427b85ba7eaf5eb64dcef53a85c0f9

maliberty:
I did revert ng45/swerv_wrapper back from mpl2 -> mpl for now.  If you merge master it should fix the CI errors

PR #624, opened 2022-10-24
add github action to update submodules

habibayassin:

### DIFF
--- a/.github/workflows/github-actions-cron-update-OR.yml
+++ b/.github/workflows/github-actions-cron-update-OR.yml
@@ -31,7 +31,8 @@ jobs:
           delete-branch: true
           title: 'Update OpenROAD submodule'
           reviewers: |
-            habibayassin
+            vvbandeira
+            maliberty
           draft: true
           branch: update-openroad
 
--- a/.github/workflows/github-actions-cron-update-OR.yml
+++ b/.github/workflows/github-actions-cron-update-OR.yml
@@ -19,10 +19,10 @@ jobs:
           cd tools/OpenROAD
           git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git config --local user.name "github-actions[bot]"
-          echo "has_update=$(git --no-pager diff --name-only origin/master..HEAD)\n" >> $GITHUB_OUTPUT
+          echo "::set-output name=has_update::$(git --no-pager diff --name-only origin/master..HEAD)"
           git checkout master
           git pull
-      - if: ${{ steps.remote-update.outputs.has_update }} != ''"
+      - if: "steps.remote-update.outputs.has_update != ''"
         name: Create Draft PR
         uses: peter-evans/create-pull-request@v4
         with:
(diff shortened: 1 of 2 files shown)


habibayassin:
merged the changes on my fork's master and triggered the workflow and it created a draft PR: https://github.com/habibayassin/OpenROAD-flow-scripts/pull/14

maliberty:
Yosys releases once a month with a new tag.  I think it would be better to use the latest tag rather than master.

habibayassin:
> Yosys releases once a month with a new tag. I think it would be better to use the latest tag rather than master.

@maliberty when the fork of Yosys was updated by pulling upstream changes in master the tag was not affected so I don't think we will be able to pull the latest tag

maliberty:
Use git pull --tags to update the tags as well

habibayassin:
@vvbandeira @maliberty I fetched the latest tags in the forked yosys and the GHA is using the latest which is 0.23 not 0.22

habibayassin:
commits need to be squashed before merge

maliberty:
lgtm but I'll leave it to @vvbandeira to approve

.github/workflows/github-actions-cron-update-OR.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      - name: Check out repository code recursively
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Pull changes in OpenROAD submodule
+        id: remote-update
+        run: |
+          cd tools/OpenROAD
+          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
vvbandeira: Where did the `41898282` came from?

.github/workflows/github-actions-cron-update-OR.yml
```diff
@@ -0,0 +1,40 @@
...
+          git config --local user.name "github-actions[bot]"
+          echo "::set-output name=has_update::$(git --no-pager diff --name-only origin/master..HEAD)"
+          git checkout master
+          git pull
```
vvbandeira, line 24: Are we sure on the order here? Should we do a `fetch` before setting the `has_update` value?

.github/workflows/github-actions-cron-update-yosys.yml
```diff
@@ -0,0 +1,43 @@
...
+          submodules: recursive
+          fetch-depth: 0
+      - name: Pull changes in yosys submodule
+        id: remote-update
+        run: |
+          cd tools/yosys
+          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          git config --get remote.origin.fetch
+          echo "::set-output name=has_update::$(git --no-pager diff --name-only origin/master..HEAD)"
```
habibayassin: @vvbandeira do you think I should check the difference between the head and the latest tag not the master? or it wouldn't matter?

vvbandeira: since we are going to update to the tag (going from Matt's comment earlier) it makes sense

vvbandeira (approved):

PR #630, opened 2022-11-01
fix exporting variables

habibayassin:


vvbandeira (approved):
approving to unbreak tar files from CI. we still need some improvements

PR #643, opened 2022-11-11
exporting all variables to bash

habibayassin:


habibayassin:
@vvbandeira I filtered the variables from makefile and I also left the filters in the makeIssue script as an extra layer of filteration....let me know if this is what you're expecting

vvbandeira (approved):

PR #638, opened 2022-11-08
ASAP7: `v28` lib update + changes (merge via cli)

joennlae:
- floorplan: add `set_dont_use` before `repair_design`
- cts: move `set_dont_use` before `repair_clock_inverters`
- asap7: update to `v28` `.lef` files
- asap7: update to `v28` `.gds` files
- asap7: update to `v28` adaption of `config.mk` file
- global_route: remove depreceated option `-report_violating_nets`
- asap7: remove unused `yosys` blackbox files
- asap7: allow with `v28` working clock gating cells `ICG*`
- asap7: make `CTS_BUF_DISTANCE` editable via parent config
- global_route: add `MIN_CLOCK_ROUTING_LAYER` + to `asap7` config
- asap7: make tapcell work with low and super low threshold
- asap7: use correct gds, buffer, fille for corresponding vth
- asap7: update `yosys` adder, clkgate, latch and dff for multi-vt
- asap7: fix `.lef` file according to upstream repo
- asap7: revert back to one buffer
- asap7: remove `4x` gds + lef
- asap7: remove `4x` option from `asap7`
- scripts: move `set_dont_use` into `load.tcl` and remove other refs
- asap7: update to `v28` `.lib` files
- asap7: update `metadata` after `v28` lib upgrade Signed-off-by: Jannis Schönleber <joennlae@gmail.com>
- asap7: apply changes from `review`

PR #644, opened 2022-11-15
asap7: updated openlane folder as part of asap7 upgrade

oharboe:
This pull request has a single commit on top of #638 to update the openlane folder, seems to work, but needs https://github.com/The-OpenROAD-Project/OpenLane/issues/1486 fixed before it can be fully tested.


maliberty:
How is this supposed to work when the files don't exist in ORFS?

oharboe:
> How is this supposed to work when the files don't exist in ORFS?

I tried to fix this by copying files over from https://github.com/The-OpenROAD-Project/asap7

rovinski:
I think this is mostly a subset of #638 and #579?

oharboe:
> I think this is mostly a subset of #638 and #579?

Ah. Rather a lot more going on in those two pull requests...

How can I test #638 and #579 with OpenLane?

maliberty:
There are a number of CI failures that need investigating.  I see this isn't a pure upgrade as you have
```
export MIN_CLOCK_ROUTING_LAYER = M4
```
which might be good but isn't about upgrading to 28.  Likewise with removing *ICG* from DONT_USE_CELLS.  I'm ok with keeping these but you need to establish that they aren't causing degradations in QOR.

oharboe:
> There are a number of CI failures that need investigating. I see this isn't a pure upgrade as you have
> 
> ```
> export MIN_CLOCK_ROUTING_LAYER = M4
> ```
> 
> which might be good but isn't about upgrading to 28. Likewise with removing _ICG_ from DONT_USE_CELLS. I'm ok with keeping these but you need to establish that they aren't causing degradations in QOR.

This is a draft pull request on top of #638 where I only one commit: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/644/commits/d07ede1f8b740b699f82cf2791ad89a06c50599b.

Your comments should go on #638

maliberty:
I'm trying to figure out how to bring this to a successful conclusion.  There are too many PRs and changes intermixed still.  I have no idea the source of the failures.  Can I ask for the following:

- One PR that just adds the new files but does nothing else (large but easy to approve as no functional change)
- One PR that just switches from the old (27) files to the new (28) files with no other configuration changes.  This will isolate any issues with the new cells (small and targeted)
- Follow on PRs that make a single change whose QOR impact can be evaluated (small and targeted)

rovinski:
@maliberty this isn't the OP of the changes, that was @joennlae

maliberty:
@rovinski OP?

rovinski:
The person who originally made the changes in the PR. I think this one is just grabbing those same changes.

maliberty:
@rovinski I see - I'll put the comment in #638 then

oharboe:
Easily done when ASAP7 is upgraded.

PR #658, opened 2022-11-27
Change naming of design variables

habibayassin:


habibayassin:
need to do corresponding updates in OpenROAD

PR #666, opened 2022-12-02
logging build

### LINKED ISSUE #664: Create build_openroad.sh log automatically
**Is your feature request related to a problem? Please describe.**
Create a build log for `build_openroad.sh` automatically so that users don't have to follow extra steps when submitting build issues.

**Describe the solution you'd like**
See above

**Describe alternatives you've considered**
N/A

**Additional context**
We could use a solution similar to the one [here](https://stackoverflow.com/questions/3173131/redirect-copy-of-stdout-to-log-file-from-within-bash-script-itself).

habibayassin:

### DIFF
--- a/build_openroad.sh
+++ b/build_openroad.sh
@@ -339,7 +339,7 @@ __update_openroad_app_latest()
 )
 
 __common_setup()
-{       
+{
         # Clone repositories
         if [ -z "${OPENROAD_FLOW_NO_GIT_INIT+x}" ]; then
                 echo "[INFO FLW-0002] Updating git submodules."
@@ -359,6 +359,13 @@ __common_setup()
         fi
 }
 
+__logging()
+{
+        mkdir -p "${buildDir}"
+        exec > >(tee -i ${buildDir}/build-$(date +%s).log)
+        exec 2>&1
+}
+
 __common_setup
 
 if [ ! -z "${CLEAN_FORCE+x}" ]; then
@@ -371,12 +378,9 @@ if [ ! -z "${CLEAN_BEFORE+x}" ]; then
         echo "[INFO FLW-0016] Cleaning up previous binaries and build files."
         git clean ${CLEAN_CMD} tools
         git submodule foreach --recursive git clean ${CLEAN_CMD}
-        rm -rf "${buildDir}"
 fi
 
-mkdir -p "${buildDir}"
-exec > >(tee -i ${buildDir}/build-$(date +%s).log)
-exec 2>&1
+__logging
 
 # Choose install method
 if [ -z "${LOCAL_BUILD+x}" ] && command -v docker &> /dev/null; then
--- a/build_openroad.sh
+++ b/build_openroad.sh
@@ -8,7 +8,6 @@ set -eu
 cd "$(dirname $(readlink -f $0))"
 
 # Defaults variable values
-buildDir="build"
 NICE=""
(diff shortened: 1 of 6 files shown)


build_openroad.sh
```diff
@@ -365,12 +366,18 @@ if [ ! -z "${CLEAN_FORCE+x}" ]; then
...
         echo "[INFO FLW-0016] Cleaning up previous binaries and build files."
         git clean ${CLEAN_CMD} tools
         git submodule foreach --recursive git clean ${CLEAN_CMD}
+        rm -rf "${buildDir}"
 fi
 
+mkdir -p "${buildDir}"
+exec > >(tee -i ${buildDir}/build-$(date +%s).log)
+exec 2>&1
+
```
vvbandeira: Should this be at the top of the file to catch all messages? Or, to avoid conflicts with the git clean, we should refactor the steps above into a bash function and orchestrate the flow to do cleanup, start the log, and then build. To try and catch as much as possible from the output.

build_openroad.sh
```diff
@@ -338,7 +339,7 @@ __update_openroad_app_latest()
 )
 
 __common_setup()
-{
+{
```
vvbandeira: Just me nitpicking, do not add white spaces at the end of lines.

build_openroad.sh
```diff
@@ -365,12 +366,18 @@ if [ ! -z "${CLEAN_FORCE+x}" ]; then
 else
         CLEAN_CMD="-x -d --interactive"
 fi
+
 if [ ! -z "${CLEAN_BEFORE+x}" ]; then
         echo "[INFO FLW-0016] Cleaning up previous binaries and build files."
         git clean ${CLEAN_CMD} tools
         git submodule foreach --recursive git clean ${CLEAN_CMD}
+        rm -rf "${buildDir}"
```
vvbandeira: I don't think we need to delete the build dir. The git clean command should control that.

vvbandeira (changes requested):
can you also modify the Jenkinsfiles to save the build log as an artifact? Might prove useful when CI fails to build for any reason.

build_openroad.sh
habibayassin: I pushed some changes please let me know if that's what you had in mind

build_openroad.sh
```diff
@@ -358,19 +359,29 @@ __common_setup()
         fi
 }
 
+__logging()
+{
+        mkdir -p "${buildDir}"
+        exec > >(tee -i ${buildDir}/build-$(date +%s).log)
```
rovinski: Hmmm, I was only thinking about saving the most recent build log. It seems like most users won't need to save multiple logs. On the other hand, saving logs doesn't hurt much. @vvbandeira can decide.

build_openroad.sh
vvbandeira: Almost. I already had a branch with these changes locally so I just merged our changes.

build_openroad.sh
vvbandeira: At first, I thought it would be nice to have multiple logs, but that will also add the possibility of the user submitting the wrong log. So let's keep only the latest to simplify things.

vvbandeira:
@habibayassin I think the only change needed is in the Jenkinsfiles to save the log as an artifact. Other than that it looks good to merge.

vvbandeira (approved):

PR #681, opened 2022-12-10
defer expansion of KLayout variables

### LINKED ISSUE #669: KLayout is required to run any flow step
### Subject

[Flow] for any util, flow Makefile, or flow script issues.

### Describe the bug

Because of changes to the issue script, KLayout is now required to do anything. The reason is because of this line:
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/f2b56bf1a4e9aa5733552faec9584e5f5173dac8/flow/util/utils.mk#L79

This forces the check for KLayout to be immediately expanded:
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/f2b56bf1a4e9aa5733552faec9584e5f5173dac8/flow/Makefile#L261-L263

This variable is supposed to be a delayed expansion so that it only gets checked if it actually gets used.

### Expected Behavior

KLayout should only be required to run KLayout steps.

### Environment

```shell
Git commit: 6fe8dc53d9fe53d67d604fede6c7b2228ae2a196
kernel: Linux 5.10.16.3-microsoft-standard-WSL2
os: Ubuntu 20.04.5 LTS (Focal Fossa)
cmake version 3.16.3
gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
./etc/Env.sh: line 45: clang: command not found
```

### To Reproduce

Run any make command without KLayout installed.

### Relevant log output

```shell
rovinski@DESKTOP:~/OpenROAD-flow-scripts/flow$ make
[INFO][FLOW] Using platform directory ./platforms/nangate45
util/utils.mk:79: *** KLayout not found in PATH.  Stop.
rovinski@DESKTOP:~/OpenROAD-flow-scripts/flow$ make clean_all
[INFO][FLOW] Using platform directory ./platforms/nangate45
util/utils.mk:79: *** KLayout not found in PATH.  Stop.
```

### Screenshots

_No response_

### Additional Context

You might be able to change the `:=` into a `=` to make `ISSUE_VARIABLES` a delayed expansion as well. However, the problem would still occur anytime this variable is expanded.

habibayassin:


flow/util/utils.mk
```diff
@@ -76,15 +76,20 @@ define \n
 
 
 endef
-export ISSUE_VARIABLES := $(foreach V, $(.VARIABLES),$(if $(filter-out environment% default automatic, $(origin $V)), $(if $(filter-out .% %QT_QPA_PLATFORM% %TIME_CMD%, $(V)),$V=$($V))${\n}))
+
+ifneq ($(shell command -v klayout),)
+  export ISSUE_VARIABLES := $(foreach V, $(.VARIABLES),$(if $(filter-out environment% default automatic, $(origin $V)), $(if $(filter-out .% %QT_QPA_PLATFORM% %TIME_CMD%, $(V)),$V=$($V))${\n}))
+else
+  export ISSUE_VARIABLES = 
+endif
```
vvbandeira: This will not allow users that do not have `klayout` installed to create make issues. We should support that. Could you test if adding the KLAYOUT_CMD and KLAYOUT_FOUND variables to the filter-out command would fix the issue? If not, we will need to change how we handle these variables, I think changing the makeIssue is not the best path.

vvbandeira (changes requested):

vvbandeira (approved):

PR #702, opened 2022-12-19
Makefile: remove DESIGN_CONFIG hacking vestiges at top of Makefile

oharboe:


maliberty:
I believe some people use this as a fast way to turn on designs.  I'd like more input first as this is stylistic.

rovinski:
I don't care much one way or another, but I know there are some people that really like them.

PR #703, opened 2022-12-19
Makefile: remove ascii art

oharboe:
Part of making Makefile terser.


maliberty:
They are marking sections which seems reasonable.  If  you want to make the delimiters smaller that's fine but removing them altogether reduces the readability imho

rovinski:
I like the ASCII art 😢

PR #707, opened 2022-12-21
Additional gds misspelling

oharboe:
I think this is a misspelling, but I'm unsure...

### DIFF
--- a/flow/designs/sky130hd/chameleon/config.mk
+++ b/flow/designs/sky130hd/chameleon/config.mk
@@ -34,7 +34,7 @@ export CORE_AREA  = 20 20 2900 3500
 
 export chameleon_DIR = ./designs/$(PLATFORM)/$(DESIGN_NICKNAME)
 
-export ADDITIONAL_GDS_FILES  = $(chameleon_DIR)/gds/apb_sys_0.gds.gz \
+export ADDITIONAL_GDS  = $(chameleon_DIR)/gds/apb_sys_0.gds.gz \
                                $(chameleon_DIR)/gds/DMC_32x16HC.gds.gz \
                                $(chameleon_DIR)/gds/DFFRAM_4K.gds.gz \
                                $(chameleon_DIR)/gds/ibex_wrapper.gds.gz
--- a/flow/designs/sky130hd/microwatt/config.mk
+++ b/flow/designs/sky130hd/microwatt/config.mk
@@ -17,7 +17,7 @@ export ABC_CLOCK_PERIOD_IN_PS = 25000
 
 export microwatt_DIR = ./designs/$(PLATFORM)/$(DESIGN_NICKNAME)
 
-export ADDITIONAL_GDS_FILES  = $(wildcard $(microwatt_DIR)/gds/*.gds.gz)
+export ADDITIONAL_GDS  = $(wildcard $(microwatt_DIR)/gds/*.gds.gz)
 
 export ADDITIONAL_LEFS  = $(wildcard $(microwatt_DIR)/lef/*.lef)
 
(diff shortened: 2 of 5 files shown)


rovinski:
No, I think it was mostly intentional. `ADDITIONAL_LIBS` is clearly plural and so is `ADDITIONAL_LEFS`, so it is obvious that they are list variables. How do you make it obvious that GDS is plural? `GDSS` is bad, `GDSES` isn't good either.

maliberty:
I agree with @rovinski that the pluralization is important and hard to effect otherwise.

If you really want consistency then you could change ADDITIONAL_LIBS -> ADDITIONAL_LIB_FILES  etc

oharboe:
@rovinski @maliberty The purpose of this PR is to make the code consistent as it stands. To rename ADDITIONAL_GDS to something else, is a concern for a different pull request, no?

The bp_multi_top nangate45 failure, I've seen in other PRs too. I think it is a false positive.

maliberty:
Making it consistent also makes it worse in a different way so it isn't a pure win.  That's why I suggested going the other way if you want consistency "If you really want consistency then you could change ADDITIONAL_LIBS -> ADDITIONAL_LIB_FILES etc"

"bp_multi_top nangate45 failure," is fixed in master now.

oharboe:
@maliberty I am happy to do a search and replace on the entire repository, but it is going touch all scripts and lots of code....

This pull request boards up the broken window and allows us to choose a time to fix everything everywhere....

maliberty:
I'm not sure the cure is better than the disease.  The current is irregular but clear; your alternative is regular but unclear.  I find clarity more important.

I would expect ADDITIONAL_GDS to be a single file not a list of files based on the name.

maliberty:
I seems to have been an intentional choice not a "broken window" or a "misspelling"

oharboe:
> I seems to have been an intentional choice not a "broken window" or a "misspelling"

I guess I am weary of changing the interface to the ORFS. This name is not an implementation detail, it is part of the ORFS interface.

Changing the interface to improve naming is a separate concern from fixing a problem in the implementation. Hence, I would do this as two commits(one concern per commit) even if it was done as a single pull request. If I was a stickler for one concern per commit.

maliberty:
I don't want to weary you and appreciate you working on this.  I don't understand your comment about "fixing a problem in the implementation".  When I look at this PR I just see a rename.  What is the fixed problem?

maliberty:
I think I see what you are referring to.  The various platforms are out of sync:
```
platforms/asap7/config.mk\011:                                 $(ADDITIONAL_GDS)
platforms/asap7/config.mk\0142:																	  $(ADDITIONAL_GDS)
platforms/asap7/config.mk\0193:																	 $(ADDITIONAL_GDS)
platforms/nangate45/config.mk\013:                     $(ADDITIONAL_GDS)
platforms/sky130hd/config.mk\013:                     $(ADDITIONAL_GDS_FILES)
platforms/sky130hs/config.mk\013:                     $(ADDITIONAL_GDS_FILES)
platforms/sky130hd_fakestack/config.mk\013:                     $(ADDITIONAL_GDS_FILES)
platforms/gf180/config.mk\021:                                                $(ADDITIONAL_GDS)
```

maliberty (approved):
I finally get it.  We are already using ADDITIONAL_GDS in most places so this change is to use it everywhere.  I think a bit more explanation in the PR would have made this go a lot more smoothly.

We can take up the rename separately as you suggest.

oharboe:
> I finally get it. We are already using ADDITIONAL_GDS in most places so this change is to use it everywhere. I think a bit more explanation in the PR would have made this go a lot more smoothly.
> 
> We can take up the rename separately as you suggest.

Indeed. Will work on it. As I learn more I will be able to articulate better, there is a lot of guesswork still....

PR #710, opened 2022-12-21
update metrics for public designs after ppl change

eder-matheus:


maliberty (approved):

eder-matheus:
@maliberty swerv_wrapper nangate45 is still failing, but I think it happens in every run. Can you merge it even with this error?

maliberty:
I see riscv32i sky130hd is failing but not swerv_wrapper.  It was a day or two ago so I guess you have a slightly older branch point.  I'll go ahead and merge as this looks like it will fix riscv32i

PR #709, opened 2022-12-21
Reproduce via on pin in lef crash

oharboe:
To reproduce crashes associated with VIAs on PINs in LEFs, run the below and you get the following crash as on latest master:

```
make DESIGN_CONFIG=designs/asap7/uart-blocks/config.mk
[deleted]
Stack trace:
 0# 0x0000000000CC0AC7 in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 1# 0x00007F7C47054400 in /lib64/libc.so.6
 2# odb::dbTechLayer::getDirection() in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 3# pdn::InstanceGrid::getInstanceObstructions(odb::dbInst*, std::array<int, 4ul> const&) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
```

If you apply https://github.com/The-OpenROAD-Project/OpenROAD/pull/2652, the crash appears later in the flow:

```
[INFO GPL-0075] Routability numCall: 1 inflationIterCnt: 1 bloatIterCnt: 0
Signal 11 received
Stack trace:
 0# 0x0000000000CCB087 in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 1# 0x00007F1255BA6400 in /lib64/libc.so.6
 2# odb::dbTechLayer::getType() in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 3# grt::GlobalRouter::findInstancesObstructions(odb::Rect&, std::vector<int, std::allocator<int> > const&) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 4# grt::GlobalRouter::computeObstructionsAdjustments() in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 5# grt::GlobalRouter::applyAdjustments(int, int) in /OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad

```

PR #714, opened 2022-12-23
blocks: changed environment variables leaked through to sub-make invocation

oharboe:
@maliberty This fixes a problem where the top-level build fails when BLOCKS is used, but where the macros build fine when invoked separately.

This can fail before this fix:

```
make DESIGN=designs/sky130hd/foo/config.mk
```

Whereas, this works before the fix:

```
make DESIGN=designs/sky130hd/foo/bar/config.mk generate_abstract
make DESIGN=designs/sky130hd/foo/config.mk
```

This fix is tested as part of #709, however #709 won't be ready to be merged for some time and it is useful to have this in master before then to assist other testing.


maliberty (approved):
It looks like you need to merge master to fix the CI error

PR #725, opened 2022-12-28
asap7: large amount of fillers test

oharboe:


maliberty:
Is there a reason to add this as a flow test rather than just an asap7 issue?

oharboe:
> Is there a reason to add this as a flow test rather than just an asap7 issue?

This is a draft pull request that I intend to refer to from issues, if I can reproduce the problems I am seeing.

I havent gotten around to trying to create reproduction issues yet.

maliberty:
So there is no intention to ever merge this?  You can use make_issue to generate a reproducer without needing a PR.

oharboe:
> So there is no intention to ever merge this? You can use make_issue to generate a reproducer without needing a PR.

Iterating on the code still and need to find the time to articulate the case.

PR #726, opened 2022-12-28
asap7: small flip-flop memory test

oharboe:


maliberty:
Have you looked at dffram?  I am thinking of rewriting it in OR directly.

oharboe:
> Have you looked at dffram? I am thinking of rewriting it in OR directly.

No, thanks for the reference!

Note that for now, the problems/concerns are elsewhere in the design so this simple macro is enough to allow me to separate concerns and tackle them one at the time.

I don't actually expect I will be the one doing the SRAM, I'm just trying to put together a flow + CI/CD that is good 'nuf to use around the house for what we need.

Later we've got a long list of refinements to do that require domain knowledge...

maliberty:
I don't understand what you mean by "Note that for now, the problems/concerns are elsewhere in the design so this simple macro is enough to allow me to separate concerns and tackle them one at the time.".   What is goal here?

oharboe:
> I don't understand what you mean by "Note that for now, the problems/concerns are elsewhere in the design so this simple macro is enough to allow me to separate concerns and tackle them one at the time.". What is goal here?

Again a draft pr that I intend to see if I can refer to from issues.

oharboe:
Not needed to reproduce problems I've been seeing lately.

PR #749, opened 2023-01-03
sky130hd: Update PDN

antonblanchard:
The current sky130hd PDN is very dense, which heavily impacts the available routing resources on met4 and met5. Create a less dense PDN for larger designs, and use -snap_to_grid so the PDN ends up on track offsets.

### DIFF
--- a/flow/platforms/sky130hd/pdn.tcl
+++ b/flow/platforms/sky130hd/pdn.tcl
@@ -20,8 +20,31 @@ set_voltage_domain -name {CORE} -power {VDD} -ground {VSS}
 ####################################
 define_pdn_grid -name {grid} -voltage_domains {CORE}
 add_pdn_stripe -grid {grid} -layer {met1} -width {0.48} -pitch {5.44} -offset {0} -followpins
-add_pdn_stripe -grid {grid} -layer {met4} -width {1.600} -pitch {27.140} -offset {13.570}
-add_pdn_stripe -grid {grid} -layer {met5} -width {1.600} -pitch {27.200} -offset {13.600}
+
+set db [::ord::get_db]
+set dbu_per_uu [[$db getTech] getDbUnitsPerMicron]
+set block [[$db getChip] getBlock]
+set core_bbox [$block getCoreArea]
+set core_width [$core_bbox dx]
+set core_height [$core_bbox dy]
+
+set core_width [expr $core_width / $dbu_per_uu ]
+set core_height [expr $core_height / $dbu_per_uu ]
+
+# Use multiples of the met4 pitch (0.92um)
+if [ expr $core_height < 100 ] {
+    add_pdn_stripe -grid {grid} -layer {met4} -width {1.600} -pitch {27.600} -offset {13.800} -snap_to_grid
+} else {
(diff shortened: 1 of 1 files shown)


antonblanchard:
The exact values for the PDN are still being discussed, but this is a first pass.

maliberty:
Are you intending that this should be merged as is or is it pending on final values?  Whatever values are chosen should come with some quantitative justification.

maliberty:
@louiic @rovinski @gadfort  any opinions?

antonblanchard:
> Are you intending that this should be merged as is or is it pending on final values? Whatever values are chosen should come with some quantitative justification.

Happy to wait until we can agree on the final values. In the meantime, there's a few things we might want to discuss:
- In order to handle trivial designs (eg gcd), I needed two versions of the PDN, selected via some TCL. Does this look reasonable?
- The original sky130hd pdn had offsets of `pitch/2`. Most points on the chip will be `pitch/2` away from a power/ground rail, but right at the edges of the die it is going to be `pitch` distance away from one of the rails. Should we be using smaller offsets, or smaller spacing between power/ground?
sky130hd-pdn-offset

- I'm using `-snap_to_grid` but for some PDN wire widths it seems like we want the wire to snap to the half track. In the pic below the PDN strap (top wire) appears to obstruct 3 tracks, and if it was centered on the half track it would only obstruct 2 tracks. For comparison, the bottom wire is a min width signal wire:

sky130hd-met5-pdn

flow/platforms/sky130hd/pdn.tcl
```diff
@@ -20,8 +20,31 @@ set_voltage_domain -name {CORE} -power {VDD} -ground {VSS}
...
+
+set db [::ord::get_db]
+set dbu_per_uu [[$db getTech] getDbUnitsPerMicron]
+set block [[$db getChip] getBlock]
+set core_bbox [$block getCoreArea]
+set core_width [$core_bbox dx]
+set core_height [$core_bbox dy]
+
+set core_width [expr $core_width / $dbu_per_uu ]
+set core_height [expr $core_height / $dbu_per_uu ]
```
gadfort, line 32: You can use `ord::dbu_to_microns` instead

gadfort (commented):
I don`t see anything particularly wrong (assuming this will meet IR drop requirements).
Another possibility would be to reduce the width of the metals so they occupy fewer tracks (I would be somewhat concerned at a 75um pitch, but it appears to only be a few ohms so maybe that is okay).

rovinski:
In my opinion if the worst case IR drop is under 5% then it's completely fine. You can change the power grid pretty much however you want as long as it meets that.

With respect to the strap widths, in practice I only usually see widths of 1 or 3 tracks so that the straps are on-grid. In theory, properly aligned 2-track straps are ok but I just don't usually see them.

I believe that PDN grid sizing is a bit of an art because you just have to do trial and error to find one in the sweet spot.

maliberty:
@rovinski 5% under what assumption? The grid default is defined at the platform level not per design (and rarely overridden)

rovinski:
Perhaps across all designs in the platform? It's definitely the responsibility of the designer to check that the IR drop is ok, but there are so many variables and modeling limitations we have that I would say eyeballing is sufficient.

maliberty:
abandoned

Issue #732, opened 2022-12-30
ASAP7 has a single celler fill causing millions of instances to be created


oharboe:
### Subject

[Stage]: asap7 platform PDK problem. Clearly visible in 4_2_cts_fillcell.log

### Describe the bug

ASAP7 only has a single filler cell defined. This causes millions of instances to be created, which again causes problems for the OpenROAD GUI.

Reporting takes a really long time since the OpenROAD GUI is slow. The OpenROAD GUI is run to during reporting stage to generate some images.

### Expected Behavior

Multiple filler cell sizes, so that so many instances doesn't have to be created.

### Environment

```shell
Git commit: d297c5b18d08684b550e74a3115d4c366bd92e01
kernel: Linux 5.19.0-26-generic
os: Ubuntu 22.10 (Kinetic Kudu)
cmake version 3.24.2
gcc (Ubuntu 12.2.0-3ubuntu1) 12.2.0
Ubuntu clang version 15.0.2-1
```

### To Reproduce

Run asap7 design vs sky130hd, look at number of filler cells created. Many more filler cells in asap7.

Here is a design with a lot of filler cells to test with: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/725

### Relevant log output

```shell
From 4_2_cts_fillcell.log:

[INFO DPL-0001] Placed 9362608 filler instances.
```

vijayank88:
There are two filler cells available in ASAP7.
```
FILLER_ASAP7_75t_R
FILLERxp5_ASAP7_75t_R
```
Tapcell list is:
```
TAPCELL_ASAP7_75t_R
TAPCELL_WITH_FILLER_ASAP7_75t_R
```
With `FILLERxp5_ASAP7_75t_R` logs for design `gcd`
```
[INFO DPL-0001] Placed 10268 filler instances.
```

By replacing `FILLER_ASAP7_75t_R` the logs as follows for design `gcd`
```
[INFO DPL-0001] Placed 3576 filler instances.
```

oharboe:
@vijayank88 I tried:

```
export FILL_CELLS              = FILLER_ASAP7_75t_R FILLERxp5_ASAP7_75t_R
export TAP_CELL_NAME           = TAPCELL_ASAP7_75t_R TAPCELL_WITH_FILLER_ASAP7_75t_R
```

It didn't change the number of filler instances:

```
[...]

PR #761, opened 2023-01-09
asap7: add decap cells as FILL_CELLS

maliberty:
Per Lawrence Clark this is the intention for asap7

Signed-off-by: Matt Liberty <mliberty@eng.ucsd.edu>

### DIFF
--- a/flow/platforms/asap7/config.mk
+++ b/flow/platforms/asap7/config.mk
@@ -102,7 +102,14 @@ export CTS_BUF_CELL            ?= BUFx4_ASAP7_75t_R
 export CTS_BUF_DISTANCE        ?= 60
 
 # Fill cells used in fill cell insertion
-export FILL_CELLS              = "FILLERxp5_ASAP7_75t_R"
+export FILL_CELLS              = \
+         FILLERxp5_ASAP7_75t_R \
+         FILLER_ASAP7_75t_R \
+         DECAPx1_ASAP7_75t_R \
+         DECAPx2_ASAP7_75t_R \
+         DECAPx4_ASAP7_75t_R \
+         DECAPx6_ASAP7_75t_R \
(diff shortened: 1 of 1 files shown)


maliberty:
@oharboe hopefully this will help/fix #732

vvbandeira (approved):

rovinski:
I question this - decap cells will be inserted everywhere because they are the largest fill cell. Over-insertion of decap cells is usually bad for leakage power.

maliberty:
@rovinski  Since asap7 isn't manufacturable I'm not sure it matters.  Real PDKs have a full range of fill cells.  OL does the same in sky130 in production but maybe leakage matters less there?

maliberty:
From Prof. Clark
```
On the final note, don't fill with fill cells. Use DECAP! That is what it is for. Fill is for the tiny spots left over that decap does not fit.
```

rovinski:
This is unfortunately exposing some shortcomings of ASAP7. Assuming DECAP cells actually have caps in them and aren't just poorly named, they shouldn't be used everywhere. These caps are leaky and contribute significantly to leakage power. Unfortunately, we can't even model this because there aren't .lib views of DECAP cells. IIRC commercial PDKs normally do have lib views just to model the leakage power.

So I guess we just have to accept two wrongs making a right.

oharboe:
> This is unfortunately exposing some shortcomings of ASAP7. Assuming DECAP cells actually have caps in them and aren't just poorly named, they shouldn't be used everywhere. These caps are leaky and contribute significantly to leakage power. Unfortunately, we can't even model this because there aren't .lib views of DECAP cells. IIRC commercial PDKs normally do have lib views just to model the leakage power.
> 
> So I guess we just have to accept two wrongs making a right.

Maybe create a list in the documentation of known "approximations" made by ASAP7 and move on?

PR #771, opened 2023-01-16
add table in docs for build options locally and with docker

habibayassin:


vvbandeira (changes requested):
We should not have the same text on two locations as it is harder to keep both updated. Can we refactor/change the organization of the build instructions to avoid this?

docs/user/Build.md
```diff
@@ -0,0 +1,54 @@
...
+least one of the new version of GCC/Clang. To enable GCC-8 or Clang-7 you
+need to run:
+
+```shell
+# to enable gcc-8
+source /opt/rh/devtoolset-8/enable
+# or to enable clang-7
+source /opt/rh/llvm-toolset-7.0/enable
+```
+:::
```
vvbandeira: image

The warning marker does not show well on GH.

docs/user/BuildWithDocker.md
```diff
@@ -7,29 +7,27 @@ For this method you only need to install
...
+> ```
 
 You can restrict the number of CPUs with the `-t|--threads N` argument:
 
 ``` shell
 ./build_openroad.sh --threads N
 ```
-
----
+:::
```
vvbandeira: Remove this and I think it is good to be merged. Thanks!

vvbandeira (approved):

PR #778, opened 2023-01-19
github: add devcontainer configuration

oharboe:
@maliberty The idea is that the community can collaborate on this configuration such that all that is required to test and develop OpenROAD-flow-scripts is to fire up a github codespaces from any brows.

This is change is just meant to be an initial seed.

Further down the road, improvements to this config will make it faster and easier to set up OpenROAD-flow-scripts.

Rough idea for use-case:

1. Click on "code" in github and create github codespace
2. Run: build_openroad.sh --local (this will fail, but bring in scripts we need in the next step)
4. Run: sudo tools/OpenROAD/etc/DependencyInstaller.sh, then restart build_openroad.sh --local
5. . setup_env.sh
6. To create a launch configuraiton in tools/OpenROAD/.vscode/launch.settings, run: make DESIGN_CONFIG=designs/asap7/aes/config.mk vscode_debug_pdn
7. Run: code tools/OpenROAD
8. hit f7 to build
9. Set breakpoint, press f5, wait for the debugger to hit the breakpoint for the failing stage in step 5 above.

I didn't try, but it should be possible to run OpenROAD GUI, just, too with this setup: https://babyprogrammer.com/blog/running-gui-code-applications-in-github-codespaces-on-windows/

image

image

image

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,5 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-22.04",
+  "features": {
+  }
+}


maliberty:
This sounds cool.  @vvbandeira please take a look.

maliberty:
@proppy @quantamhd FYI

maliberty:
We do have a container that the CI uses to build with the dependencies installed.  Perhaps that could be used.

oharboe:
> We do have a container that the CI uses to build with the dependencies installed. Perhaps that could be used.

The advantage of having a container published to the github container repository is that you can be up and running with a full dev,  class-room, or conference demo, environment in 60 seconds... No need to install anything.

proppy:
A few other options if we wanted to prepopulate the container image with openroad and other ORFS dependencies:
- we could reuse the container images at https://hdl.github.io/containers/ToolsAndImages.html
- we could reuse the conda packages published at https://hdl.github.io/conda-eda/

oharboe:
> A few other options if we wanted to prepopulate the container image with openroad and other ORFS dependencies:
> 
> * we could reuse the container images at https://hdl.github.io/containers/ToolsAndImages.html
> * we could reuse the conda packages published at https://hdl.github.io/conda-eda/

@proppy What should I do with this PR? It is just an example/idea. My thinking was that an initial seed could be merged and then if pull requests are forthcoming this could grow into a nice feature.

My initial experience is that the installation process takes at least an hour on the smallest machine. Everything seemed to work, but deleted the codespace after 1 hour worth of installing/compiling.

oharboe:
@maliberty @proppy The idea was put forth. After my experience, I think for this to be useful for interviews, classrooms, presentations, etc. there needs to be a docker image that allows getting started with the smallest github virtual machine in minutes. Running a regular install with the weakest machine takes at least an hour.

maliberty:
I'm not sure why this was closed.

oharboe:
@maliberty it's three lines of code  + some notes on my experience :-) 

The next thing that is needed, for this to be effective is to have a suitable github image(nightly?) published to the github repository, then point to that image instead of the image I point to.

maliberty:
It probably would have been better as a discussion rather than a PR but there isn't an easy way to migrate so let's just use this for that purpose.

oharboe:
> It probably would have been better as a discussion rather than a PR but there isn't an easy way to migrate so let's just use this for that purpose.

github doesn't really distinguish much between issues/discussions and pull requests....

maliberty:
It looks like the devcontainer could just reference docker/Dockerfile.dev.  We would then need to enable [prebuilds](https://docs.github.com/en/codespaces/prebuilding-your-codespaces/about-github-codespaces-prebuilds) automatically get updated images for code changes.

maliberty:
> github doesn't really distinguish much between issues/discussions and pull requests....

Numerically that's true but they go in different areas and have different work flows.  It is worth distinguishing them for that reason.

oharboe:
@maliberty There's nothing in this PR and the idea is recorded in the history of this closed PR.

PR #791, opened 2023-01-27
detailed route: add DETAILED_ROUTE_ARGS for debugging purposes

oharboe:
DETAILED_ROUTE_ARGS is used when debugging detailed, route, e.g. append
"-droute_end_iter 5" to look at routing violations after only 5 iterations,
speeding up iterations on a problem where detailed routing doesn't converge
or converges slower than expected.

### DIFF
--- a/flow/scripts/detail_route.tcl
+++ b/flow/scripts/detail_route.tcl
@@ -40,11 +40,19 @@ if { [info exists ::env(REPAIR_PDN_VIA_LAYER)]} {
   append additional_args " -repair_pdn_vias $::env(REPAIR_PDN_VIA_LAYER)"
 }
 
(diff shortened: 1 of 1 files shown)


maliberty (approved):

flow/scripts/detail_route.tcl
```diff
@@ -40,11 +40,19 @@ if { [info exists ::env(REPAIR_PDN_VIA_LAYER)]} {
   append additional_args " -repair_pdn_vias $::env(REPAIR_PDN_VIA_LAYER)"
 }
 
+append additional_args "-save_guide_updates -verbose 1"
```
maliberty: you need a leading space to avoid

-disable_via_gen-save_guide_updates

PR #811, opened 2023-02-06
designs/src/mock-big-array: fix gaffes in Chisel code

oharboe:
this only changes the generated verilog when
the array is not ```n*n```, but ```m*n``` where ```m != n```.


oharboe:
@maliberty Nag. Ready to go in, small bugfix if Verilog needs to be regenerated.

maliberty:
You should request a review when a PR is ready

maliberty (approved):

PR #814, opened 2023-02-09
asap7: AREA constraints are 4x too large

antonblanchard:
When creating the 1x tech LEF, it looks like the area constraints were only reduced by 4x and not 16x.


antonblanchard:
I was looking at areas of an asap7 design that drt was struggling to route, and noticed quite large patches. Looking at the ASAP7 design rules:

m1-area

The adjusted area doesn't quite match the design rules, and I think it's because of:

```
  #  MINSIZE is set so that the minimum lenght of a segment is 36nm. At the proper rule size of
  #  31nm, the lines can't be minimum space. This causes DRCs (like crazy). Same for M3
  #  MINSIZE 0.112 0.072 ; 
  # area is adjusted to match this (Nanoroute requires both AREA and MINSIZE)

 AREA 0.000666 ;
 MINSIZE 0.037 0.018 ;
```

Is this a peculiarity of Nanoroute? Should we adjust the area to match the design rules, and remove the minsize constraint?

antonblanchard:
I also can't find any AREA constraints for M4-M7 in the design rules, even though they are in the tech lef.

maliberty:
My guess is the missing min area for m4-7 is a DRM bug.

maliberty:
I'm confused by your description.  We don't patch on m1 afaik so I'm not sure why the minarea rule matters there.  The other quote is from m2 and I don't see the issue there (the drm doesn't have a rule and this value doesn't seem obviously wrong).

antonblanchard:
> I'm confused by your description. We don't patch on m1 afaik so I'm not sure why the minarea rule matters there. The other quote is from m2 and I don't see the issue there (the drm doesn't have a rule and this value doesn't seem obviously wrong).

M2 and M3 should use the same area as M1 I think:

m1-3-drm

maliberty:
Thanks I think I finally get what you are saying.  However I do see

image

and in the 1x lef
```
LAYER M8
 AREA 7.52 ;
```

which doesn't seem to follow the pattern.  (it also seems off by 10^3)

antonblanchard:
Good catch, that seems very wrong. In the original 4x LEF: 

```
LAYER M7
  TYPE ROUTING ;
  DIRECTION VERTICAL ;
  PITCH 0.256 ; 
  WIDTH 0.128 ;
  SPACING 0.128 ;
  AREA 0.035 ;   # Areas still need tweaking

...

LAYER M8
  TYPE ROUTING ;
  DIRECTION HORIZONTAL ;
  PITCH 0.32 0.32 ;
  WIDTH 0.16 ;
  AREA 30.08 ;

...

LAYER M9
  TYPE ROUTING ;
  DIRECTION VERTICAL ;
  PITCH 0.32 0.32 ;
  WIDTH 0.16 ;
  AREA 30.08 ;

```

So M8 and M9 AREA does seem very wrong. I'll update this patch and submit a patch to ASAP7.

maliberty:
I'm not sure what to make of the intervening layers since they have no DRM entries.

maliberty (approved):

PR #820, opened 2023-02-13
nangate45 bp_be_top drc fix

vijayank88:
Signed-off-by: vijayank88 <paruthi143@gmail.com>
With latest OpenROAD head `nangate45/bp_be_top` design facing DRC error end of detail routing.
```
[ERROR] detailedroute__route__drc_errors fail test: 3.0 <= 0.0
```
fixes #721 `nangate45/bp_be` metadata check failure.


vvbandeira:
Nothing to comment on the change itself. @maliberty to approve, also should we create a Jira to investigate the root cause of the router needing the extra halo spacing or is this already being tracked somewhere?

maliberty (approved):

maliberty:
This is ok as a temporary solution but we should investigate.  @vvbandeira go ahead with a Jira and assign to @osamahammad21

PR #824, opened 2023-02-14
designs/asap7/mock-array-big: add simple combinational path through E…

oharboe:
@yupferris This allows us to study timing constraint files for a wire running through the macro.

Later there might be some very simple logic contributing, to the lsb being forwarded through the macros to the right edge of the top level macro

### DIFF
--- a/flow/designs/src/mock-array-big/Element.v
+++ b/flow/designs/src/mock-array-big/Element.v
@@ -7,7 +7,23 @@ module Element(
   output [7:0] io_outs_0,
   output [7:0] io_outs_1,
   output [7:0] io_outs_2,
-  output [7:0] io_outs_3
+  output [7:0] io_outs_3,
+  input        io_lsbIn_0,
+  input        io_lsbIn_1,
+  input        io_lsbIn_2,
+  input        io_lsbIn_3,
+  input        io_lsbIn_4,
+  input        io_lsbIn_5,
+  input        io_lsbIn_6,
+  input        io_lsbIn_7,
+  output       io_lsbOut_0,
+  output       io_lsbOut_1,
+  output       io_lsbOut_2,
+  output       io_lsbOut_3,
+  output       io_lsbOut_4,
+  output       io_lsbOut_5,
+  output       io_lsbOut_6,
+  output       io_lsbOut_7
 );
 `ifdef RANDOMIZE_REG_INIT
   reg [31:0] _RAND_0;
@@ -15,19 +31,27 @@ module Element(
   reg [31:0] _RAND_2;
   reg [31:0] _RAND_3;
 `endif // RANDOMIZE_REG_INIT
-  reg [7:0] REG; // @[MockArray.scala 33:42]
-  reg [7:0] REG_1; // @[MockArray.scala 33:42]
-  reg [7:0] REG_2; // @[MockArray.scala 33:42]
-  reg [7:0] REG_3; // @[MockArray.scala 33:42]
-  assign io_outs_0 = REG; // @[MockArray.scala 33:13]
-  assign io_outs_1 = REG_1; // @[MockArray.scala 33:13]
-  assign io_outs_2 = REG_2; // @[MockArray.scala 33:13]
-  assign io_outs_3 = REG_3; // @[MockArray.scala 33:13]
+  reg [7:0] REG; // @[MockArray.scala 36:42]
+  reg [7:0] REG_1; // @[MockArray.scala 36:42]
+  reg [7:0] REG_2; // @[MockArray.scala 36:42]
+  reg [7:0] REG_3; // @[MockArray.scala 36:42]
+  assign io_outs_0 = REG; // @[MockArray.scala 36:13]
+  assign io_outs_1 = REG_1; // @[MockArray.scala 36:13]
+  assign io_outs_2 = REG_2; // @[MockArray.scala 36:13]
+  assign io_outs_3 = REG_3; // @[MockArray.scala 36:13]
+  assign io_lsbOut_0 = io_lsbIn_1; // @[MockArray.scala 40:15]
+  assign io_lsbOut_1 = io_lsbIn_2; // @[MockArray.scala 40:15]
+  assign io_lsbOut_2 = io_lsbIn_3; // @[MockArray.scala 40:15]
+  assign io_lsbOut_3 = io_lsbIn_4; // @[MockArray.scala 40:15]
+  assign io_lsbOut_4 = io_lsbIn_5; // @[MockArray.scala 40:15]
+  assign io_lsbOut_5 = io_lsbIn_6; // @[MockArray.scala 40:15]
+  assign io_lsbOut_6 = io_lsbIn_7; // @[MockArray.scala 40:15]
+  assign io_lsbOut_7 = io_outs_0[0]; // @[MockArray.scala 40:54]
   always @(posedge clock) begin
-    REG <= io_ins_3; // @[MockArray.scala 33:42]
-    REG_1 <= io_ins_2; // @[MockArray.scala 33:42]
-    REG_2 <= io_ins_1; // @[MockArray.scala 33:42]
-    REG_3 <= io_ins_0; // @[MockArray.scala 33:42]
+    REG <= io_ins_3; // @[MockArray.scala 36:42]
+    REG_1 <= io_ins_2; // @[MockArray.scala 36:42]
+    REG_2 <= io_ins_1; // @[MockArray.scala 36:42]
+    REG_3 <= io_ins_0; // @[MockArray.scala 36:42]
   end
 // Register and memory initialization
 `ifdef RANDOMIZE_GARBAGE_ASSIGN
--- a/flow/designs/asap7/mock-array-big/Element/io.tcl
+++ b/flow/designs/asap7/mock-array-big/Element/io.tcl
@@ -7,6 +7,7 @@ proc pin1 {fmt p} {
 }
 
 set data_width [expr {[info exists ::env(MOCK_ARRAY_DATAWIDTH)] ? $::env(MOCK_ARRAY_DATAWIDTH) : 8}]
+set cols [expr {[info exists ::env(MOCK_ARRAY_WIDTH)] ? $::env(MOCK_ARRAY_WIDTH) : 8}]
 
 set assignments [list \
     top bottom \
@@ -26,6 +27,13 @@ set assignments [list \
     [ concat \
         {*}[pin1 {io_outs_1[%d]} $data_width] \
         {*}[pin1 {io_ins_1[%d]} $data_width] \
+    ]] \
+    left right \
+    [list [ concat \
+        {*}[pin1 {io_lsbIn_%d} $cols] \
+    ] \
+    [ concat \
+        {*}[pin1 {io_lsbOut_%d} $cols] \
     ]] \
 ]
 
--- a/flow/designs/asap7/mock-array-big/io.tcl
+++ b/flow/designs/asap7/mock-array-big/io.tcl
@@ -8,6 +8,14 @@ proc pin2 {fmt p q} {
     return $result
 }
 
+proc pin1 {fmt p} {
+    set result [list]
+    for {set m 0} {$m < $p} {incr m} {
+        lappend result [format $fmt $m]
+    }
+    return $result
+}
+
 set data_width [expr {[info exists ::env(MOCK_ARRAY_DATAWIDTH)] ? $::env(MOCK_ARRAY_DATAWIDTH) : 8}]
 set rows [expr {[info exists ::env(MOCK_ARRAY_HEIGHT)] ? $::env(MOCK_ARRAY_HEIGHT) : 8}]
 set cols [expr {[info exists ::env(MOCK_ARRAY_WIDTH)] ? $::env(MOCK_ARRAY_WIDTH) : 8}]
@@ -32,6 +40,7 @@ set assignments [list \
     [ concat \
         {*}[pin2 {io_insVertical_0_%d[%d]} $rows $data_width] \
         {*}[pin2 {io_outsVertical_0_%d[%d]} $rows $data_width] \
+        {*}[pin1 {io_lsbs_%d} $cols] \       
     ] \
 ]
 
--- a/flow/designs/src/mock-array-big/MockArray.v
+++ b/flow/designs/src/mock-array-big/MockArray.v
@@ -130,583 +130,1607 @@ module MockArray(
   output       io_lsbs_62,
   output       io_lsbs_63
 );
-  wire  ces_0_0_clock; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_0_io_ins_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_0_io_ins_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_0_io_ins_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_0_io_ins_3; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_0_io_outs_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_0_io_outs_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_0_io_outs_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_0_io_outs_3; // @[MockArray.scala 36:52]
-  wire  ces_0_1_clock; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_1_io_ins_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_1_io_ins_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_1_io_ins_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_1_io_ins_3; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_1_io_outs_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_1_io_outs_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_1_io_outs_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_1_io_outs_3; // @[MockArray.scala 36:52]
-  wire  ces_0_2_clock; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_2_io_ins_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_2_io_ins_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_2_io_ins_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_2_io_ins_3; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_2_io_outs_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_2_io_outs_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_2_io_outs_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_2_io_outs_3; // @[MockArray.scala 36:52]
-  wire  ces_0_3_clock; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_3_io_ins_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_3_io_ins_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_3_io_ins_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_3_io_ins_3; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_3_io_outs_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_3_io_outs_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_3_io_outs_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_3_io_outs_3; // @[MockArray.scala 36:52]
-  wire  ces_0_4_clock; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_4_io_ins_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_4_io_ins_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_4_io_ins_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_4_io_ins_3; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_4_io_outs_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_4_io_outs_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_4_io_outs_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_4_io_outs_3; // @[MockArray.scala 36:52]
-  wire  ces_0_5_clock; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_5_io_ins_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_5_io_ins_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_5_io_ins_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_5_io_ins_3; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_5_io_outs_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_5_io_outs_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_5_io_outs_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_5_io_outs_3; // @[MockArray.scala 36:52]
-  wire  ces_0_6_clock; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_6_io_ins_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_6_io_ins_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_6_io_ins_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_6_io_ins_3; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_6_io_outs_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_6_io_outs_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_6_io_outs_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_6_io_outs_3; // @[MockArray.scala 36:52]
-  wire  ces_0_7_clock; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_7_io_ins_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_7_io_ins_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_7_io_ins_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_7_io_ins_3; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_7_io_outs_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_7_io_outs_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_7_io_outs_2; // @[MockArray.scala 36:52]
-  wire [7:0] ces_0_7_io_outs_3; // @[MockArray.scala 36:52]
-  wire  ces_1_0_clock; // @[MockArray.scala 36:52]
-  wire [7:0] ces_1_0_io_ins_0; // @[MockArray.scala 36:52]
-  wire [7:0] ces_1_0_io_ins_1; // @[MockArray.scala 36:52]
-  wire [7:0] ces_1_0_io_ins_2; // @[MockArray.scala 36:52]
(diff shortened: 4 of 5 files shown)


maliberty:
DCO needs fixing, see the Details link

oharboe:
@maliberty Ready for review.

maliberty:
Are you talking about a combinational path containing logic or a pure wire feedthrough?  The later has a lot of tricky implications for timing.

maliberty:
> @maliberty Ready for review.

You can request a review in 'reviewers' in the upper right

flow/designs/src/mock-array-big/Element.v
```diff
@@ -7,27 +7,51 @@ module Element(
...
-  assign io_outs_3 = REG_3; // @[MockArray.scala 33:13]
+  reg [7:0] REG; // @[MockArray.scala 36:42]
+  reg [7:0] REG_1; // @[MockArray.scala 36:42]
+  reg [7:0] REG_2; // @[MockArray.scala 36:42]
+  reg [7:0] REG_3; // @[MockArray.scala 36:42]
+  assign io_outs_0 = REG; // @[MockArray.scala 36:13]
+  assign io_outs_1 = REG_1; // @[MockArray.scala 36:13]
+  assign io_outs_2 = REG_2; // @[MockArray.scala 36:13]
+  assign io_outs_3 = REG_3; // @[MockArray.scala 36:13]
+  assign io_lsbOut_0 = io_lsbIn_1; // @[MockArray.scala 40:15]
```
maliberty, line 42: I would recommend you either buffer the feedthrough or do over the block routing.  Pushing a wire only feedthrough down will no lead to good timing modeling as you can't build a model for a wire segment.

maliberty: Perhaps buffer_inputs will actually put in a buffer for you.  Have you looked at the generated result?

oharboe:
> > @maliberty Ready for review.
> 
> You can request a review in 'reviewers' in the upper right

I dont have the access rights.

maliberty:
> > > @maliberty Ready for review.
> > 
> > 
> > You can request a review in 'reviewers' in the upper right
> 
> I dont have the access rights.

I didn't realize that.

flow/designs/src/mock-array-big/Element.v
yupferris: We're missing some conceptual understanding here, so I'd like to try and reduce scope even further and ask some questions that I hope can help us target our efforts more effectively.

Consider a design called `line` consisting of two instances, side-by-side, of a module which we'll call `element`. Since `element` is instantiated multiple times within a `line` instance, we want to make that a macro. Similarly, `line` may also be replicated further up in the hierarchy, so we want to make a macro out of that as well. Through a `line`, we'd like to route a bus, which we'll call `x`. This is illustrated below:

line

Note that `x` forms a fully-combinational path (which may have additional inputs from registers inside `element`, not shown/considered here) threaded through `line` and its `element` instances. I've added registers on both sides of the bus outside of `line` to illustrate that the path is eventually registered in the design hierarchy, but that these registers are outside of these particular macros.

Additionally, we assume that had we run idealized tools on a flattened design hierarchy, the design would be able to meet timing.

The following questions arise:

1. Do we need to specify timing constraints for each macro in the hierarchy?
   - I assume *yes*, since the modules may also contain registered logic, and the tools should be aware of the clock and target clock rate when hardening each macro, if nothing else to achieve good QOR.
2. Do we need to specify timing constraints for each (non-clock, non-power) IO for each macro in the hierarchy?
   - Naively I would assume *no*, because `line` and `element` are both macros which will be instantiated inside other macros (eg. not chip-level IOs), and as far as I understand, `.lib` files have the ability to express timing delays between arbitrary IOs, so a hardened macro could fully capture the relevant delays required for hardening macros further up in the hierarchy. However, we may _want_ to specify constraints for these IOs, which may help the tools achieve better placement/timing overall for the design.
3. If we have/want to specify timing constraints for all IOs for each macro, what is required to support partially/fully-combinational paths?
   - For example, the delay for the path `line_0.x_in -> element_0.x_in` is clearly different than the delay for the path `line_0.x_in -> element_1.x_in`, which suggests that however we specify delays for `x_in`/`x_out`, it must be conservative enough to capture a wide range of possible delays, and I don't see how that can be realizable when the tools can basically only insert delay cells to add delay along a path AFAIK.

Coming back to what you said:

> I would recommend you either buffer the feedthrough or do over the block routing. Pushing a wire only feedthrough down will [not] lead to good timing modeling as you can't build a model for a wire segment.

I assume the issue here is basically what I've described under question 3 above. As for the suggested workarounds, I'm not sure how to explicitly buffer the path assuming it's only a wire. Over-the-block routing should work, again assuming just a wire. However, what we're actually interested in is having some combinational logic along the path in `element` (and potentially `line` as well), so the pure-wire path added above might not be sufficient for discussion. It's of course possible to have that combinational logic outside `element`, but seeing as it's the same for each `element`, it seems like it would be ideal to contain that logic inside the `element` macro.

Any insight/corrections to the above is appreciated. Let me know if I can make any of this more clear.

maliberty: @tspyrou please responsed to this

oharboe:
@torerams Have a look at this one

flow/designs/src/mock-array-big/Element.v
rovinski: > 1. Do we need to specify timing constraints for each macro in the hierarchy?

You should always constrain all paths where you care about the timing. A clock constraint will constrain reg->reg paths, but you will need to add input and output delays to properly constrain in->reg, reg->out, and in->out paths. These input and output delays will depend on your manual time partitioning (i.e. how much time you are allocating inside the block vs. outside the block).

> 2. Do we need to specify timing constraints for each (non-clock, non-power) IO for each macro in the hierarchy?

These constraints _do_ need to be specified somewhere. These signals _will_ ultimately be constrained somewhere unless your chip is completely asynchronous. If you do not constrain your in->out signals at the macro-level, there is no way to hit the constraint at the top-level. I mean, yes, the timing will be recorded when you export the macro `.lib`, but for all intents and purposes, this should be constrained otherwise you almost certainly get bogus numbers.

> as far as I understand, `.lef` files have the ability to express timing delays between arbitrary IOs

`.lib` files. `.lef` is only physical information.

> 3. If we have/want to specify timing constraints for all IOs for each macro, what is required to support partially/fully-combinational paths?

In your example, you basically need 3 sets of constraints:

1. Constrain the clock, inputs, and outputs when building `element`
2. Constrain the clock, inputs, and output when building `line`
3. Constrain the clock, inputs, and output when building the top-level

This needs to be an iterative process. You need to estimate how much time should be spent inside `element` and how much time should be spent outside `element`, then set the macro's constraints accordingly.
Then, you need to export the timing from the `element` macro and start iterating on the timing constraints for `line`. If the timing constraints are not possible to meet, you need to potentially adjust the timing for `element`.

This is the basic flow for any kind of hierarchical timing closure.

>    * For example, the delay for the path `line_0.x_in -> element_0.x_in` is clearly different than the delay for the path `line_0.x_in -> element_1.x_in`, which suggests that however we specify delays for `x_in`/`x_out`, it must be conservative enough to capture a wide range of possible delays, and I don't see how that can be realizable when the tools can basically only insert delay cells to add delay along a path AFAIK.

I'm not sure I follow what you are saying. If `x_in` at the top level takes e.g. 1.0 ns, then you simply need to specify inside the `line` constraints `set_input_delay 1.0 -clock <CLOCK>` where `<CLOCK>` is the reference clock. You are specifying the delay on the signal input, not the required time from source to a destination. The rest of the timing is automatically derived. If you specify the input delay of `line_0.x_in`, the output delay of `line_0.x_out`, and the period of the clock which presumably connects them all, then all of the timing constraints are inferred from there.

>However, what we're actually interested in is having some combinational logic along the path in `element` (and potentially `line` as well), so the pure-wire path added above might not be sufficient for discussion. 

This means that the design in the current PR is not actually representative of what you are trying to do.

As a note of advice from personal experience, chaining all of the feedthrough (in->out) paths you have like this is a recipe for disaster. Every time you export a design as a macro, there is some loss in timing accuracy. So if you chain many macros together like this, there is accumulated loss in model accuracy and could potentially break the chip (or at least make it run much slower than expected). Converting these paths to clocked paths can substantially increase the reliability. Another option is to change the grouping of your hierarchy so there isn't as much loss in precision. If your design is extremely large, you will have to worry about clock synchronization because balancing large clock trees is difficult.

Source: I taped out tiled manycore designs with 400+ cores.

yupferris: Thanks for the thoughtful response, much appreciated! Some follow-up:

> `.lib` files. `.lef` is only physical information.

Indeed, thanks for the correction. I'll fix my comment.

> I'm not sure I follow what you are saying. If `x_in` at the top level takes e.g. 1.0 ns, then you simply need to specify inside the `line` constraints `set_input_delay 1.0 -clock <CLOCK>` where `<CLOCK>` is the reference clock. You are specifying the delay on the signal input, not the required time from source to a destination. The rest of the timing is automatically derived. If you specify the input delay of `line_0.x_in`, the output delay of `line_0.x_out`, and the period of the clock which presumably connects them all, then all of the timing constraints are inferred from there.

Most of this makes sense to me, but the part I'm missing from the above (and perhaps failed to articulate clearly in my question) is how to specify input/output delay for `element`. For example, `element_0.x_in` and `element_1.x_in` would require different delays, right? But both would have to be captured in the timing constraints for the `element` macro? I'm not seeing how that's possible.

> As a note of advice from personal experience, chaining all of the feedthrough (in->out) paths you have like this is a recipe for disaster. Every time you export a design as a macro, there is some loss in timing accuracy. So if you chain many macros together like this, there is accumulated loss in model accuracy and could potentially break the chip (or at least make it run much slower than expected). Converting these paths to clocked paths can substantially increase the reliability. Another option is to change the grouping of your hierarchy so there isn't as much loss in precision. If your design is extremely large, you will have to worry about clock synchronization because balancing large clock trees is difficult.

Right, this makes sense, thanks for the tip. We'll have to consider alternatives as well.

tspyrou: @yupferris @oharboe I think that its good to take a step back. Adding hierarchical implementation to a flow adds complexity to the timing analysis and to the implementation. The feedthroughs and constraints in the blocks mentioned are good examples. Its best to use hierarchy sparingly, as high up in the hierarchy as possible to overcome long runtimes. Creating a model whenever a module is used multiple times will be unwieldy and also removes optimization flexibility for that module to be optimized based on its context.

tspyrou: @yupferris @oharboe In the current implementation of write_timing_model its best to avoid making a block model for blocks which have timing exceptions on nodes in the block. Those exceptions cannot be captured in the model yet. Each model that is created will have to be qa'd. We do have some work going on to automate this process. The model can't be trusted until its timing is compared to the timing of the block from which it was derived and that accuracy is ensured.

tspyrou:
@openroadie This will be a good testcase for the model QA code.

maliberty:
I have pointed two GSoC potential contributors at this.  @tspyrou

flow/designs/src/mock-array-big/Element.v
rovinski: > Most of this makes sense to me, but the part I'm missing from the above (and perhaps failed to articulate clearly in my question) is how to specify input/output delay for `element`. For example, `element_0.x_in` and `element_1.x_in` would require different delays, right? But both would have to be captured in the timing constraints for the `element` macro? I'm not seeing how that's possible.

There are a few different ways of handling this.

One option is to set the constraints on `element` conservatively enough so that each instantiation (`element_0` and `element_1`) and their interfaces can meet their timing targets. If there isn't that much difference in path delay between `line_0.x_in` and the net connecting `element_0` to `element_1`, this is a fine option.

* Pros: Easier to implement and constrain
* Cons: Can be too conservative and reduce performance. May not work if the timing is too different.

The other option is to create a different block with new constraints. You currently have a module `element` constrained by `constraint.sdc` and two instantiations of it: `element_0` and `element_1`. The new method would be to create a second module e.g. `element_first` which is a copy of the source code from `element`. This module, however, is going to be constrained by `constraint_first.sdc`, which has the constraints necessary for the first element instantiation in `line_0`. This way, you can have the constraints you need applied to each block.

* Pros: More detailed and can better tune for performance
* Cons: More work (both from engineering and tool side) because you now have two blocks to implement instead of one.

As an aside, it may seem like a good idea to simply implement the same module twice with two different constraints, rather than creating a copy module. This will _not_ work, because the modules will have the same name but different implementations. Modules and the blocks must have distinct names so that your chip can pass LVS checks.

------

The above is to try to address your question, but I also very much agree with @tspyrou. ASIC tools in general - not just OpenROAD - are a balancing act of human effort vs. automation effort. The tools can solve these problems for you automatically, but you have to give them full control (i.e. no manual hierarchy) and deal with potentially very slow runtimes. On the other hand, you can take manual control and partition designs yourself which can save greatly on runtime, but at the cost of optimization _and_ a significant amount of expertise.

yupferris: Thanks again for info/tips. This is making sense. Lots to consider on how to move forward.

b224hisl:
@oharboe Hello, I'm new of openroad,  I have some questions about integrating macros:
1. when you harden the `element`, how you generate its lib/lef ? Because by default, openroad only generates gds in the `result` dir
2. how to invoke the process of hardening `element` in the process of hardening the whole array

maliberty:
The key is the BLOCKS statement at https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/888926fccf862d00b9c847b87ed59be1df3b8699/flow/designs/asap7/mock-array-big/config.mk#L16

oharboe:
> The key is the BLOCKS statement at
> 
> https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/888926fccf862d00b9c847b87ed59be1df3b8699/flow/designs/asap7/mock-array-big/config.mk#L16

@b224hisl To demystify that BLOCKS statement, all it does is to automatically invoke the command below *if* there is no .gds/.lib/.lef file:

So if you modify e.g. the Verilog files for the macro in BLOCKS, you need to either delete the .gds/.lib/.lef files, or invoke the command below:

```
make DESIGN_CONFIG=designs/asap7/mock-array-big/element/config.mk generate_abstract
```

b224hisl:
Thanks, But how you generate `block.lef` and `block.lib`. I harden your `element` first, but in the result directory, I didn't see the two files.  Are they necessary for integrating macros to the whole design?

oharboe:
> Thanks, But how you generate `block.lef` and `block.lib`. I harden your `element` first, but in the result directory, I didn't see the two files. Are they necessary for integrating macros to the whole design?

You can just run the command below and then the element macro is built first, then the top level is built:

```
make DESIGN_CONFIG=designs/asap7/mock-array-big/config.mk
```

b224hisl:
if I don't set the `macro-placement.tcl`, Does openroad places the marcos randomly?

maliberty:
> if I don't set the `macro-placement.tcl`, Does openroad places the marcos randomly?

It tries to place them intelligently not randomly.  In this design that macros form a perfect array so it is easier to just specify the placement.

b224hisl:
Thank you very much! I am able to harden my own macro and integrate it into the whole design, but I got an error when doing the floorplanning, do you know what happens?
```
`[WARNING PPL-0015] Macro dmem.dmem2 is not placed.
[WARNING PPL-0015] Macro dmem.dmem3 is not placed.
[WARNING PPL-0015] Macro riscv.dp.rf is not placed.
Found 0 macro blocks.
Using 1u default distance from corners.
Using 2 tracks default min distance between IO pins.
[INFO PPL-0007] Random pin placement.
Elapsed time: 0:02.07[h:]min:sec. CPU time: user 2.01 sys 0.05 (99%). Peak memory: 185160KB.
(/usr/bin/time -f 'Elapsed time: %E[h:]min:sec. CPU time: user %U sys %S (%P). Peak memory: %MKB.' /home/ubuntu/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad -exit -no_init  ./scripts/tdms_place.tcl -metrics ./logs/asap7/riscv32i/b_regfile/2_3_tdms.json) 2>&1 | tee ./logs/asap7/riscv32i/b_regfile/2_3_tdms_place.log
OpenROAD v2.0-6895-g5c85b36f8 
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13178, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13211, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13244, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13277, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13310, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13343, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13376, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 14772, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 14805, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 14838, timing group from output port.
RTLMP flow enabled. Skipping tdms place.
Elapsed time: 0:02.07[h:]min:sec. CPU time: user 1.99 sys 0.08 (99%). Peak memory: 184332KB.
(/usr/bin/time -f 'Elapsed time: %E[h:]min:sec. CPU time: user %U sys %S (%P). Peak memory: %MKB.' /home/ubuntu/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad -exit -no_init  ./scripts/macro_place.tcl -metrics ./logs/asap7/riscv32i/b_regfile/2_4_mplace.json) 2>&1 | tee ./logs/asap7/riscv32i/b_regfile/2_4_mplace.log
OpenROAD v2.0-6895-g5c85b36f8 
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13178, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13211, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13244, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13277, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13310, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13343, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 13376, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 14772, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 14805, timing group from output port.
[WARNING STA-0164] ./platforms/asap7/lib/asap7sc7p5t_SIMPLE_RVT_FF_nldm_211120.lib.gz line 14838, timing group from output port.
HierRTLMP Flow enabled...
Call Macro Placer  -max_num_inst 10000 -min_num_inst 5000 -max_num_macro 4 -min_num_macro 1 -halo_width 10 -report_directory ./objects/asap7/riscv32i/b_regfile/rtlmp
area_weight_ = 0.1
outline_weight_ = 100.0
wirelength_weight_ = 100.0
guidance_weight_ = 10.0
fence_weight_ = 10.0
boundary_weight_ = 20.0
notch_weight_ = 10.0
macro_blockage_weight_ = 10.0
halo_width_ = 10.0
Floorplan Outline: (0.0, 0.0) (120.0, 120.0),  Core Outline: (5, 5) (109, 109)
Traversed logical hierarchy
        Number of std cell instances : 1995
        Area of std cell instances : 210.61
        Number of macros : 5
        Area of macros : 1100.75
        Total area : 1311.36
        Design Utilization : 0.12
        Core Utilization: 0.02

num level: 2, max_macro: 40, min_macro: 10, max_inst:100000, min_inst:50000

Create Data Flow

Perform Clustering..

Print Physical Hierachy Tree

root  (0)  num_macro :  5   num_std_cell :  1995  macro_area :  1100.752  std_cell_area : 210.60783

+---B0  (10)  num_macro :  0   num_std_cell :  0  macro_area :  0.0  std_cell_area : 0.0

+---B1  (11)  num_macro :  0   num_std_cell :  0  macro_area :  0.0  std_cell_area : 0.0

+---B2  (12)  num_macro :  0   num_std_cell :  0  macro_area :  0.0  std_cell_area : 0.0


Break mixed clusters into std cell and macro clusters.

Print Physical Hierachy Tree after splitting std cell and macros in leaf clusters

root  (0)  num_macro :  5   num_std_cell :  1995  macro_area :  1100.752  std_cell_area : 210.60783

+---B0  (10)  num_macro :  0   num_std_cell :  0  macro_area :  0.0  std_cell_area : 0.0

+---B1  (11)  num_macro :  0   num_std_cell :  0  macro_area :  0.0  std_cell_area : 0.0

+---B2  (12)  num_macro :  0   num_std_cell :  0  macro_area :  0.0  std_cell_area : 0.0

Determine shaping function for clusters -- Macro Tilings.

/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = float; _Alloc = std::allocator<float>; std::vector<_Tp, _Alloc>::reference = float&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Signal 6 received
Stack trace:
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = float; _Alloc = std::allocator<float>; std::vector<_Tp, _Alloc>::reference = float&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Signal 6 received
Stack trace:
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = float; _Alloc = std::allocator<float>; std::vector<_Tp, _Alloc>::reference = float&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Signal /usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = float; _Alloc = std::allocator<float>; std::vector<_Tp, _Alloc>::reference = float&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
6 received
Stack trace:
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = float; _Alloc = std::allocator<float>; std::vector<_Tp, _Alloc>::reference = float&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Signal 6 received
Stack trace:
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = float; _Alloc = std::allocator<float>; std::vector<_Tp, _Alloc>::reference = float&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Signal 6 received
Stack trace:
Signal 6 received
Stack trace:
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = float; _Alloc = std::allocator<float>; std::vector<_Tp, _Alloc>::reference = float&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Signal 6 received
Stack trace:
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = float; _Alloc = std::allocator<float>; std::vector<_Tp, _Alloc>::reference = float&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Signal 6 received
Stack trace:
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = float; _Alloc = std::allocator<float>; std::vector<_Tp, _Alloc>::reference = float&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Signal 6 received
Stack trace:
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = float; _Alloc = std::allocator<float>; std::vector<_Tp, _Alloc>::reference = float&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Signal 6 received
Stack trace:
 0# 0x0000564EC6A084EB in /home/ubuntu/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 1# 0x00007F3285674520 in /lib/x86_64-linux-gnu/libc.so.6
 2# pthread_kill in /lib/x86_64-linux-gnu/libc.so.6
 3# raise in /lib/x86_64-linux-gnu/libc.so.6
 4# abort in /lib/x86_64-linux-gnu/libc.so.6
 5# 0x0000564EC6A064F8 in /home/ubuntu/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 6# mpl2::SimulatedAnnealingCore<mpl2::SoftMacro>::packFloorplan() in /home/ubuntu/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 7# mpl2::SimulatedAnnealingCore<mpl2::SoftMacro>::fastSA() in /home/ubuntu/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 8# 0x00007F3285A3F2B3 in /lib/x86_64-linux-gnu/libstdc++.so.6
 9# 0x00007F32856C6B43 in /lib/x86_64-linux-gnu/libc.so.6
10# 0x00007F3285758A00 in /lib/x86_64-linux-gnu/libc.so.6
Command terminated by signal 6
Elapsed time: 0:02.07[h:]min:sec. CPU time: user 1.86 sys 0.10 (95%). Peak memory: 188328KB.
make: *** [Makefile:464: results/asap7/riscv32i/b_regfile/2_4_floorplan_macro.odb] Error 134`
```

maliberty:
@b224hisl can you open an issue with a test case?

maliberty:
The path from io_lsbIn_2 to io_lsbOut_1 though a pure feedthrough in Verilog is buffered in the Element:
image

There are a input & output buffers as well as one created by yosys.  So this is an ordinary combination paths and not a feedthru (ie top level ports directly connected by metal).

I don't see any concern about such a path.   In the generated .lib it is modeled:
```
    pin("io_lsbOut_1") {
      direction : output;
      capacitance : 0.0000;
      timing() {
        related_pin : "io_lsbIn_2";
        timing_sense : positive_unate;
        timing_type : combinational;
	cell_rise(template_3) {
          values("34.28338,36.59597,40.69716,48.45711,63.71187,94.10424,154.80426");
	}
	rise_transition(template_3) {
          values("7.61333,11.68242,19.95987,36.84299,71.06668,139.93100,277.88904");
	}
	cell_fall(template_4) {
          values("35.85234,38.06271,41.87856,48.82059,62.27911,89.06332,142.55353");
	}
	fall_transition(template_4) {
          values("6.99799,10.42904,17.26769,31.14957,59.35966,116.35493,230.77393");
	}
      }
    }
```
Are there any other concerns to be addressed here?  Are you wanting this PR to be merged or was it for discussion?

oharboe:
> Are there any other concerns to be addressed here? Are you wanting this PR to be merged or was it for discussion?

@yupferris Is heading up this work. He may modify or incorporate this PR.  Marking do not merge for now.

yupferris:
Yes, I'm still deciding if I'll use this change or not, but I'm in the middle of cleaning up a few other things first. I'll update and/or close shortly.

yupferris:
Superceded by #955.

yupferris:
@oharboe I don't have rights to close this; can you do it?

PR #829, opened 2023-02-16
Macro placement in tcl

oharboe:


oharboe:
@yupferris OK?

flow/designs/asap7/mock-array-big/macro-placement.tcl
```diff
@@ -0,0 +1,26 @@
...
+set vals [regexp -all -inline {\S+} [exec sh -c {cd designs/asap7/mock-array-big && python3 -c "import config;print(f'{config.rows} {config.cols} {config.x_pitch} {config.offset_x} {config.y_pitch} {config.offset_y}')"}]]
+puts "$vals"
+lassign $vals rows cols x_pitch offset_x y_pitch offset_y
+
+puts "$rows $cols $x_pitch $offset_x $y_pitch $offset_y"
+puts "$y_pitch"
+puts "$offset_y"
+
+puts "asfd2 [expr {$offset_y + 3 * $y_pitch}]"
+puts "asfd [expr {$offset_x + 5 * $x_pitch}] [expr {$offset_y + 3 * $y_pitch}]"
```
yupferris:
```suggestion
lassign $vals rows cols x_pitch offset_x y_pitch offset_y
```

removing unnecessary prints

yupferris (approved):
lgtm, though it'd be nice to see unnecessary prints removed

flow/designs/asap7/mock-array-big/macro-placement.tcl
```diff
@@ -0,0 +1,18 @@
+# fish out values from single source of truth: config.py
+set vals [regexp -all -inline {\S+} [exec sh -c {cd designs/asap7/mock-array-big && python3 -c "import config;print(f'{config.rows} {config.cols} {config.x_pitch} {config.offset_x} {config.y_pitch} {config.offset_y}')"}]]
```
maliberty, line 2: where is "import config" satisfied from?

maliberty: nvm I see it now.  Why not move that to tcl?

oharboe: I would rather program in Python.

I'm now seeing that the https://en.wikipedia.org/wiki/Inversion_of_control used by ORFS is making it hard to do a number of things. 

Examples:

- two macros of different types need to share pin layout, how to do this? First one macro needs to be built, then in the IO pin placement of the second macro, information from the first macro needs to be used.
- it is tricky to communicate the floorplan to small scriptlets that are being invoked here and there(like die area size)


Anyway, solving that in this PR is a bridge to far.

I know there are some efforts to deal with the framework inversion in OpenROAD: openroad is not a module in Python, but is the Python interpreter that invokes Python scripts.

maliberty: You could add a MACRO_PLACEMENT_PY option and do the manual placement from python.  The python API is still under development but for a simple use case like this one it should suffice.

maliberty (approved):
Do you want to merge this or try w/python?

oharboe:
> Do you want to merge this or try w/python?

I think I will add MACRO_PLACEMENT_PY in addition to MACRO_PLACEMENT_TCL

maliberty:
> > Do you want to merge this or try w/python?
> 
> I think I will add MACRO_PLACEMENT_PY in addition to MACRO_PLACEMENT_TCL

Do you want to take that up in another PR then if it is additional?

oharboe:
> > > Do you want to merge this or try w/python?
> > 
> > 
> > I think I will add MACRO_PLACEMENT_PY in addition to MACRO_PLACEMENT_TCL
> 
> Do you want to take that up in another PR then if it is additional?

On second thought, I don't know how to inject a MACRO_PLACEMENT_PY here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/8c78b885f072d7c262d333f06dc260243ef6bc14/flow/scripts/macro_place.tcl#L110

I can't invoke the Python command line, because then I won't have access to the OpenROAD objects.

Doesn't the openroad application pick invoke itself with the python or tcl interpreter? Can it thread the python interpreter into the tcl interpreter?

I think this PR can go in as-is. It is an implementation and simple integration test of MACRO_PLACEMENT_TCL, which I'm sure could be useful to some.

MACRO_PLACEMENT_PY is a separate concern. Perhaps we could support a generic  MACRO_PLACEMENT_SCRIPT that supports either .py or .tcl called from either .py or .tcl flow/scripts in a separate pull request.

maliberty:
The user chooses by adding -python

oharboe:
> The user chooses by adding -python

It is being invoked from the ORFS makefile [macro_place.tcl](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/8c78b885f072d7c262d333f06dc260243ef6bc14/flow/scripts/macro_place.tcl#L110), so can I add -tcl AND -python?

maliberty:
> > The user chooses by adding -python
> 
> It is being invoked from the ORFS makefile [macro_place.tcl](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/8c78b885f072d7c262d333f06dc260243ef6bc14/flow/scripts/macro_place.tcl#L110), so can I add -tcl AND -python?

No you would have to change at the level of the Makefile how OR is invoked and pass it your script instead.

maliberty:
I'll merge this for now and we can take up python in another PR if desired.

PR #841, opened 2023-02-27
swerv_wrapper ASAP7 test case added

vijayank88:
`swerv_wrapper` implemented with ASAP7 platform.


vvbandeira (changes requested):
please add the design to CI

vijayank88:
@vvbandeira 
This design is already available in sky130/nangate45. In ASAP7 platform run time is 2Hr 30Mins. Is that ok to add in CI?
If yes mean need to add both public and nightly jenkins or public CI alone is enough?

vvbandeira:
Currently, the PR pipeline is taking nearly 2h30, so it should be fine to add to both nightly and all tests.

All designs need to be in at least one pipeline; for large designs, we usually add to the nightly since it will not slow down the PR process.

maliberty (approved):
This is fine but you'll need to resolve the merge conflict

vijayank88:
@vvbandeira 
Conflicts got resolved. Can you please check once and merge it?

vvbandeira (changes requested):
> This design is already available in sky130/nangate45. In ASAP7 platform run time is 2Hr 30Mins.

image

This is longer than expected. Please only enable the design in the Nightly build.

vvbandeira (approved):

PR #877, opened 2023-03-18
added git clean command to tutorial

BhardwajBhat:
git repository clean command added to flow tutorial docs


vijayank88:
@BhardwajBhat 
Depending on the OS it will clean results /reports/logs of previous runs.
User must be careful, while selecting the option. So it is not recommended to include here.

PR #882, opened 2023-03-19
Infrastructure suggestions for OpenROAD Flow Scripts for faster user and developer debug experience

12562:
# Infrastructure suggestions for improving debug experience of users of OpenROAD Flow Scripts (ORFS).

## Infrastructure suggestions include 
* splitting the top-level Makefile into individual config.mk files such as synthesis.mk, placement.mk, routing.mk, setup.mk; 
* Creating separate folders for the OpenROAD flow steps containing the related config.mk file as well as related scripts; 
* Add ability to dump custom flow script which contains only the relevant tool calls based on user provided options and default configs;
* Script-based (TCL) run environment where a top-level TCL script can be sourced to run the flow.


rovinski:
I have spent significant time using both flows which consolidate into a single Makefile and flows which disaggregate steps into multiple Makefiles. I strongly prefer the former in terms of readability and usability. This flow is not so insanely complicated that it requires splitting things up. In my opinion, having to navigate across multiple Makefiles when trying to locate targets, variables, etc. is more trouble than it's worth.

Also, please do not add images (.png) of something which could be a text file or .svg diagram.

> * Add ability to dump custom flow script which contains only the relevant tool calls based on user provided options and default configs;
> * Script-based (TCL) run environment where a top-level TCL script can be sourced to run the flow.

These may be of some interest, but I can't see where these changes are because they're buried in the structure change.

vijayank88:
@12562 Please check and review this locally in your machine and complete the contest. Based on selection we will decide PR creation again.

PR #887, opened 2023-03-19
Open road installation par update

risheb:
updated clear git repo on flow tutorial doc

### DIFF
--- a/docs/tutorials/FlowTutorial.md
+++ b/docs/tutorials/FlowTutorial.md
@@ -55,6 +55,7 @@ following commands:
 
 ```
 cd OpenROAD-flow-scripts
+git clean -xdi
 git checkout master
 git pull
 git submodule update


tools/OpenROAD
risheb, line 1: have removed -m64 flag from scr/par/CMakeList.txt to build and install OpenROAD tool in ubuntu 22.04 arm64 version

docs/tutorials/FlowTutorial.md
```diff
@@ -55,6 +55,7 @@ following commands:
 
 ```
 cd OpenROAD-flow-scripts
+git clean -xdi
```
maliberty, line 58: This is a dangerous command (as previously noted).  I do not think we should suggest users use this unless they really need to and understand the implications.

vijayank88:
@risheb 
Depending on the OS it will clean results /reports/logs of previous runs.
User must be careful, while selecting the option. So it is not recommended to include here.

PR #899, opened 2023-03-20
asap7: Fix M8/M8 min area

antonblanchard:
These values came from incorrect values in the upstream ASAP7 which are now fixed.


maliberty:
The failure is related to the lemon repo being offline but all test pass.

maliberty (approved):

PR #902, opened 2023-03-20
Add a fake lvt platform and design using nangate45 as a base. 

openroadie:
Added a LEF file with copies of cells with "LVT" prefix. For example, for AND gates we now have:

[AND2_X1, AND2_X2, AND2_X4, NAND2_X1, NAND2_X2, NAND2_X4]

and we have:
[LVT_AND2_X1, LVT_AND2_X2, LVT_AND2_X4, LVT_NAND2_X1, LVT_NAND2_X2, LVT_NAND2_X4]

Also added a timing library file for the LVT cells where the max rising slew at max load  was reduced for a few cells (buffers/flops for now) so the resizer would pick them over the the normal cells. 

I will also try to setup a more realistic design with asap7 where we don't currently read the LVT libs in any design flow.


openroadie:
I always miss one sign off. Let me redo this.

PR #903, opened 2023-03-20
Add a fake lvt platform and design using nangate45 as a base

openroadie:
Added a LEF file with copies of cells with "LVT" prefix. For example, for AND gates we now have:

[AND2_X1, AND2_X2, AND2_X4, NAND2_X1, NAND2_X2, NAND2_X4]

and we have:
[LVT_AND2_X1, LVT_AND2_X2, LVT_AND2_X4, LVT_NAND2_X1, LVT_NAND2_X2, LVT_NAND2_X4]

Also added a timing library file for the LVT cells where the max rising slew at max load was reduced for a few cells (buffers/flops for now) so the resizer would pick them over the the normal cells.

I will also try to setup a more realistic design with asap7 where we don't currently read the LVT libs in any design flow.

### DIFF
--- a/flow/designs/nangate45_fakelvt/jpeg/metadata-base-ok.json
+++ b/flow/designs/nangate45_fakelvt/jpeg/metadata-base-ok.json
@@ -1,0 +1,1 @@
+../../nangate45/jpeg/metadata-base-ok.json
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ flow/*.opennet
 flow/platforms/*
 !flow/platforms/asap7
 !flow/platforms/nangate45
+!flow/platforms/nangate45_fakelvt
 !flow/platforms/sky130hd
 !flow/platforms/sky130hs
 !flow/platforms/sky130io
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -17,6 +17,8 @@
 # DESIGN_CONFIG=./designs/nangate45/tiny-tests/config.mk
 # DESIGN_CONFIG=./designs/nangate45/tinyRocket/config.mk
 
+# DESIGN_CONFIG=./designs/nangate45_fakelvt/jpeg/config.mk
+
 # DESIGN_CONFIG=./designs/tsmc65lp/aes/config.mk
 # DESIGN_CONFIG=./designs/tsmc65lp/ariane/config.mk
 # DESIGN_CONFIG=./designs/tsmc65lp/black_parrot/config.mk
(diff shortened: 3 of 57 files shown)


flow/designs/nangate45_fakelvt/jpeg/metadata-base-ok.json
```diff
@@ -0,0 +1 @@
+../../nangate45/jpeg/metadata-base-ok.json
```
maliberty, line 1: Wouldn't you expect this to differ?  Same with rules-base.json

maliberty (commented):
Is this just a stop gap until you get asap7 working?  I'm wondering if you view this as permanent or temporary.

flow/designs/nangate45_fakelvt/jpeg/metadata-base-ok.json
openroadie: Since I was only looking at timing, it did not seem necessary. We can always change it later if needed for a more robust test.

openroadie:
@maliberty I look at it as a permanent addition (if it adds value for others). If not then I am fine to just keep it as a branch in my fork (because it is a lot simpler/faster/easier to use). asap7 is also synthetic anyways :).

flow/designs/nangate45_fakelvt/jpeg/metadata-base-ok.json
maliberty: Fwiw the metrics contain slacks

openroadie: Learned something new. I will update this file after running the baseline flow.

maliberty:
> @maliberty I look at it as a permanent addition (if it adds value for others). If not then I am fine to just keep it as a branch in my fork (because it is a lot simpler/faster/easier to use). asap7 is also synthetic anyways :).

"simpler/faster/easier" than asap7 in what way?  

You need to add it to the jenkins/... nightly & all in order for it to run.

openroadie:
Smaller number of files for me to deal with

rovinski:
I would be careful with this. In the past, people have added hacked or synthetic design libraries to aid in testing / development, but it always turned out that a fully accurate model was needed and many troublesome problems were caused by not having an accurate model.

One example is the FakeRAM memory generator, which gave a good first-order timing model of an SRAM. But the shortcomings of the model (no default transition, unreasonable hold time, unreasonable drive strength, unrealistic liberty tables, etc.) lead to many unconventional errors and problems with the tools.

I can easily see an issue of overfitting without a complete library here. If the leakage and power consumption are not accurately modeled, then it makes it very difficult to properly balance power vs. slack.

I would recommend including a README about the limitations and shortcomings of this synthetic PDK vs. a real PDK, and the limited ways it can be used effectively. Hopefully that will help people realize the limitations if this gets used in the future.

Also, you will need to create a GDS copy and update the cell names if you plan on exporting to GDS. Otherwise there will be a cell name mismatch.

maliberty:
I'm not sure of the general utility of this.  It seems particular to rsz development.  I think it might be better in a personal branch.

openroadie:
Sounds good. Personal branch it is.

PR #906, opened 2023-03-21
modified the load script

jaswanthpullepu:


vijayank88:
@jaswanthpullepu 
Test this code locally on your machine.
I think still you're not understood flow requirement.
Concentrate on contest and check the pinned message in the channel.
We will let you know in that channel to make PR or not. Thanks

PR #910, opened 2023-03-22
Makefile: add generate_mock_abstract target

oharboe:
@maliberty Any tips?

Perhaps it is OK to have mock abstracts work only to and including the "make route"(detailed route) stage and not all the way to "make finish"?

When I try:

```
make DESIGN_CONFIG=designs/asap7/mock-array-big/Element/config.mk generate_mock_abstract
make DESIGN_CONFIG=designs/asap7/mock-array-big/config.mk
```

I get:

```
[deleted]
[INFO] Merging GDS/OAS files...
	./platforms/asap7/gds/asap7sc7p5t_28_R_220121a.gds
	./results/asap7/mock-array-big_Element/base/6_final.gds
ERROR: Unable to open file: ./results/asap7/mock-array-big_Element/base/6_final.gds (errno=2) in Layout.read
  ./util/def2stream.py:134
Elapsed time: 0:01.99[h:]min:sec. CPU time: user 1.60 sys 0.16 (89%). Peak memory: 330036KB.
make: *** [results/asap7/mock-array-big/base/6_1_merged.gds] Error 1
```

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -664,8 +664,23 @@ $(LOG_DIR)/6_report.log: $(RESULTS_DIR)/6_1_fill.odb $(RESULTS_DIR)/6_1_fill.sdc
 
 $(RESULTS_DIR)/6_final.def: $(LOG_DIR)/6_report.log
 
+# To speed up turnaround times when composing
+# macros to a higher level design, such as doing floorplanning,
+# or tweaking options for detailed routing, it can be useful to quickly
+# generate a mock abstract.
+#
+# This mock abstract has the same
+# interface(pins in the same places, blockages, etc.) as the final abstract,
+# but it is essentially an eviscerated macro.
+#
+# At a higher level, it is then possible to take the designs all
+# the way through "make route" (detailed route), but beyond that a .gds file is
+# needed, which the mock abstract does not contain.
+generate_mock_abstract: $(RESULTS_DIR)/3_place.odb $(RESULTS_DIR)/3_place.sdc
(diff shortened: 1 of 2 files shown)


oharboe:
@maliberty  I think this PR is good to go. There will be some followup issues, but those are not related to this change as such. 

For instance, I think there is a performance issue: https://github.com/The-OpenROAD-Project/OpenROAD/issues/3069

oharboe:
@maliberty The failure in in the build of this PR is not related to the change.

maliberty:
It depends what the goal is for this mock abstract.  The .lib timing values will not be the equivalent after gpl vs drt.  

Now that we use -bloat_occupied_layers the lef abstracts will be more similar.  Note that we only block layers that have shapes on them.  After gpl the only shapes will be from the pdn network.  If your pdn network covers all your signal routing layers then you should be ok physically.  If not you will have unblocked layers that could be used later.

Does this match your intended use model?

oharboe:
> Does this match your intended use model?

What I want to do is to quickly iterate on combining macros: floorplanning as well as all the way to detailed route. 

It doesnt have to be perfect, but I would like to deal with most issues before I start full flows that will take a long time to iterate on.

Of course I expect details on timing and power to be at best a poor approximation.

maliberty:
From earlier 
 If your pdn network covers all your signal routing layers then you should be ok physically. If not you will have unblocked layers that could be used later.

That would be my only concern for physical planning

maliberty:
It might be better to not generate a .lib as it will be misleading at best.

oharboe:
> It might be better to not generate a .lib as it will be misleading at best.

There are two problems here:

1. generating the .lib file takes a long time. Even for generate_abstract, I think it can be in the neighbourhood of 50%
2. Unless you are using `BLOCKS` in the makefile, which doesn't work for multi-level designs there will be an ```ADDITIONAL_LIBS``` variable referring to a .lib file. Without a .lib file one has to tweak the higher level makefiles, which worsens the workflow.

maliberty (approved):

PR #915, opened 2023-03-23
DO NOT MERGE - Example slow generate mock abstract

oharboe:
Attempt at reproducing slow generate_mock_abstract

Rebase this PR on master.

Next, generate the Element macro:

```
MOCK_ARRAY_WIDTH=16 MOCK_ARRAY_HEIGHT=8 MOCK_ARRAY_DATAWIDTH=64 MOCK_ARRAY_PITCH_SCALE=20 make DESIGN_CONFIG=designs/asap7/mock-array-big/Element/config.mk generate_abstract
```

Next, create a mock abstract:

```
MOCK_ARRAY_WIDTH=16 MOCK_ARRAY_HEIGHT=8 MOCK_ARRAY_DATAWIDTH=64 MOCK_ARRAY_PITCH_SCALE=40 make DESIGN_CONFIG=designs/asap7/mock-array-big/config.mk generate_mock_abstract
[deleted]
Elapsed time: 4:29.68[h:]min:sec. 

```

Hmm... so that took only 4 minutes and 30 seconds...

The example that took 3 hours generated ca. 5x larger files: https://github.com/The-OpenROAD-Project/OpenROAD/issues/3069

```
-rw-r--r-- 1 1000 1000 4.7M Mar 23 06:12 MockArray.lib
-rw-r--r-- 1 1000 1000  16M Mar 23 06:12 MockArray.lef
```

I think it can be useful to profile this test-case as it has faster turnaround times and it might provide some insight into why geneate_mock_abstract is having trouble scaling.

This PR is here to avoid having to run this step, which generates a 16x8 64 bit mock-array-big Verilog from Chisel:

```
MOCK_ARRAY_WIDTH=16 MOCK_ARRAY_HEIGHT=8 MOCK_ARRAY_DATAWIDTH=64 ./configure.sh
```

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -664,8 +664,23 @@ $(LOG_DIR)/6_report.log: $(RESULTS_DIR)/6_1_fill.odb $(RESULTS_DIR)/6_1_fill.sdc
 
 $(RESULTS_DIR)/6_final.def: $(LOG_DIR)/6_report.log
 
+# To speed up turnaround times when composing
+# macros to a higher level design, such as doing floorplanning,
+# or tweaking options for detailed routing, it can be useful to quickly
+# generate a mock abstract.
+#
+# This mock abstract has the same
+# interface(pins in the same places, blockages, etc.) as the final abstract,
+# but it is essentially an eviscerated macro.
+#
+# At a higher level, it is then possible to take the designs all
+# the way through "make route" (detailed route), but beyond that a .gds file is
+# needed, which the mock abstract does not contain.
+generate_mock_abstract: $(RESULTS_DIR)/3_place.odb $(RESULTS_DIR)/3_place.sdc
+	(ABSTRACT_FROM=3_place $(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/generate_abstract.tcl -metrics $(LOG_DIR)/generate_abstract.json ) 2>&1 | tee $(LOG_DIR)/generate_abstract.log
+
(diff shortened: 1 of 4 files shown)


maliberty:
Is this a PR to demonstrate a problem or one that you want merged?

oharboe:
> Is this a PR to demonstrate a problem or one that you want merged?

It is a draft PR intended only as an example. Added DO NOT MERGE to synopsis

rovinski:
@oharboe I added a "do not merge" label for you.

maliberty:
I don't have chisel installed so this isn't useful in this form.  Would you generate the relevant files and add them to this PR?

oharboe:
> I don't have chisel installed so this isn't useful in this form. Would you generate the relevant files and add them to this PR?

This is indeed the raison d'etre for this PR: I ran Chisel to generate the Verilog foles.

QuantamHD:
@oharboe Is there any chance you could run linux perf on your openroad binary for about a minute?

As a side note I noticed that you ran you abstract lef generator after fill. The fill shapes are very numerous, and I suspect could be causing some of the slow down. You could test this by running generate_abstract after detailed routing, but before fill.

maliberty:
I was confused by your description which seemed to imply a set of steps to reproduce the problem.  I guess that was the set of steps used to generate the branch.

oharboe:
@eder-matheus More pins in mock-array-big.

To test, run:

```
MOCK_ARRAY_WIDTH=16 MOCK_ARRAY_HEIGHT=8 MOCK_ARRAY_DATAWIDTH=64 MOCK_ARRAY_PITCH_SCALE=20  make DESIGN_CONFIG=designs/asap7/mock-array-big/Element/config.mk place
```

PR #917, opened 2023-03-23
flow: remove report power

vvbandeira:


maliberty (approved):
The removal of report_power is temporary until an sta bug can be fixed.

PR #928, opened 2023-03-29
setup: safe, no choices to be made, setup & build locally

oharboe:
Safer and easier to build OpenROAD locally and no choices to be made:

```
sudo ./setup.sh
./build_openroad --local
. setup_env.sh
cd flow
make DESIGN_CONFIG=designs/asap7/gcd/config.mk
```

Voila!

### DIFF
--- a/dev_env.sh
+++ b/dev_env.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+function setpaths() {
+    local DIR=$(readlink -f "$(dirname "${BASH_SOURCE[0]}")")
+    export PATH="$DIR/dependencies/bin:$PATH"
+}
+
+setpaths
--- a/setup.sh
+++ b/setup.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# allow this script to be invoked from any folder
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+if [ $EUID -ne 0 ]; then
+  echo "This script must be run with sudo"
+  exit 1
+fi
+
+"$DIR/tools/OpenROAD/etc/DependencyInstaller.sh" -base
+
+sudo -u $SUDO_USER "$DIR/tools/OpenROAD/etc/DependencyInstaller.sh" -common -prefix="$DIR/dependencies"
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,5 @@ core.*
 
 # Python bytecode
 __pycache__/
+
+dependencies/
--- a/build_openroad.sh
+++ b/build_openroad.sh
@@ -7,6 +7,10 @@ set -eu
 # Make sure we are on the correct folder before beginning
 cd "$(dirname $(readlink -f $0))"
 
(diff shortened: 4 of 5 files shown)


setup.sh
```diff
@@ -0,0 +1,15 @@
...
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+if [ $EUID -ne 0 ]; then
+  echo "This script must be run with sudo"
+  exit 1
+fi
+
+$DIR/tools/OpenROAD/etc/DependencyInstaller.sh -base
+
+sudo -u $SUDO_USER $DIR/tools/OpenROAD/etc/DependencyInstaller.sh -common -prefix=$DIR/dependencies
```
vvbandeira:
```suggestion
"$DIR/tools/OpenROAD/etc/DependencyInstaller.sh" -base

sudo -u $SUDO_USER "$DIR/tools/OpenROAD/etc/DependencyInstaller.sh" -common -prefix="$DIR/dependencies"
```

dev_env.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+function setpaths() {
+    local DIR=$(readlink -f "$(dirname "${BASH_SOURCE[0]}")")
+    export PATH=$DIR/dependencies/bin:$PATH
```
vvbandeira:
```suggestion
    export PATH="$DIR/dependencies/bin:$PATH"
```

vvbandeira (changes requested):
Thanks for the contribution. Just a suggestion to wrap variables in quotes to consider if any folder in the path has a space.

setup.sh
oharboe: Makes sense, but did you try it? I worry about the -prefix bit, shouldnt that be quotes around the entire argument?

oharboe:
@vvbandeira Tested your changes. Worked fine!

oharboe:
@maliberty Build server problems, false positive.

maliberty:
@vvbandeira is looking at it. Seems like out of disk space

oharboe:
> @vvbandeira is looking at it. Seems like out of disk space

👍 Classic.

vvbandeira (approved):

oharboe:
@vvbandeira Thanks!

vijayank88:
@vvbandeira 
Due to confusion raising this question.
Now with `setup.sh` required packages installed from `tools/OpenROAD/etc/DependecyInstaller.sh`.
But it is different from `OpenROAD-flow-scripts/etc/DependencyInstaller.sh` right?
Because ORFS flow installer script will do KLayout installation also part of the script. But not `./setup.sh` right?

oharboe:
> @vvbandeira Due to confusion raising this question. Now with `setup.sh` required packages installed from `tools/OpenROAD/etc/DependecyInstaller.sh`. But it is different from `OpenROAD-flow-scripts/etc/DependencyInstaller.sh` right? Because ORFS flow installer script will do KLayout installation also part of the script. But not `./setup.sh` right?

Like so?

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/949

PR #929, opened 2023-03-29
Enable OpenROAD GUI access in WSL Ubuntu22.04

vijayank88:
For WSL based Ubuntu22.04 by installing `xterm` OpenROAD gui will be opened. 
Otherwise you may face following error and get stuck for long.
```
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-'
This plugin does not support propagateSizeHints()
```

### DIFF
--- a/docs/user/BuildLocally.md
+++ b/docs/user/BuildLocally.md
@@ -28,3 +28,8 @@ yosys -help
 openroad -help
 exit
 ```
+
+> **Note:** For WSL based Ubuntu22.04 install `xterm` to access OpenROAD GUI.
+ ```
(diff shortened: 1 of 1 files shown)


maliberty:
I use wsl ubuntu 22.04 from Windows Terminal and can start the GUI without error.   What do you do that produces that error?

vijayank88:
I got this from slack channel discussion, who installed ORFS 1st time and unable to view gui with `make gui_final`.
After installing `xterm` he is informed it was working fine. So for document purpose, updated this.

vijayank88:
Can we close this PR? @maliberty

Issue #919, opened 2023-03-23
Unconnected PDN node on net VDD


yupferris:
### Subject

[Stage]: Other. Please describe below.

### Describe the bug

In https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/918, I'm attempting to clean up floorplanning setup for the `mock-array-big` example. In doing so, I either get DRC errors as power straps end up too close to macros, or, if I adjust the pdn config [as such](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/918/files#diff-27ac0f3dd04316cbe20c08153c07233c020eae57be8e3711d0f220e13af23463L16-R17) to align better with the macros, I get a bunch of `Unconnected PDN node on net VDD` errors, which I don't know how to fix.

### Expected Behavior

The flow completes successfully and generates GDSII.

### Environment

```shell
[WARNING] Your current OpenROAD version is outdated.
It is recommened to pull the latest changes.
If problem persists, file a github issue with the re-producible test case.
kernel: Linux 5.15.0-60-generic
os: Ubuntu 22.04.1 LTS (Jammy Jellyfish)
cmake version 3.22.1
-- The CXX compiler identification is GNU 11.3.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-7229-gd76c80d8e
-- System name: Linux
-- Compiler: GNU 11.3.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- Could NOT find GTest (missing: GTEST_LIBRARY GTEST_INCLUDE_DIR GTEST_MAIN_LIBRARY) 
-- The C compiler identification is GNU 11.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python: /usr/bin/python3.10 (found version "3.10.6") found components: Interpreter 
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
[...]

PR #918, opened 2023-03-23
WIP: Fix io alignment for mock-array-big

yupferris:
Also clean up some of the config/setup stuff to be a bit more straightforward.

Currently breaks IR drop setup (#919).

PR #934, opened 2023-03-30
Fix io alignment for mock-array-big

yupferris:
Also clean up some of the config/setup stuff to be a bit more straightforward.

This is basically #918 rebased onto latest master. #918 is being left as-is for debugging #919.


yupferris:
This crashes currently, see #935.

eder-matheus:
> This crashes currently, see #935.

I have this branch that should fix the crash: https://github.com/eder-matheus/OpenROAD/tree/ppl_groups. Could you try it?

yupferris:
I've upgraded OpenROAD to a version which contains these fixes and verified that this PR gets past the previous ppl errors, however now I'm getting a divergence in global placement, which I'll investigate a bit.

yupferris:
Tweaking some options works around this and makes the design a lot tighter with negligible additional runtime, so this should be good to go.

yupferris:
@maliberty can you review when you get the chance? (I don't have rights to request reviews, hence asking this way)

yupferris:
I don't understand the failures - it looks like there's some kind of results check that happens after building `mock-array-big` which now fails as the results have changed. Is this correct? Are there docs on how should I reproduce this locally and update?

vijayank88:
@yupferris 
Following metrics failing.
```
[ERROR] detailedroute__route__wirelength fail test: 39452.0 <= 31191.0
```
If it is acceptable, do `make update_ok` in your local machine and push the changes to pass the CI.

yupferris:
Done, thanks for the clarification.

tools/OpenROAD
maliberty, line 1: Is this necessary for this PR?

yupferris: Yes, otherwise we hit #935.

maliberty (approved):

PR #936, opened 2023-03-30
designs/asap7/mock-array-big: Skip analyze_power_grid

oharboe:
@yupferris I think we talked about making this change yesterday?


yupferris:
We did, but I don't think we should skip the step if we can afford to do it (runtime-wise) and it's not blocking other work (which it doesn't after #933), if nothing else to get some additional sanity checks. We'll want to re-enable it later anyways if/when we articulate a proper pdn.

oharboe:
@maliberty @yupferris We'll restore this PR if we decide we want to disable this step.

PR #939, opened 2023-03-30
Fix all violation

openroadie:
Enable TNS_END_PERCENT in the cts.tcl script and also use it for two designs where it help. I am going to look at the rest of the designs because in some cases it makes the final timing worse. @maliberty thinks is may be due to the cells getting moved too much during detailed placement/routing ... or it could just be a bug in buffering/upsizing.


maliberty (approved):
We should add a tns metric now that we are optimizing it.

PR #960, opened 2023-04-06
macro grid for asap7

vijayank88:
Added macro grid for ASAP7 platform.

### DIFF
--- a/flow/designs/asap7/uart-blocks/BLOCKS_grid_strategy.tcl
+++ b/flow/designs/asap7/uart-blocks/BLOCKS_grid_strategy.tcl
@@ -1,0 +1,29 @@
+####################################
+# global connections
+####################################
+add_global_connection -net {VDD} -inst_pattern {.*} -pin_pattern {^VDD$} -power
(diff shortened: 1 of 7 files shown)


vijayank88:
@maliberty 
Any suggestion why uart_blocks failing during IR drop?
IR Drop is disable right?

vijayank88:
#965 causing PSM IR Drop report error.

maliberty:
> @maliberty Any suggestion why uart_blocks failing during IR drop? IR Drop is disable right?

ir drop is enabled except for asap7/swerv_wrapper (tbd)

maliberty:
In your pdn changes you have:
add_pdn_connect -grid {CORE_macro_grid_1} -layers {M4 M5}

in the uart design the power pins are on M5 and need to be connected to M6.

vijayank88:
@maliberty 
`uart-blocks` is based on `BLOCKS` implementation. Resolved by adding separate pdn grid for the same.

maliberty (approved):

PR #967, opened 2023-04-08
make: fix elapsed time regression and add elapsed target

oharboe:
the elapsed target is useful when wanting to know where e.g. a failed spent it's time or invoked from generate_abstract

@habibayassin @vvbandeira   genElapsedTime.py doesn't work, it doesn't seem to distinguish between hours and minutes

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -640,8 +640,11 @@ finish: $(LOG_DIR)/6_report.log \
         $(RESULTS_DIR)/6_final.v \
         $(RESULTS_DIR)/6_final.sdc \
         $(GDS_FINAL_FILE)
+	$(MAKE) elapsed
+
+elapsed:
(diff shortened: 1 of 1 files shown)


maliberty:
We should fix genElapsedTime.py as it is used elsewhere rather than spot fix one rule

oharboe:
> We should fix genElapsedTime.py as it is used elsewhere rather than spot fix one rule

@maliberty Makes sense to me. Does the .py scripts in that folder have unit-tests set up?

oharboe:
@maliberty waiting for fix...

habibayassin:
> @maliberty waiting for fix...

@oharboe I added [here](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/979) the fix and unit tests for it. The reason behind using python script instead of awk is that awk wasn’t compatible with all operating systems and required adding an extra dependency in the installer.

PR #969, opened 2023-04-09
DO NOT MERGE - macros on top of eachother

oharboe:
The macros in this modified mock-array-big are on top of eachother, yet `make floorplan` does not produce an error.

image


maliberty:
Manually placed macros are not checked for overlap.  It could be intentional.

oharboe:
> Manually placed macros are not checked for overlap. It could be intentional.

The same problem exists for automatically placed macros.

maliberty:
Can you modify this to show that?

oharboe:
> Can you modify this to show that?

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/972

image

PR #977, opened 2023-04-11
make: add missing .PHONY statements

oharboe:
without this, you can accidentally trigger a default make rule.

E.g.:

touch floorplan.sh
make floorplan

This results in make running a default rule, unless .PHONY statements are added:

cat floorplan.sh > floorplan
chmod a+x floorplan


yupferris:
nice catch!

maliberty (approved):

PR #982, opened 2023-04-12
fixed timing violation for ibex/asap7

naveen17440:
Based on 7nm OpenROAD contest, I have made cts buffer sizing and updated sdc constraints with false path for reset and achieved TNS/WNS 0.


vijayank88:
@dralabeing fyi

dralabeing:
The false path sdc constraint only applies to the reset pin not the tie cells

maliberty:
There are too many different things mixed together to tell if they are all necessary and beneficial.  Please break out the benefit of the CTS buffer change from the false path change.

Why are the IO constraints changed?  That isn't an optimization but just a more optimistic assumption.  I don't see the point of this change.

vijayank88:
@maliberty 
I have ran the design with the update of CTS buffer change, final timing report had wns as `-33.13ns`.
The violated endpoint path is `instr_addr_o[25]`
I think user updated `instr_addr_o` output delay to achieve `WNS  0`.
Adding output delay to specific net is sufficient right?

naveen17440:
I am new to the automated flow. I am just experimented with default sdc template. But unable to over-write `set_output_delay`.

So I've changed `IO_PCT` to reduce the WNS further as I'm not able to over-ride default output delay for particular endpoint. 

I have restored sdc file into default format with false path for `rst_ni` port.

Please let me know is anything else to update on this. Thanks for the guidance.

maliberty:
The output delay represents the requirements of registers outside this block.  We make some reasonable assumption for the blocks for which we have no actual context.  Making that assumption more optimistic isn't really an optimization.

maliberty:
stale

PR #955, opened 2023-04-04
Clean up/extend mock-array-big

yupferris:

PR #986, opened 2023-04-14
mock-array-big/Element: Proper timing constraints

yupferris:
Currently on top of https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/955, but 725b5f61c690c71a7e577152c3ac1e873a16f68d is the important thing here.


flow/designs/asap7/mock-array-big/Element/constraints.sdc
```diff
@@ -0,0 +1,39 @@
...
+set max_delay [expr $budget_per_element * ($cols - 1) + $headroom]
+
+# REG[0] (io_outs_left[0] in the source) -> io_lsbOuts_7
+set_output_delay -clock $clk_name -min $min_delay [get_ports {io_lsbOuts_7}]
+set_output_delay -clock $clk_name -max $max_delay [get_ports {io_lsbOuts_7}]
+
+# All remaining non-clock IOs are only connected to one another without going through any
+#  registers (in -> out paths). Such paths should not be checked for setup/hold violations
+#  and do not need to be constrained.
+set_false_path -from [get_ports {io_lsbIns_*}] -to [get_ports {io_lsbOuts_*}]
```
yupferris: I'm wondering if the following is actually equivalent to this last line in this case:

```
set non_clk_inputs [lsearch -inline -all -not -exact [all_inputs] $clk_port]
set_false_path -from $non_clk_inputs -to [all_outputs]
```

This would be a bit simpler, especially when applying this approach to our internal design.

Related: Is there a good way to list the paths that are qualified by a given set of patterns? Eg. can I see a full list of paths that are covered by a given constraint?

maliberty: You could use 
```report_checks -through [get_ports {io_lsbIns_*}] -through [get_ports {io_lsbOuts_*}]```

to get paths through those points.  The number of paths is potentially exponential so I wouldn't recommend getting all of them.  You can choose how many are reported.

yupferris: Thanks for the tip. I've looked at all of the relevant paths and verified that all relevant paths are covered as intended, and for handling in -> out paths, indeed it is equivalent (and imo, simpler) to mark all paths starting with (non-clock) inputs to outputs as false in this case.

yupferris:
@maliberty can we re-run jenkins here, or something? I think this should be good now that #955 is passing, which this is based on.

vijayank88:
@yupferris 
Problem with `docker-build`. Re-running again.

maliberty (approved):

PR #991, opened 2023-04-18
Do not merge test wide design

oharboe:
Works:

```
MOCK_ARRAY_WIDTH=32 MOCK_ARRAY_HEIGHT=1 make DESIGN_CONFIG=designs/asap7/mock-array-big/config.mk
```


oharboe:
Not the specific problem I was looking for. Easily restored if interesting.

PR #993, opened 2023-04-18
tools/yosys: update to 0.28

oharboe:
Allay concerns that ORFS is far behind yosys and hard to upgrade.


oharboe:
Requires some updates to install scripts. This PR is easily recreated when the time comes to upgrade to >=0.28.

PR #994, opened 2023-04-18
designs/sky130hs/mock-array-big: add sky130hs configuration to be able to compare

oharboe:
Not quite straight wires...

image


oharboe:
Easily resurrected if we decide to have a sky130hs version of this design.

PR #997, opened 2023-04-19
metrics: do not round synth area

vvbandeira:
It can cause issues with small designs such as asap7/mock-array-big which has area less than 1.0


maliberty (approved):

vvbandeira:
@vijayank88 fix for https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/955#issuecomment-1511003051

PR #1000, opened 2023-04-20
Makefile: use ln instead of cp to save space

oharboe:
this starts to matter for .odb files in the gigabytes size, which will happen e.g. in exploration of mm^2 sized ASAP7 designs with lots of empty space.

Empty space on a die can be common when exploring floorplans and starting with an overgenerous placement of macros.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -743,7 +743,7 @@ $(GDS_MERGED_FILE): $(RESULTS_DIR)/6_final.def $(OBJECTS_DIR)/klayout.lyt $(GDSO
 $(RESULTS_DIR)/6_final.v: $(LOG_DIR)/6_report.log
 
 $(GDS_FINAL_FILE): $(GDS_MERGED_FILE)
-	cp $^ $@
+	cp $< $@
 
 .PHONY: drc
 drc: $(REPORTS_DIR)/6_drc.lyrdb
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -399,11 +399,11 @@ $(RESULTS_DIR)/1_1_yosys.v: $(DONT_USE_LIBS) $(WRAPPED_LIBS) $(DONT_USE_SC_LIB) 
 
 $(RESULTS_DIR)/1_synth.v: $(RESULTS_DIR)/1_1_yosys.v
 	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)
-	cp $< $@
+	ln -sf $(shell realpath --relative-to=$(dir $@) $<) $@
 
 $(RESULTS_DIR)/1_synth.sdc: $(SDC_FILE)
 	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)
-	cp $< $@
+	ln -sf $(shell realpath --relative-to=$(dir $@) $<) $@
 
 clean_synth:
 	rm -f  $(RESULTS_DIR)/1_*.v $(RESULTS_DIR)/1_synth.sdc
@@ -443,7 +443,7 @@ $(RESULTS_DIR)/2_2_floorplan_io.odb: $(RESULTS_DIR)/2_1_floorplan.odb $(IO_CONST
 ifndef IS_CHIP
 	($(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/io_placement_random.tcl -metrics $(LOG_DIR)/2_2_floorplan_io.json) 2>&1 | tee $(LOG_DIR)/2_2_floorplan_io.log
 else
-	cp $< $@
+	ln -sf $(shell realpath --relative-to=$(dir $@) $<) $@
 endif
 
 # STEP 3: Timing Driven Mixed Sized Placement
@@ -453,7 +453,7 @@ ifeq ($(MACRO_PLACEMENT),)
 	($(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/tdms_place.tcl -metrics $(LOG_DIR)/2_3_tdms.json) 2>&1 | tee $(LOG_DIR)/2_3_tdms_place.log
 else
 	$(info [INFO][FLOW] Using manual macro placement file $(MACRO_PLACEMENT))
-	cp $< $@
+	ln -sf $(shell realpath --relative-to=$(dir $@) $<) $@
 endif
 
 # STEP 4: Macro Placement
@@ -473,7 +473,7 @@ $(RESULTS_DIR)/2_6_floorplan_pdn.odb: $(RESULTS_DIR)/2_5_floorplan_tapcell.odb $
 	($(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/pdn.tcl -metrics $(LOG_DIR)/2_6_pdn.json) 2>&1 | tee $(LOG_DIR)/2_6_pdn.log
 
 $(RESULTS_DIR)/2_floorplan.odb: $(RESULTS_DIR)/2_6_floorplan_pdn.odb
-	cp $< $@
+	ln -sf $(shell realpath --relative-to=$(dir $@) $<) $@
 
 $(RESULTS_DIR)/2_floorplan.sdc: $(RESULTS_DIR)/2_1_floorplan.odb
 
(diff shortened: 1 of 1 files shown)


maliberty:
Have you verified that make will get the right time stamp when using links (the time of the linked file not the symlink itself)?

oharboe:
> Have you verified that make will get the right time stamp when using links (the time of the linked file not the symlink itself)?

Note that cp is run on small .odb files that can get the sub-seconds range for the timestamp difference between the source and destination. Also, make, for compatibility reasons and also depending on the underlying file system, can't be trusted for sub second precision and accuracy, for that something like ninja is required.

When you run the command ln -s a b, you create a symbolic link (symlink) named 'b' that points to the file or directory 'a'. The command does not modify the timestamps of either the source file (a) or the symlink (b).

However, the symlink 'b' will have its own inode, and the timestamps associated with it (such as creation and modification time) will be set to the time when the symlink is created. These timestamps are independent of the source file 'a'.

So... we should be good...

vijayank88:
Seems all CI fails.

vijayank88:
```
+ make DESIGN_CONFIG=./designs/asap7/aes/config.mk final_report_issue

+ tee -a ./logs/asap7/aes.log

[INFO][FLOW] Using platform directory ./platforms/asap7

[INFO-FLOW] ASU ASAP7 - version 2

Default PVT selection: BC

./util/makeIssue.sh final_report

Creating run-me-aes-asap7-base.sh script

Creating vars-aes-asap7-base.sh/tcl script

Archiving issue to final_report_aes_asap7_base_2023-04-20_06-53.tar.gz

tar: ./results/asap7/aes/base/6_final.gds: File removed before we read it

make: *** [util/utils.mk:91: final_report_issue] Error 1

script returned exit code 2
```

oharboe:
> Seems all CI fails.

Thanks! I'll look at it.

oharboe:
@maliberty Some problems with docker builds, looks like network outage during installation.

vijayank88:
Re-started the CI regression. @oharboe

oharboe:
@maliberty Should be good to go now.

maliberty (dismissed):

flow/Makefile
```diff
@@ -396,11 +396,11 @@ $(RESULTS_DIR)/1_1_yosys.v: $(DONT_USE_LIBS) $(WRAPPED_LIBS) $(DONT_USE_SC_LIB)
 
 $(RESULTS_DIR)/1_synth.v: $(RESULTS_DIR)/1_1_yosys.v
 	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)
-	cp $< $@
+	ln $< $@
```
maliberty: What happened to the -s?

oharboe: Full explanation here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1000#issuecomment-1515814679

maliberty:
While I see the appeal I'm not a fan of using hard links.  They are very hard to distinguish from a regular file and can have surprising semantics (hey I edited foo and bar changed).   Were you unable to get symlinks to work?

oharboe:
> ks. They are very hard to distinguish from a regular file and can have surprising semantics (hey I edited foo and bar changed). Were you unable to get symlinks to work?

No. 

Also, the semantics you describe above apply to symlinks too.

oharboe:
@maliberty This is ready to go in as far as I am concerned.

maliberty:
I'm not in favor of using hardlinks.  I find a rule like this works fine:
```
$(RESULTS_DIR)/2_floorplan.odb: $(RESULTS_DIR)/2_6_floorplan_pdn.odb
	[ -L $@ ] || ln -s $(notdir $<) $@
```

Make looks at the linked files mtime not the link itself.  The -L is needed to avoid problems when the link already exists.

oharboe:
@maliberty Changed to softlinks as requested.

flow/Makefile
```diff
@@ -736,7 +736,7 @@ $(GDS_MERGED_FILE): $(RESULTS_DIR)/6_final.def $(OBJECTS_DIR)/klayout.lyt $(GDSO
 $(RESULTS_DIR)/6_final.v: $(LOG_DIR)/6_report.log
 
 $(GDS_FINAL_FILE): $(GDS_MERGED_FILE)
-	cp $^ $@
+	ln $^ $@
```
maliberty: Still hard linking here.

maliberty (changes requested):
Please look at the test failures

flow/Makefile
oharboe: Fixed

maliberty:
Why did you change -L to -e ?  If a file exists but is not a link we have a problem

oharboe:
> Why did you change -L to -e ? If a file exists but is not a link we have a problem

Why? Someone may manually copy a file,  e.g. when using deltaDebug.py.

maliberty:
> > Why did you change -L to -e ? If a file exists but is not a link we have a problem
> 
> Why? Someone may manually copy a file, e.g. when using deltaDebug.py.

In which case the code as written will not work.  The file exists so nothing happens when in fact it should be either copied or linked.

oharboe:
> > > Why did you change -L to -e ? If a file exists but is not a link we have a problem
> > 
> > 
> > Why? Someone may manually copy a file, e.g. when using deltaDebug.py.
> 
> In which case the code as written will not work. The file exists so nothing happens when in fact it should be either copied or linked.

deltaDebug.py is a usecase that has to be supported.

Which brings me back to hard links. How exactly are soft links preferable?

The objections you have to soft links, I seem to recall apply to hard links as well.

maliberty:
How does this relate to delta debug?  It operates on a single odb.

oharboe:
> How does this relate to delta debug? It operates on a single odb.

It copies the modified .odb file into place and then the step to repeat the bug may be invocation of the makefile. Hence, copying a .odb file as an input to a build stage needs to be supported.

maliberty:
> > How does this relate to delta debug? It operates on a single odb.
> 
> It copies the modified .odb file into place and then the step to repeat the bug may be invocation of the makefile. Hence, copying a .odb file as an input to a build stage needs to be supported.

If it is in place then why is there any copying involved.  I still don't get your point.

oharboe:
> > > How does this relate to delta debug? It operates on a single odb.
> > 
> > 
> > It copies the modified .odb file into place and then the step to repeat the bug may be invocation of the makefile. Hence, copying a .odb file as an input to a build stage needs to be supported.
> 
> If it is in place then why is there any copying involved. I still don't get your point.

A modified, bisected .odb file is written to the results folder. So if I later clean an earlier stage, I can find a real file and not just a  symlink. So, I need to test for the presence of a file, not of a symlink to avoid mysterious errors.

oharboe:
Ah. I will change to `ln -sf a b`, i.e. force overwrite. This mimics cp more closely.

tools/OpenROAD
maliberty, line 1: Why is this part of this PR?

maliberty:
I like the change to ln -sf

maliberty (approved):

maliberty:
One concern occurred to me after approving.  The use of abspath means you can't relocate the tree (eg in a bug report). Is there a reason it is required?

oharboe:
> One concern occurred to me after approving. The use of abspath means you can't relocate the tree (eg in a bug report). Is there a reason it is required?

I ran into problems with relative symlinks, so I switched to absolute path.

We have come this far with this PR in reviewing and diving into the possible effects, that I might as well do one more round of investigation on that.

oharboe:
Hmmmm.... need to investigate, when I use `ln -sf $(basename $<) $@` yyy, I get:

```
Error: floorplan.tcl, 3 cannot read file ./results/nangate45/gcd/base/1_synth.v.
```

maliberty:
> Hmmmm.... need to investigate, when I use `ln -sf $(basename $<) $@` yyy, I get:
> 
> ```
> Error: floorplan.tcl, 3 cannot read file ./results/nangate45/gcd/base/1_synth.v.
> Command exited with non-zero status 1
> ```

The link is in the same directory with the file so you probably need $(notdir $<)

oharboe:
@maliberty Relative paths. Some edgecases to be weary of. The source isn't always in the same folder as the destination, hence the somewhat convoluted command line. Strangely ln doesn't has an option to do this...

flow/Makefile
```diff
@@ -743,7 +743,7 @@ $(GDS_MERGED_FILE): $(RESULTS_DIR)/6_final.def $(OBJECTS_DIR)/klayout.lyt $(GDSO
 $(RESULTS_DIR)/6_final.v: $(LOG_DIR)/6_report.log
 
 $(GDS_FINAL_FILE): $(GDS_MERGED_FILE)
-	cp $< $@
+	ln -sf $(shell realpath --relative-to=$(dir $@) $<) $@
```
oharboe, line 746: @maliberty The relative path is relative to the destination folder of the symlink location.

PR #1002, opened 2023-04-20
DO NOT MERGE designs/asap7/mock-array-big: add asynchronous reset and soft (synchronous) reset

oharboe:
@yupferris @maliberty @tspyrou Test-case for asynchronous and synchronous reset ref. discussions.


maliberty:
Is there something still pending related to this?

oharboe:
> Is there something still pending related to this?

I will restore when we/if we get to this

PR #1003, opened 2023-04-21
make: add skip_resize target for mock abstracts

oharboe:


maliberty:
What is your goal with a mock abstract?  Without calling rsz the timing results will be mostly meaningless.

oharboe:
> What is your goal with a mock abstract? Without calling rsz the timing results will be mostly meaningless.

I am working on the top level floorplan and routing congestion, so I need pins in the right places and the correct dimensions.

oharboe:
@maliberty Ready to go, unrelated docker failure.

maliberty (approved):

PR #1004, opened 2023-04-23
Tighten timing + enable all violations to be fixed

openroadie:
Also exercises pin swapping


openroadie:
Need to wait until the resizer PR is merged.

rovinski:
It would be interesting if the new resizer passes had reports for the amount of improvement (sort of like HPWL for DPL).

maliberty:
rsz is merged.  This PR will need updating

PR #1009, opened 2023-04-25
Mock array big group order

oharboe:


maliberty:
Fails with
```
[ERROR PPL-0086] Both -region and -group constraints not allowed.
Error: io.tcl, 49 PPL-0086
```

oharboe:
> Fails with
> 
> ```
> [ERROR PPL-0086] Both -region and -group constraints not allowed.
> Error: io.tcl, 49 PPL-0086
> ```

Fixed.

flow/designs/asap7/mock-array-big/Element/io.tcl
```diff
@@ -50,7 +50,7 @@ proc zip {list1 list2} {
 foreach {direction direction2 names} $assignments {
     set mirrored [zip {*}$names]
     set_io_pin_constraint -region $direction:* -pin_names [lindex $names 0]
-    set_io_pin_constraint -group -order -pin_names [lindex $names 0]
```
oharboe, line 53: @eder-matheus Why didn't this fail before this fix?

eder-matheus: I don't understand the question. It should not fail. You can use `-group` and `-order` together.

flow/designs/asap7/mock-array-big/io.tcl
```diff
@@ -45,5 +45,5 @@ set assignments [list \
 ]
 
 foreach {direction names} $assignments {
-    set_io_pin_constraint -region $direction:* -pin_names $names
+    set_io_pin_constraint -order -region $direction:* -pin_names $names
```
eder-matheus: This is not correct, as the `-order` flag is only used with the `-group` flag. I think PPL will not error out, but the `-order` flag will do nothing here.

flow/designs/asap7/mock-array-big/Element/io.tcl
oharboe: OK.

flow/designs/asap7/mock-array-big/io.tcl
oharboe: @maliberty That's strange. It seems completely reasonable to want to specify the region(which edge) and that the pins should be ordered.

eder-matheus: We don't have support for it yet, but it can be implemented.

oharboe: I see. To order many pins on an edge unoptimized(fallback because of too many pins) is our main use-case.

oharboe:
@maliberty @eder-matheus If we are getting ordered pins on an edge, then I think it makes sense to wait for this PR, so that it can use that feature. In an array like this, it is pretty obvious that pins grouped, ordered and mirrored makes esthetic sense...

maliberty:
Marked as a draft while waiting.

oharboe:
Not the effect I was looking for :-)  This creates a lot of congestion congestion.

image

oharboe:
@eder-matheus @maliberty  Actually, the code on master, before your fix, allowed specifying -region -group and -order for the same set of pins, you just had to specify them on two separate set_io_constraint commands. 

All these options to set_io_constraint makes it devilishly hard to enumerate legal and illegal combinations, especially as you can work around some of the error checks by using multiple set_io_constraints, such as mock-array-big does on master.

PR #1014, opened 2023-04-27
util: genElapsedTime module and test adjustments

habibayassin:


flow/util/test/test_genElapsedTime.py
```diff
@@ -11,15 +11,17 @@
 from datetime import datetime
```
vvbandeira:
```py
from unittest.mock import MagicMock
import io
from datetime import datetime
import argparse  # argument parsing
from pathlib import Path
```
Let's cleanup these unused imports.

flow/util/test/test_genElapsedTime.py
```diff
@@ -28,23 +30,24 @@ def test_elapsed_time(self, mock_stdout):
             f.write("Some log entry\n")
             f.write("Elapsed time: 01:30:00[h:]min:sec.\n")
         # call the script with the test log file
-        sys.argv = ["util/genElapsedTime.py", "--logDir", str(self.tmp_dir.name)]
+        sys.argv = ["./genElapsedTime.py", "--logDir", str(self.tmp_dir.name)]
         with patch.object(sys, 'argv', sys.argv):
-                import genElapsedTime
+            module = importlib.import_module(self.module_name)
```
vvbandeira: Is this missing the `importlib.reload(module)` like on the other tests?

flow/util/test/test_genElapsedTime.py
vvbandeira, line 1: I think this file should be moved to `flow/test`.

vvbandeira (changes requested):

flow/util/test/test_genElapsedTime.py
habibayassin: I need to reload for the rest of the test cases but the first test case I only need importing

vvbandeira:
@habibayassin I think you missed the DCO check

vvbandeira (approved):

PR #1016, opened 2023-04-28
Swap tests

openroadie:


openroadie:
@maliberty : Another PR that should be merged after we have pin swapping code enabled.

cc: @tspyrou

PR #1030, opened 2023-05-03
setup: Automatically setup submodules if missing

pettermk:


setup.sh
```diff
@@ -10,6 +10,10 @@ if [ $EUID -ne 0 ]; then
   exit 1
 fi
 
+if git submodule status --recursive | grep -q "^-"; then
+  sudo -u $SUDO_USER git submodule update --init
```
maliberty: Why are you using sudo here?  It shouldn't be necessary.  Only the installer needs sudo.

pettermk: Makes sense - I've removed the sudo statement

setup.sh
```diff
@@ -10,6 +10,10 @@ if [ $EUID -ne 0 ]; then
   exit 1
 fi
 
+if git submodule status --recursive | grep -q "^-"; then
+  git submodule update --init
```
maliberty: Please add --recursive

pettermk: Done

maliberty (approved):

PR #1033, opened 2023-05-03
autotuner: fix error due to use of deprecated module

oharboe:
Fixes:

DeprecationWarning: The module `ray.tune.suggest` has been moved to `ray.tune.search` and the old location has been deprecated. Please adjust your imports to point to the new location. Example: Do a global search and replace `ray.tune.suggest` with `ray.tune.search`.


maliberty:
@vvbandeira is there any issue to update this?  I just want make sure our ray setup is current enough that this won't be a problem.

oharboe:
@vvbandeira Ping?

oharboe:
@maliberty No response, I'm taking this off our radar.

PR #1034, opened 2023-05-03
fix: Change TNS default to 5 from 0

openroadie:
Changing default from 0 to 5 for TNS end percent
cc: @tspyrou


flow/scripts/cts.tcl
```diff
@@ -74,6 +74,9 @@ if { [info exists ::env(HOLD_SLACK_MARGIN)] && $::env(HOLD_SLACK_MARGIN) > 0.0}
 if { [info exists ::env(TNS_END_PERCENT)] && $::env(TNS_END_PERCENT) > 0.0} {
   puts "TNS end percent $::env(TNS_END_PERCENT)"
   append additional_args " -repair_tns $::env(TNS_END_PERCENT)"
+} else {
+  puts "Default TNS end percent 5"
+  append additional_args " -repair_tns 5"
 }
```
maliberty: It would be simpler to drop the whole if statement and do 
```
export TNS_END_PERCENT ?= 5
```
in the Makefile to set the default and allow overriding.

openroadie: But then TNS is set in two places .... the makefile and the cts.tcl file. But I don't have a strong opinion on this. Let me change it.

maliberty: Why would it be set in cts.tcl?

openroadie: It is parsed/mentioned in two places... sorry for not being clear.I tried "export TNS_END_PERCENT ?= 5 in the Makefile and the cts.tcl script did not like it (no message printed to show that it is using a non zero TNS value). Will need to figure out why.

maliberty: There is no need to parse it in cts.tcl afaik.

maliberty (approved):

PR #1045, opened 2023-05-04
changed sed flags from gnu-specific '-r' to posix '-E'

jankcorn:
Updated sed flag usage for portability.
On MacOS sed, the gnu-specific flag is not supported.


maliberty (approved):
Thanks for the fix.

PR #1051, opened 2023-05-05
designs/asap7/mock-array-big/Element: restricted clk-io-pin into specific region in the grid

nerriav:
Currently, when we group (by order) the IO pins, the tool aggregates them in one edge (for example, in the element, when region is top - the middle of it). W'd like to have the ability to specify exactly where the pins group would be located (along the region)
Tried using the this command : (from openroad PIN_CONSTRAINTS manual)
``` -region up:{llx lly urx ury} ``` 

Floorplan show the following error :
``` 
[ERROR STA-0406] set_io_pin_constraint positional arguments not supported.
Error: io.tcl, 54 STA-0406
```
Is there any way to force the exact location of pins group ?

FYI @oharboe

### DIFF
--- a/flow/designs/asap7/mock-array-big/Element/io.tcl
+++ b/flow/designs/asap7/mock-array-big/Element/io.tcl
@@ -51,4 +51,4 @@ foreach {direction direction2 names} $assignments {
     set_io_pin_constraint -mirrored_pins $mirrored
 }
 
-set_io_pin_constraint -region top:* -pin_names clock
+set_io_pin_constraint -region top:{0 0 3 0} -pin_names clock

### CODE REVIEW

File: flow/designs/asap7/mock-array-big/Element/io.tcl [hunk 1]
```diff
@@ -51,4 +51,4 @@ foreach {direction direction2 names} $assignments {
     set_io_pin_constraint -mirrored_pins $mirrored
 }
 
-set_io_pin_constraint -region top:* -pin_names clock
+set_io_pin_constraint -region top:{0 0 3 0} -pin_names clock
```
vijayank88, line 54: @nerriav 
Can you check with below format?
```
set_io_pin_constraint -direction input -region up:{0 0 3 0} -pin_names clock
```

nerriav: same error massage ):

vijayank88: @maliberty 
Can you check this?

vijayank88: Refer this [example](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/ppl/test/add_constraint8.tcl).
```
set_io_pin_constraint -region top:0-3 -pin_names clock
```

nerriav: Thanks !!!. it worked.
@eder-matheus 
documentation is a little bit confusing here , probably need an update
https://github.com/The-OpenROAD-Project/OpenROAD/blob/86d1769f8d658dd25ee55c1adc5405d4e76c5e97/src/ppl/README.md?plain=1#L109

vijayank88: @nerriav 
Try above constraint, it resolve your issue and place pin withing 0-3um region.
in log `[INFO PPL-0048] Restrict pins [clock] to region 0.0u-3.0u at the top edge.`
Screenshot from 2023-05-05 17-30-05

vijayank88, line 54: set_io_pin_constraint -region top:0-3 -pin_names clock

vijayank88 (changes requested):

maliberty:
Individual pins can be placed with https://github.com/The-OpenROAD-Project/OpenROAD/tree/master/src/ppl#place-individual-pin but I don't think there is a group equivalent.

maliberty:
@eder-matheus is there a correct syntax here?

flow/designs/asap7/mock-array-big/Element/io.tcl
eder-matheus: @nerriav The `-region up` is used only when placing the pins in the grid created with `define_pin_shape_pattern `, and it's not for the boundaries of the edge. I'll modify the documentation to make it clear.

eder-matheus:
> @eder-matheus is there a correct syntax here?

Currently, we don't have a command to place a group in a specific position. I think the solution proposed by @vijayank88 is the best way we have now.

oharboe:
> > @eder-matheus is there a correct syntax here?
> 
> Currently, we don't have a command to place a group in a specific position. I think the solution proposed by @vijayank88 is the best way we have now.

@nerriav and I worked on this today. We dont need an exact location, what is currently implemented appears to work. AFAIK, only the documentation needs an update.

oharboe:
This was part of a discussion that has been resolved.

PR #1066, opened 2023-05-13
mock-array-big: get_ports caveats addressed

oharboe:
- use natural sort on names for io pins, get_ports returns in undefined order
- regex in get_ports is not a Tcl regex, but a C++ regex, there are differences, so use Tcl regex


oharboe:
@maliberty unrelated failure 
image

vijayank88:
I'm also facing same failure.

maliberty (approved):

maliberty:
The CI issue is clearly unrelated - I'll update the OR submodule to resolve that (@vijayank88 )

PR #1076, opened 2023-05-18
global route: change default iterations to 20

oharboe:
a 100 iterations can take a very long time. If global route doesn't complete in 20, it normally requires some attention from the developer to find out what is wrong.

I have a guess, easily checked by this draft PR, is that none of the designs in the test-suite need very many global route iterations.

### DIFF
--- a/flow/designs/sky130hd/microwatt/config.mk
+++ b/flow/designs/sky130hd/microwatt/config.mk
@@ -30,3 +30,5 @@ export CTS_CLUSTER_DIAMETER = 100
 export CTS_CLUSTER_SIZE = 30
 
 export export SETUP_SLACK_MARGIN = 0.2
+
+export GLOBAL_ROUTE_ARGS=-congestion_iterations 100 -verbose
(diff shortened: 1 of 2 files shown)


maliberty:
Apparently 10 is not enough.

oharboe:
@maliberty Curiosity satisfied, unless you think it makes sense to reduce the default to say 20 and add an exception to the single failing case.

oharboe:
@maliberty I think this is good to go. Fast failure for global routing by default makes ORFS a bit more user-friendly, I think. Only one of the designs don't pass at 20 iterations, so it seems like a pretty high default still.

maliberty:
It used to be 200 (https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/e72959cff5a1cce98ed7cd1301433e13e5b0ec84) 

Do you know at what level all designs pass?  I'll also have to run the private designs to see the impact there too.

oharboe:
> It used to be 200 ([e72959c](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/e72959cff5a1cce98ed7cd1301433e13e5b0ec84))

I see.

> Do you know at what level all designs pass? 

ref commit comments, there is one design that doesnt pass at 40 sky130hd microwatt, so I set that one to 100(same value as before) by modifying config.mk


> I'll also have to run the private designs to see the impact there too.

Perhaps a few could set the GLOBAL_ROUTE_ARGS to a larger number of iterations.

If no designs set that variable, I am inclined to think that the default is too high?

maliberty (approved):

PR #1091, opened 2023-05-24
docker: add a docker_shell script to run GUI from within docker image

oharboe:
Building example design using Docker image
==========================================

The `or` script is used as a prefix to launch a command within the OpenROAD docker image.

This script is set up to run the OpenROAD GUI from
within the Docker image on the host platform.

Also, the current working directory is mapped into the Docker image using the current user's credentials.

Build docker image
------------------

First build the docker image:

```
cd OpenROAD-flow-scripts
sudo ./setup.sh
./build_openroad.sh
```

Build an example design and run the GUI:

```
cd flow
./or make
./or make gui_final
```


flow/README.md
```diff
@@ -0,0 +1,28 @@
...
+Also, the current working directory is mapped into the Docker image using the current user's credentials.
+
+Build docker image
+------------------
+
+First build the docker image:
+
+```
+cd OpenROAD-flow-scripts
+sudo ./setup.sh
```
vijayank88: For docker build, it is not needed. @vvbandeira can you confirm it?

maliberty:
From @rovinski in #1092 

> I think that contributed helper scripts should probably not go in the top-level directory. It can get quite cluttered there. Maybe in /etc or /flow/util. Also I think or isn't a very descriptive name of what the script does. Maybe docker_shell or similar.

I agree on both points.

flow/or
```diff
@@ -0,0 +1,35 @@
...
+ -e DISPLAY=$DISPLAY \
+ -e QT_XKB_CONFIG_ROOT=/usr/share/X11/xkb \
+ -v $XSOCK:$XSOCK \
+ -v $XAUTH:$XAUTH \
+ -e XAUTHORITY=$XAUTH \
+ -e FLOW_HOME=/OpenROAD-flow-scripts/flow/ \
+ -e MAKEFILES=/OpenROAD-flow-scripts/flow/Makefile \
+ -v $WORKSPACE:`pwd` \
+ $DOCKER_INTERACTIVE \
+ openroad/flow-centos7-builder:latest \
```
maliberty: The image name should be a parameter with a default as it is in etc/DockerHelper.sh

flow/or
```diff
@@ -0,0 +1,35 @@
...
+ bash -c "set -ex
+ mkdir /tmp/xdg-run
```
maliberty: why is this needed?

maliberty (approved):

PR #1105, opened 2023-05-30
check klayout version for centos

### LINKED ISSUE #1080: setup.sh should not exit if required package exists
### Subject

### Describe the bug

Trying to run latest `sudo ./setup.sh` to install dependencies.
Facing following error and package installer exit as KLayout already installed.
```
klayout-0.27.10-0.x86_64.rpm                                                                                                                                           |  29 MB  00:00:03     
Examining /var/tmp/yum-root-yDnTtz/klayout-0.27.10-0.x86_64.rpm: klayout-0.27.10-0.x86_64
/var/tmp/yum-root-yDnTtz/klayout-0.27.10-0.x86_64.rpm: does not update installed package.
Error: Nothing to do
```
Please update the dependency installer script to not to exit if required or above version already installed.

### Expected Behavior

dependency installer script to not to exit if required or above version already installed.

### Environment

```shell
With latest master
[WARNING] Your current OpenROAD version is outdated.
It is recommened to pull the latest changes.
If problem persists, file a github issue with the re-producible test case.
kernel: Linux 3.10.0-1160.49.1.el7.x86_64
os: CentOS Linux 7 (Core)
cmake version 3.24.2
```

### To Reproduce

Install KLayout and  try running
```
sudo ./setup.sh
```

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

vijayank88:
Centos exit the dependency installer script if KLayout already exists.
This PR will check for existing KLayout version, only below required version installed mean, it will remove and install required version.
If KLayout not installed already, install required version by default.

fixes #1080


maliberty:
I'll leave this to @vvbandeira to comment/approve

vijayank88:
`bp_multi_top/nangate45 ` metrics fails
```
[ERROR] finish__timing__setup__ws fail test: -1.88859 >= -1.88
```
@maliberty It causing all latest PR fails. Shall I update the metrics as well?

vijayank88:
@vvbandeira 
FYI

etc/DependencyInstaller.sh
```diff
@@ -39,7 +39,21 @@ _installCentosPackages() {
...
+      currentver=$(echo $currentver | tr -d '.')
+      requiredver=$(echo $klayoutVersion | tr -d '.')
+      if [ "$currentver" -ge "$requiredver" ]; then
+        echo "KLayout version greater than or equal to ${klayoutVersion}"
+      else
+        echo "KLayout version less than ${klayoutVersion}"
+        sudo yum remove -y klayout
+        yum install -y https://www.klayout.org/downloads/CentOS_7/klayout-${klayoutVersion}-0.x86_64.rpm
+      fi 
+    fi
```
vvbandeira: I think this should be refactored into a function that takes the URL as an argument so we can use it with other OSes -- then add it to the Ubuntu install functions. Also, replace for readability `currentver` -> `currentVersion` and `requiredver` `requiredVersion`

vvbandeira (changes requested):

etc/DependencyInstaller.sh
vijayank88: @vvbandeira 
I checked with Ubuntu it automatically update or overwrite with KLayout installation. 
Only for Centos 7 currently it is exit from installer script.
```
klayout-0.28.8-0.x86_64.rpm                                                                                                                                    |  31 MB  00:00:03     
Examining /var/tmp/yum-root-yDnTtz/klayout-0.28.8-0.x86_64.rpm: klayout-0.28.8-0.x86_64
/var/tmp/yum-root-yDnTtz/klayout-0.28.8-0.x86_64.rpm: does not update installed package.
Error: Nothing to do
```

vvbandeira: Ok, then you can skip the refactoring. But please fix the readability issue above.

---

Also, there's a problem with your regex/logic. One example where it would fail:
- Version installed on user's system = 0.27.13
- Version set by klayoutVersion = 0.28.1

In this example, the currentver variable would be 02713 and the requiredver variable 0281. The if statement would say that the installed version is greater than the required version.

The above example is possible with the current versioning scheme used by klayout, see below:

image

vijayank88: @vvbandeira 
Thanks for catching the regex logic for >=. I will change equation.

vijayank88: @vvbandeira 
Can you check the regex logic now?
added `_versionCompare`

etc/DependencyInstaller.sh
```diff
@@ -39,7 +45,20 @@ _installCentosPackages() {
...
         tcl-devel
 
-    yum install -y https://www.klayout.org/downloads/CentOS_7/klayout-${klayoutVersion}-0.x86_64.rpm
+    if ! [ -x "$(command -v klayout)" ]; then
+      yum install -y https://www.klayout.org/downloads/CentOS_7/klayout-${klayoutVersion}-0.x86_64.rpm
+    else
+      currentVersion="$(klayout -v)"
+      currentVersion=$(echo $currentVersion | grep -o '[0-9]\+\.[0-9]\+\.[0-9]\+')
+      requiredVersion=$(echo $klayoutVersion)
+      if _versionCompare $currentVersion -ge $requiredVersion; then
```
```suggestion
      if _versionCompare $currentVersion -ge $klayoutVersion; then
```

etc/DependencyInstaller.sh
```diff
@@ -39,7 +45,20 @@ _installCentosPackages() {
         ruby-devel \
         tcl-devel
 
-    yum install -y https://www.klayout.org/downloads/CentOS_7/klayout-${klayoutVersion}-0.x86_64.rpm
+    if ! [ -x "$(command -v klayout)" ]; then
+      yum install -y https://www.klayout.org/downloads/CentOS_7/klayout-${klayoutVersion}-0.x86_64.rpm
+    else
+      currentVersion="$(klayout -v)"
+      currentVersion=$(echo $currentVersion | grep -o '[0-9]\+\.[0-9]\+\.[0-9]\+')
```
```suggestion
      currentVersion=$(klayout -v | grep -o '[0-9]\+\.[0-9]\+\.[0-9]\+')
```

etc/DependencyInstaller.sh
```diff
@@ -12,6 +12,12 @@ fi
 # package versions
 klayoutVersion=0.28.8
 
+_versionCompare(){
+   local a b IFS=. ; set -f
+   printf -v a %08d $1; printf -v b %08d $3
+   test $a "$2" $b
+}
+
```
```suggestion
_versionCompare() {
    local a b IFS=. ; set -f
    printf -v a %08d $1; printf -v b %08d $3
    test $a "$2" $b
}

```

vvbandeira (changes requested):

vvbandeira (approved):

PR #1110, opened 2023-05-31
Asap7 Platform With Fakeram Memory Model

louiic:
Created additional memory models / macros with Fakeram.


oharboe (changes requested):
Add a README.md to explain what this is inside flow/platforms/asap7/

Add Verilog behavioral model so we know what these RAMs are doing

I would rebase on master and squash to start out with a single commit before reviewing

oharboe:
@louiic I suggest doing an interactive rebase to remove all the merge commits with master and probably squash to a single commit. The merge commits are problematic to review, because it is hard to know why they are there and this pull request should not e.g. merge in a new tools/OpenROAD + it makes the history harder to follow on master.

```
git fetch origin
git rebase origin/master -i
# change "pick" to "squash" to squash commits
```

Now you can force update your branch:

```
git push origin HEAD:asap7-mem --force
```

Other than that, the pull request seems ready to be reviewed by @maliberty  to go into master.

I assume you will use this in a subsequent pull request, but I think it is fine to merge this with master first and then add tests/examples later.

maliberty (approved):

PR #1114, opened 2023-06-02
Remove PRIVATE_DIR variable from config.mk

louiic:
PRIVATE_DIR is a reserved variable - need to update OpenROAD        
Remove from config.mk - both mock-array-big and Element
Pull in target from private.mk, then remove private.mk

### DIFF
--- a/flow/designs/asap7/mock-array-big/Element/config.mk
+++ b/flow/designs/asap7/mock-array-big/Element/config.mk
@@ -23,8 +23,6 @@ export PDN_TCL = designs/asap7/mock-array-big/Element/pdn.tcl
 # to the ring and stipe
 export MAX_ROUTING_LAYER = M5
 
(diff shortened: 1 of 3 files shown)


oharboe (changes requested):
Please dont merge, this breaks the "make verilog" target.

This target is used to generate Verilog from Chisel using the settings.mk or defaults.mk parameters.

oharboe (approved):
Moving the verilog target into the config.mk is less convoluted than using the private.mk feature. 👍

oharboe:
@louiic I am curious. This is cleaner, but I what motivated this cleanup? Why is it necessarry?

louiic:
Cleanup is by-product!

PRIVATE_DIR is used to hook in other scripts and apps.

We (I) should update the readme for ORFS.

PR #1146, opened 2023-06-17
jenkins Url, add branches/commits, add rules, nightly tag

habibayassin:


vvbandeira (approved):
Missing DCO, but otherwise looks ok.

PR #1152, opened 2023-06-20
update yosys to 0.30

habibayassin:


vvbandeira (commented):
The compilation error needs investigation and fixing. Coul you also try to remove the hard-coded ABC_REV from the build script?

maliberty:
Stale: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1468

PR #1163, opened 2023-06-20
makefile: add vars target

oharboe:
@maliberty This allows generating vars.sh/tcl/gdb files, which is useful when debugging, openroad command line can be copied and pasted after sourcing vars.sh.

Also, it could be useful in separating the concern of executing ORFS stages and checking dependencies, opening the road implementing Bazel support on top of ORFS with a minimal maintenance burden for ORFS and the Bazel support.

### DIFF
--- a/flow/.gitignore
+++ b/flow/.gitignore
@@ -1,1 +1,4 @@
 settings.mk
+vars.sh
+vars.gdb
+vars.tcl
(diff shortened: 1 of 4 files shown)


oharboe:
@maliberty Ready to go, I tested `make floorplan_issue` and saw no errors and the files looked OK.

vvbandeira (approved):

maliberty (approved):

PR #1169, opened 2023-06-21
makefile: first example on how to move logic from makefile to tcl

oharboe:
@maliberty moving logic from Makefile to tcl is a step in the direction of separating the concern of dependencies from how to execute ORFS stages.

### DIFF
--- a/flow/scripts/io_placement_random.tcl
+++ b/flow/scripts/io_placement_random.tcl
@@ -1,4 +1,4 @@
-if {![info exists ::env(IS_CHIP)]} {
+if {[info exists ::env(IS_CHIP)]} {
   file copy -force $::env(RESULTS_DIR)/2_1_floorplan.odb $::env(RESULTS_DIR)/2_2_floorplan_io.odb
   # Ouch! The file date is copied by Tcl, but the precision is seconds, so
   # this means that a copied file can be *older* than the source, or
--- a/flow/scripts/io_placement_random.tcl
+++ b/flow/scripts/io_placement_random.tcl
@@ -1,3 +1,12 @@
+if {![info exists ::env(IS_CHIP)]} {
+  file copy -force $::env(RESULTS_DIR)/2_1_floorplan.odb $::env(RESULTS_DIR)/2_2_floorplan_io.odb
+  # Ouch! The file date is copied by Tcl, but the precision is seconds, so
+  # this means that a copied file can be *older* than the source, or
+  # worse, older than some *other* dependency for this target.
+  exec touch $::env(RESULTS_DIR)/2_2_floorplan_io.odb
+  exit
+}
+
 source $::env(SCRIPTS_DIR)/load.tcl
 load_design 2_1_floorplan.odb 1_synth.sdc "Starting random IO placement"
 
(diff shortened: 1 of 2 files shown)


flow/scripts/io_placement_random.tcl
```diff
@@ -1,3 +1,12 @@
+if {![info exists ::env(IS_CHIP)]} {
```
maliberty: Is IS_CHIP exported?

flow/scripts/io_placement_random.tcl
```diff
@@ -1,3 +1,12 @@
+if {[info exists ::env(IS_CHIP)]} {
+  file copy -force $::env(RESULTS_DIR)/2_1_floorplan.odb $::env(RESULTS_DIR)/2_2_floorplan_io.odb
```
maliberty, line 2: It looks like it creates a new file https://github.com/tcltk/tcl/blob/9a18abc8558016b133faab39ea7702494c902069/unix/tclUnixFCmd.c#L541

I'm not clear why this would happen.  Would you clarify how this precision issue happens.

oharboe: Tcl copies file date and truncates date to seconds...

This breaks for two reasons:

1. truncated date is older than full date
2. if the copied file is out of date in the makefile due to a dependency other than the file being copied, then the file will be out of date even the date is copied with full precision.

Example. If bar2 is newer than bar, then if the date of bar is copied, foo will still be out of date if the date is copied:

```
foo: bar bar2
    cp bar foo
```





```
oyvind@ogre:~/ascenium/hardware/openroad$ touch blah.txt
oyvind@ogre:~/ascenium/hardware/openroad$ stat blah.txt
  File: blah.txt
  Size: 0         	Blocks: 0          IO Block: 4096   regular empty file
Device: 10302h/66306d	Inode: 102825130   Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1000/  oyvind)   Gid: ( 1000/  oyvind)
Access: 2023-06-23 08:44:31.010655312 +0200
Modify: 2023-06-23 08:44:31.010655312 +0200
Change: 2023-06-23 08:44:31.010655312 +0200
 Birth: 2023-06-23 08:44:31.010655312 +0200
$ tclsh
% file copy -force blah.txt foo
% 
$ stat foo
  File: foo
  Size: 0         	Blocks: 0          IO Block: 4096   regular empty file
Device: 10302h/66306d	Inode: 102825866   Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1000/  oyvind)   Gid: ( 1000/  oyvind)
Access: 2023-06-23 08:44:31.000000000 +0200
Modify: 2023-06-23 08:44:31.000000000 +0200
Change: 2023-06-23 08:44:55.078214243 +0200
 Birth: 2023-06-23 08:44:55.078214243 +0200
```

Above you can see:

1. the file date is copied
5. precision is cut to second

```
$ touch foo
$ stat foo
  File: foo
  Size: 0         	Blocks: 0          IO Block: 4096   regular empty file
Device: 10302h/66306d	Inode: 102825866   Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1000/  oyvind)   Gid: ( 1000/  oyvind)
Access: 2023-06-23 08:45:12.225899982 +0200
Modify: 2023-06-23 08:45:12.225899982 +0200
Change: 2023-06-23 08:45:12.225899982 +0200
 Birth: 2023-06-23 08:44:55.078214243 +0200
```

maliberty:
This reminds me how much I hate tcl.  I would just do 'exec cp <from> <to>' instead.

oharboe:
> This reminds me how much I hate tcl. I would just do 'exec cp ' instead.

This change is moot if we go for my other option to separate dependency checking and target execution...

oharboe:
#1175 is more promising

PR #1175, opened 2023-06-23
Demonstrate separation of concerns: execution of a flow stage and dependency checking

oharboe:
With this sequence of changes, it is possible to execute a step without checking dependencies, without copy & paste of code in the Makefile:

```
make do-2_1_floorplan.odb
```


oharboe:
@maliberty No relevant failures in CI, ready for review.

maliberty:
Would you explain the intent behind UNSET_AND_MAKE

oharboe:
> Would you explain the intent behind UNSET_AND_MAKE

If invoke $(MAKE), then some emvironment variables will be added to, rather than set up to scratch. This, minimally, leads to warnings.

oharboe:
@maliberty Force pushed rebase on top of master so that this pull request has a single concern.

If we want to go for this solution to separate the dependency checking and execution of a stage, the change illustrated will have to be done for all the targets. It is a simple change, if a bit repetitive.

maliberty:
I like that it avoid repetition.  I'm wondering if the rule pair couldn't be wrapped up in a single reusable macro.

oharboe:
> I like that it avoid repetition. I'm wondering if the rule pair couldn't be wrapped up in a single reusable macro.

Will try something like:

```
define floorplan_template
$(1): $(2)
	$(UNSET_AND_MAKE) $(3)

.PHONY: $(3)
$(3):
	($(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/floorplan.tcl -metrics $(LOG_DIR)/$$(basename $$@).json) 2>&1 | tee $(LOG_DIR)/$$(basename $$@).log
endef

$(eval $(call floorplan_template,$(RESULTS_DIR)/2_1_floorplan.odb, $(RESULTS_DIR)/1_synth.v $(RESULTS_DIR)/1_synth.sdc $(TECH_LEF) $(SC_LEF) $(ADDITIONAL_LEFS) $(FOOTPRINT) $(SIG_MAP_FILE) $(FOOTPRINT_TCL), do-2_1_floorplan.odb))
```

oharboe:
@maliberty I messed around with macros, but makefile macros are so convoluted with multiple evaluation stages and layers of escaping rules that I think KISS applies and that the current articulation can't be improved much on overall with make...

maliberty:
Make isn't fun. Do you want to do the mechanical changes then?

oharboe:
> Make isn't fun.

ChatGPT makes.editing makefiles fascinating. It is amazingly good with arcane languages that have a lot of examples...
 
Try copying and pasting some cryptic makefile snippets and ask it to explain or rewrite...

I wonder if we have seen the end of new languages?

If we rely on automating programming then a language will not be widely adopted until a lot of examples exist....

> Do you want to do the mechanical changes then?

I will.have one more go at it, then cave in and make the changes by hand.

oharboe:
@maliberty If we enforce some more concistency in naming of .odb, .log, .json files and push conditional code from Makefile to .tcl, it looks like we can have a short and sweet macro. 

If we agree on the direction, I can start making pull requests to move is in the right direction. I'm doing this as time permits and I think we will hit some cleanup along the way, so I think independent pull requests to address the concerns and merge them on an ongoing basis makes sense...

oharboe:
@maliberty This should be fine as far as it goes towards separating the concerns of executing a flow stage and dependency checking. This is a complete pass of one more of the things we need to do. If the computer says yes, it is good to go for my part.

oharboe:
@maliberty passes, failure is unrelated(macos). Ready to go for my part.

flow/Makefile
```diff
@@ -447,11 +447,31 @@ endif
...
+do-$(1):
+	($(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/$(3).tcl -metrics $(LOG_DIR)/$(1).json) 2>&1 | tee $(LOG_DIR)/$(1).log
+endef
+
 # STEP 1: Translate verilog to odb
 #-------------------------------------------------------------------------------
-$(RESULTS_DIR)/2_1_floorplan.odb: $(RESULTS_DIR)/1_synth.v $(RESULTS_DIR)/1_synth.sdc $(TECH_LEF) $(SC_LEF) $(ADDITIONAL_LEFS) $(FOOTPRINT) $(SIG_MAP_FILE) $(FOOTPRINT_TCL)
-	($(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/floorplan.tcl -metrics $(LOG_DIR)/2_1_floorplan.json) 2>&1 | tee $(LOG_DIR)/2_1_floorplan.log
-
+$(eval $(call do-step,2_1_floorplan,$(RESULTS_DIR)/1_synth.v $(RESULTS_DIR)/1_synth.sdc $(TECH_LEF) $(SC_LEF) $(ADDITIONAL_LEFS) $(FOOTPRINT) $(SIG_MAP_FILE) $(FOOTPRINT_TCL),floorplan))
```
maliberty, line 474: 2_2 & 2_3 are not converted.

flow/Makefile
```diff
@@ -681,8 +688,7 @@ elapsed:
 # ==============================================================================
 
 ifneq ($(USE_FILL),)
-$(RESULTS_DIR)/6_1_fill.odb: $(RESULTS_DIR)/5_route.odb $(FILL_CONFIG)
-	($(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/density_fill.tcl -metrics $(LOG_DIR)/6_1_fill.json) 2>&1 | tee $(LOG_DIR)/6_1_fill.log
+$(eval $(call do-step,6_1_fill,$(RESULTS_DIR)/5_route.odb $(FILL_CONFIG),density_fill))
```
maliberty, line 691: final_report isn't converted

maliberty (approved):

flow/Makefile
oharboe: 6_report.log is now also converted.

PR #1207, opened 2023-07-07
Make BLOCKS is idempotent

oharboe:
Variables at the top level should not be visible inside the macros that are built by the BLOCKS macros. If they are, then building the macro standalone is not the same as the macro being built by the top level macro.

In other words, these two sequences should be idempontent and I believe they are with this change, regardless of what variables are defined at the top level `config.mk`:

```
make DESIGN_CONFIG=designs/asap7/mock-array/config.mk
```

```
make DESIGN_CONFIG=designs/asap7/mock-array/Element/config.mk generate_abstract
make DESIGN_CONFIG=designs/asap7/mock-array/config.mk
```

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -672,7 +672,7 @@ finish: $(LOG_DIR)/6_report.log \
         $(RESULTS_DIR)/6_final.v \
         $(RESULTS_DIR)/6_final.sdc \
         $(GDS_FINAL_FILE)
-	-@$(UTILS_DIR)/genElapsedTime.py -d "$(LOG_DIR)"
+	$(UNSET_AND_MAKE) elapsed
 
 .PHONY:
 elapsed:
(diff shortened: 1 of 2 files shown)


flow/Makefile
```diff
@@ -672,7 +672,7 @@ finish: $(LOG_DIR)/6_report.log \
         $(RESULTS_DIR)/6_final.v \
         $(RESULTS_DIR)/6_final.sdc \
         $(GDS_FINAL_FILE)
-	-@$(UTILS_DIR)/genElapsedTime.py -d "$(LOG_DIR)"
+	$(UNSET_AND_MAKE) elapsed
```
oharboe, line 675: This is an example of how `$(MAKE)` would cause problems, but where `$(UNSET_AND_MAKE)` works as expected.

oharboe:
@maliberty the macos testinstaller failure is unrelated

maliberty (approved):

PR #1211, opened 2023-07-08
Plot running times

oharboe:
I was hoping to plot detailed routing time as a function of area, but so far this script has been useful to generate various test-cases(segfaults or detailed routing errors), so I thought it would be useful to share.


oharboe:
@maliberty The first commit is in master already, so this PR has only the utility script

flow/util/plot-running-times.py
```diff
@@ -0,0 +1,116 @@
...
+        # 3_3_place_gp                       5
+        # 3_4_resizer                        2
+        # 3_5_opendp                         3
+        # 4_1_cts                            5
+        # 4_2_cts_fillcell                   3
+        # 5_1_fastroute                      5
+        # 5_2_TritonRoute                  297
+        # 6_1_merge                          3
+        # 6_report                          67
+        pattern = r'^5_2_TritonRoute\s+(\d+)$'
```
maliberty: Now 5_2_route.log

oharboe: Fixed

flow/util/plot-running-times.py
```diff
@@ -0,0 +1,106 @@
+import subprocess
```
maliberty: This script is specific to mock array. It should therefore be in the design dir

oharboe:
@maliberty Good to go.

maliberty (changes requested):
Sorry I reviewed but forgot to hit submit.

Are all the non-plotting changes intended here?

oharboe:
> Sorry I reviewed but forgot to hit submit.
> 
> Are all the non-plotting changes intended here?

Those are already in master.

maliberty (approved):

PR #1217, opened 2023-07-10
gcd typo fix

vijayank88:
Typo fixed `export PLACE_DENISTY_LB_ADDON = 0.1` to `export PLACE_DENSITY_LB_ADDON = 0.1`


maliberty:
The net effect of fixing the typo is to make the timing metric worse.  I think we should just remove PLACE_DENSITY_LB_ADDON entirely instead.

vijayank88:
@maliberty 
Updated.

maliberty (approved):

PR #1220, opened 2023-07-10
Mock array route by abutment

oharboe:
@louiic @rovinski FYI, routing by abutment. Seems to work fine on my machine when I mare the nets between element macros `dont_touch`

image


flow/designs/asap7/mock-array/constraints.sdc
```diff
@@ -13,3 +13,5 @@ set non_clock_inputs [lsearch -inline -all -not -exact [all_inputs] $clk_port]
 
 set_input_delay  [expr $clk_period * $clk_io_pct] -clock $clk_name $non_clock_inputs 
 set_output_delay [expr $clk_period * $clk_io_pct] -clock $clk_name [all_outputs]
+
+set_dont_touch [get_nets ces_*_*]
```
oharboe, line 17: Without this, I this step will eventually fail. I hit ctrl-c after a few minutes, because I didn't want to wait... With the fix above, this step completes successfully instantly.

```
[INFO RSZ-0032] Inserted 10350 hold buffers.
^Cmake: *** [Makefile:579: results/asap7/mock-array/base/4_1_cts.odb] Error 130

oyvind@ogre:~/OpenROAD-flow-scripts/flow$ git diff
diff --git a/flow/designs/asap7/mock-array/constraints.sdc b/flow/designs/asap7/mock-array/constraints.sdc
index ad66e12c..dc3df1c0 100644
--- a/flow/designs/asap7/mock-array/constraints.sdc
+++ b/flow/designs/asap7/mock-array/constraints.sdc
@@ -14,4 +14,4 @@ set non_clock_inputs [lsearch -inline -all -not -exact [all_inputs] $clk_port]
 set_input_delay  [expr $clk_period * $clk_io_pct] -clock $clk_name $non_clock_inputs 
 set_output_delay [expr $clk_period * $clk_io_pct] -clock $clk_name [all_outputs]
 
-set_dont_touch [get_nets ces_*_*]
+#set_dont_touch [get_nets ces_*_*]
```

rovinski:
If `get_nets ces_*_*` covers all the nets that are connected by abutment then yes, that's the right thing to do.

You may find that you can make it through the flow, but still fail timing at the end. If so, then you need to update the setup and/or hold constraints inside the element block and try again.

oharboe:
> If `get_nets ces_*_*` covers all the nets that are connected by abutment then yes, that's the right thing to do.
> 
> You may find that you can make it through the flow, but still fail timing at the end. If so, then you need to update the setup and/or hold constraints inside the element block and try again.

I see. If I need hold cells there, then something is wrong and the tools will tell me. The fix should indeed go into the elements...

I suspect that a differently articulated .sdc file for the elements could avoid the hold cells, but I don't know how to do that... Perhaps the OpenROAD tools need this final push across the finish line?

Thanks!

oharboe:
@maliberty detailed routing time before(looked up a different PR). global routing is a hair slower and detailed routing a hair faster... No overall change really.

```
Log                       Elapsed seconds
1_1_yosys                          6
2_1_floorplan                      2
2_2_floorplan_io                   2
2_3_tdms_place                     1
2_4_floorplan_macro                1
2_5_floorplan_tapcell              1
2_6_floorplan_pdn                 50
3_1_place_gp_skip_io               2
3_2_place_iop                      2
3_3_place_gp                      13
3_4_place_resized                  7
3_5_place_dp                       6
4_1_cts                           13
4_2_cts_fillcell                   4
5_1_grt                           14
5_2_route                        427
6_1_merge                          7
6_report                          90
```

After:


```
Log                       Elapsed seconds
1_1_yosys                         10
2_1_floorplan                      3
2_2_floorplan_io                   3
2_3_tdms_place                     3
2_4_mplace                         2
2_5_tapcell                        2
2_6_pdn                           92
3_1_place_gp_skip_io               2
3_2_place_iop                      3
3_3_place_gp                      10
3_4_resizer                        8
3_5_opendp                         9
4_1_cts                           19
4_2_cts_fillcell                   5
5_1_fastroute                     40
5_2_TritonRoute                  412
6_1_merge                          4
6_report                          28
```

oharboe:
@osamahammad21 FYI

rovinski:
> Perhaps the OpenROAD tools need this final push across the finish line?

It's actually a quite simple mechanical process. Lets say that you run the top-level and find a hold time violation of 0.1 ns from `element_0/out` to `element_1/in` where `out` and `in` are generic pin names.

This means that between the two pins, you need to constrain them 0.1 ns tighter. This could either be by adding:
1. `set_output_delay -min 0.1` 
2. `set_input_delay -min 0.1`
3. `set_output_delay -min 0.05` and `set_input_delay -min 0.05`
4. Any combination of `set_output_delay -min $X` and `set_input_delay -min [expr 0.1-$X]`

What difference does it make? Well one of them will insert hold buffers before the output of `element/out`, the other will insert buffers after the input of `element/in`. It's methodologically very similar, but sometimes it can make a difference.

oharboe:
@rovinski Thanks! If this PR can be approved as-is, I will ask a colleague to follow up with a refinement to see if we can achieve the same by modifying the .sdc file for the element.

maliberty (approved):

maliberty:
I see that you have no rows in the core and therefore bad clock skew:
image

You should adjust your macro halo

maliberty:
image

with
```
+++ b/flow/designs/asap7/mock-array/config.mk
@@ -12,6 +12,9 @@ export PLATFORM               = asap7
 
 export PLACE_DENSITY          = 0.30
 
+export MACRO_HALO_X            = 1
+export MACRO_HALO_Y            = 1
```

oharboe:
@maliberty  Thanks! Implemented tip here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1221

PR #1224, opened 2023-07-11
Mock array remove set dont touch

oharboe:
@rovinski FYI, I no longer need `set_dont_touch` and I don't know what change I did that made the problem go away or why, exactly, it was a problem in the first place.

Perhaps fixing the clock skew removed the need for the hold cells?


maliberty:
> @rovinski FYI, I no longer need `set_dont_touch` and I don't know what change I did that made the problem go away or why, exactly, it was a problem in the first place.
> 
> Perhaps fixing the clock skew removed the need for the hold cells?

If there is no hold problem to fix then we won't try to modify those nets so don't touch becomes a no-op.  If you have  hold problems in the future the issue could come back but as Austin pointed out you'd have to fix those at the Element level.

maliberty (approved):

maliberty:
It looks like this subsumes #1223.

oharboe:
> It looks like this subsumes #1223.

@maliberty Yes. There are a sequence of dependent reviews. I wanted to keep the build log to ensure all steps worked and in case it had any value in reviewing.

If you merge this PR, I will close the others.

PR #1227, opened 2023-07-12
makefile: remove mysterious no-op rule

oharboe:
let us see if anything bad happens if we remove it.

4_cts.sdc is written by cts.tcl, which is run by the rule a few lines above this rule


oharboe:
@maliberty Remove? I think this might have been an attempt to explain make where 4_cts.sdc comes from, a side effect of another rule. However, this rule is incorrectly articulated anyway, the rule that generated 4_cts.odb, doesn't generate 4_cts.sdc as a side-effect.

maliberty:
Some version of *.sdc: exists at every major flow step.  I presume it is in case you specifically want to generate the sdc.  I think it would be better to fix this target than remove it.  The other similar targets can be a guide.

oharboe:
> Some version of *.sdc: exists at every major flow step. I presume it is in case you specifically want to generate the sdc. I think it would be better to fix this target than remove it. The other similar targets can be a guide.

I think it is trying to express that .sdc files are built as side effects.

It.is broken, you can try to delete som of these targets and make does not know what to do.

This is broken, but does not currently affect me and I dont have any information about the intent, so I.am.going to pass this one by for now.

maliberty:
Ok there is a solution at https://www.gnu.org/software/automake/manual/html_node/Multiple-Outputs.html to express this

PR #1230, opened 2023-07-13
use setup/hold violations from json

gadfort:
While looking through the metrics I noticed that the setup and hold violation counts in the metrics do not match the ones reported by STA.
This change moves the regexed answer which is binary and replaces it with the correct counts from sta.

There will likely need to be an update to all the metrics to account for this change.

### DIFF
--- a/flow/util/genMetrics.py
+++ b/flow/util/genMetrics.py
@@ -315,16 +315,6 @@ def extract_metrics(cwd, platform, design, flow_variant, output, hier_json):
     # =========================================================================
 
     merge_jsons(logPath, metrics_dict, "6_*.json")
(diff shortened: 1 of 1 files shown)


maliberty (changes requested):
I only see the removal of the metrics not the moving or fixing...

gadfort:
@maliberty those two tags are already captured in `merge_jsons` above, this removed lines were actually overwriting the correct answers with a different answer.
GCD forexample, has 23 failing setup paths, but gets reported as 1

maliberty:
@vvbandeira please make any needed metrics updates.  I'll leave this for you to approve & merge.

vvbandeira:
Replaced by #1239, which includes metrics updates for public and private PDKs to make CI happy.

PR #1233, opened 2023-07-15
Test recover power

openroadie:
PR with tests for repair_timing -recover_power added to the flow after global_routing (uses incremental global route for parasitics).


openroadie:
Following failures show up when running 

repair_timing
followed by
repair_timing -recover_power

```
aes_sky130hs: Final timing is off by 0.1ns ... post CTS timing is fine (and that is where recover power was done).
gcd_sky130hs: Timing basically the same (floating point diff of 0.001). 
microwatt_sky130hd: CTS results are fine. Final timing off by 2%
riscv32i_sky130hd: CTS results fine. Final cap limit has minor difference
riscv32i_sky130hs: CTS results fine. Global route has minor diff. Final timing fine.
sha3_asap7: One slew limit failure.

```
Based on these results I think the power recovery feature is ready to merge without adding the command as a default into our flow.

openroadie:
Corresponding OR PR is here:

https://github.com/The-OpenROAD-Project/OpenROAD/pull/3792

openroadie:
@maliberty this is ready (and passing CI/CD).

flow/designs/asap7/jpeg_lvt/rules-base.json
```diff
@@ -52,7 +52,7 @@
         "compare": "<="
     },
     "detailedroute__route__drc_errors": {
-        "value": 0,
+        "value": 1,
```
maliberty, line 55: Do you understand the cause of this drv?

maliberty (changes requested):
New designs need to be added to Jenkins

flow/designs/asap7/jpeg_lvt/rules-base.json
openroadie: No I did not. But the last time we had a situation like this, I was told that this sort of small instability is expected and Osama is working on some new rip up and reroute logic which might mitigate this.

openroadie:
> New designs need to be added to Jenkins

How?

cc: @vvbandeira

maliberty:
See jenkins/* in ORFS

vvbandeira:
@openroadie 
Please add a new entry in the following. If the runtime of the design is greater than 2h30min, please only add to nightly for now.

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/jenkins/public_nightly.Jenkinsfile#L53
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/jenkins/public_tests_all.Jenkinsfile#L31
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/jenkins/public_tests_small.Jenkinsfile#L31

openroadie:
@maliberty @vvbandeira I cannot add the new tests to jenkins since the head OR does not have the code I need to enable adding these tests. I can handle this two ways:

- Comment out power recovery and then update the metrics later
- Wait for ORFS master to have the latest code (which has power recover). Merge master into this PR and then add the tests.

I would prefer the second approach unless there are any objections.

maliberty:
2 is fine with me

vvbandeira:
FYI update to latest master started here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1343

openroadie:
Thanks @vvbandeira and @maliberty for the help. I have added the 3 new tests to all jenkins files (they are pretty quick).

vvbandeira (approved):

PR #1234, opened 2023-07-15
makefile: yosys 0.13 or newer required

oharboe:
use yosys that ships with OpenROAD if yosys is too old.

It is highly likely than e.g. Ubuntu has a too old Yosys installed, so this avoids a common pitfall & confusion when relying on ORFS to find the tools in the path, i.e. when not using ". env.sh"

For the normal use-case it is no longer necessary to run ". env.sh" with this change.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -268,30 +268,14 @@ endif
 # The following determine the executable location for each tool used by this flow.
 # Priority is given to
 #       1 user explicit set with variable in Makefile or command line, for instance setting OPENROAD_EXE
-#       2 user include PATH to the executable prior to running make
-#       3 ORFS compiled tools: openroad, yosys
-export OPENROAD_EXE            ?= $(shell command -v openroad)
-ifeq ($(OPENROAD_EXE),)
-  export OPENROAD_EXE           = $(abspath $(FLOW_HOME)/../tools/install/OpenROAD/bin/openroad)
-endif
+#       2 ORFS compiled tools: openroad, yosys
+export OPENROAD_EXE      ?= $(abspath $(FLOW_HOME)/../tools/install/OpenROAD/bin/openroad)
+
 OPENROAD_ARGS            = -no_init $(OR_ARGS)
 OPENROAD_CMD             = $(OPENROAD_EXE) -exit $(OPENROAD_ARGS)
 OPENROAD_NO_EXIT_CMD     = $(OPENROAD_EXE) $(OPENROAD_ARGS)
 OPENROAD_GUI_CMD         = $(OPENROAD_EXE) -gui $(OR_ARGS)
 
-YOSYS_OLDEST_VERSION	= 0.13
-YOSYS_CMD               ?= $(shell command -v yosys)
-ifneq ($(YOSYS_CMD),)
-  YOSYS_VER               := $(shell $(YOSYS_CMD) -V | sed -n 's/^Yosys \([0-9]*\.[0-9]*\).*/\1/p')
-  # Semantic Versioning Comparison
-  VERSION_GTEQ            := $(shell echo -e "$(YOSYS_VER)\n$(YOSYS_OLDEST_VERSION)" | sort -V | head -n1)
-  ifeq ($(VERSION_GTEQ),$(YOSYS_OLDEST_VERSION))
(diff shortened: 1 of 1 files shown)


flow/Makefile
```diff
@@ -279,8 +279,15 @@ OPENROAD_NO_EXIT_CMD     = $(OPENROAD_EXE) $(OPENROAD_ARGS)
...
 YOSYS_CMD               ?= $(shell command -v yosys)
-ifeq ($(YOSYS_CMD),)
+YOSYS_VER               := $(shell $(YOSYS_CMD) -V | sed -n 's/^Yosys \([0-9]*\.[0-9]*\).*/\1/p')
+
+# Semantic Versioning Comparison
+VERSION_GTEQ            := $(shell echo -e "$(YOSYS_VER)\n0.13" | sort -V | head -n1)
+ifeq ($(VERSION_GTEQ),0.13)
+  $(info Using Yosys $(YOSYS_VER) found in path, it is newer or equal to 0.13)
+else
   YOSYS_CMD              = $(abspath $(FLOW_HOME)/../tools/install/yosys/bin/yosys)
```
maliberty: Why not just use the OR version unconditionally?

oharboe: I was trying to retain the intent of the code any address only a specific concern.

same for openroad? ignore what is in the path and only if explicitly overriden use something else?

I like it. Simple. Less that can go wrong.

Will it work with the .deb installer?

oharboe:
@maliberty Good to go. Error in CI is on mac, unrelated. I think this is the best we can do while supporting the following use-cases:

- debian installer of orfs + openroad
- multiple copies of ORFS repository locally while building OpenROAD tools once. Using `. env.sh` to set up paths. multiple local repositories is useful when I have multiple things I'm working on at the same time. Each repository has it's own results/ logs/ reports/ folder, etc.
- not using too old yosys when not setting up paths and getting confusing results

maliberty:
I'm not sure all the checking is needed.  If someone doesn't use env.sh there are unlikely to have set YOSYS_CMD by accident.  If they did intentionally I don't see why we shouldn't just use it.  We don't do any similar checking for KLAYOUT_CMD.

oharboe:
> I'm not sure all the checking is needed. If someone doesn't use env.sh there are unlikely to have set YOSYS_CMD by accident. If they did intentionally I don't see why we shouldn't just use it. We don't do any similar checking for KLAYOUT_CMD.

Yosys 0.9 ships with ubuntu and is used for many other things besides ORFS, hence, ORFS makefile out of the box picks out the wrong yosys.

I could implement a similar check for klayout, I've struggled with similar mysterious crashes there due to out of date klayout.

oharboe:
lets talk about it in the call tomorrow, faster than back and forth here, where we are talking past eachother on fine points that matter...

maliberty (approved):

vvbandeira:
@luarss could you update the docs to reflect these changes? I think at least the "build with prebuilt" section would need to include:

```
export OPENROAD_EXE=$(command -v openroad)
```

gadfort:
@maliberty @oharboe this change makes running ORFS much more inconvenient if you don't build from the ORFS path or have OpenROAD and yosys in your PATH already. If version control is the concern then there are better ways to address it over removing previous functionality

maliberty:
@gadfort would you elaborate on your use model?  The usual ORFS flow is build from the submodules.

gadfort:
@maliberty this change now requires me to setup an OPENROAD_EXE and YOSYS_EXE (though oddly not a KLAYOUT_EXE), if I want to use this or attempt to generate a testcase with this. That used to not be a requirement. My current solution is to just hack the Makefile and stash those changes when I pull which is not ideal.

maliberty:
ORFS doesn't have klayout as a submodule so it has to be taken from the path.

For OR & yosys they should be taken from ORFS unless the user specifically overrides.  Are you not using the submodule versions?  I'm not quite getting your use model.

gadfort:
I don't use those and haven't for a couple of years, I am simply pointing out that before it would pick what was in the path, while now I need to setup special environmental variables (just extra effort that is all).
I noticed this when trying to reproduce the tapcell/endcap thing for gf180 and didn't get a runnable testcase, but instead was told what to run, but then couldn't run yosys or OR. It's not a huge impact to me, since I'm not running this often, just thought I'd mention it here

PR #1244, opened 2023-07-18
makefile: remove superfluous floorplan_info target, instead use print-DIE_AREA

oharboe:
no need for floorplan_info target, use the more general mechanism to print any variable instead.


maliberty (approved):
I never noticed we have a print- target - I learned something :)

oharboe:
> I never noticed we have a print- target - I learned something :)

Spelunking! :-)

PR #1277, opened 2023-07-25
mock-array: register io_lsbs_* outputs in preparation for 500ps

oharboe:
@louiic FYI, registering io_lsbs_*


maliberty (approved):

oharboe:
@maliberty  This PR only deals with the concern of registering the io_lsbs_* per my discussion with @louiic Today.

There is a lot of routing on top of the macros... I think @louiic might have some concerns about this much routing on top of the macros as well as the clock?

That's a problem for later though...

image

oharboe:
@maliberty if the computer says yes, this is good to go.

oharboe:
Had to run "make update_ok"...

All of that routing on top of the macros is due to a LOT of hold cells... Doing something about those hold cells is for another PR.


image

oharboe:
Had a quick peek at the clock-tree, nothing of note, but thought I'd paste it here if anyone wants to see. The fanout has increased, because there are now 8x8 elements + 8 rows * 8 io_lsbs_*

image

PR #1281, opened 2023-07-26
mock-array: adjust clock source latency and reduce to 500ps clock period

oharboe:
@louiic The idea here is to subsume clock network latency into clock insertion latency, and thereby being able to reduce clock period.

The mock-array is imagined to be part of a bigger system where there is at least 300ps clock delay, so subsuming the clock network latency of mock-array into the clock insertion latency is OK.

I viewed this youtube to learn about `set_clock_latency` https://www.youtube.com/watch?v=kPJCLMGegmw

image

Here the clock source latency for mock-array is 300ps(so time starts at -300ps):

image


oharboe:
500ps clock period:

image

oharboe:
image

oharboe:
mock-array is not the top level...

image

oharboe:
@maliberty FYI, some shenanigans from me for Louii to review...

flow/designs/asap7/mock-array/Element/constraints.sdc
```diff
@@ -4,10 +4,11 @@ set cols [expr {[info exists ::env(MOCK_ARRAY_COLS)] ? $::env(MOCK_ARRAY_COLS) :
 
 set clk_name clock
 set clk_port_name clock
-set clk_period 1000
+set clk_period 500
 
 set clk_port [get_ports $clk_port_name]
 create_clock -period $clk_period -waveform [list 0 [expr $clk_period / 2]] -name $clk_name $clk_port
+set_clock_latency -source -100 [get_clocks $clk_name]
```
maliberty, line 11: Is it intentional that you are adding more latency at the top level than at the block level?
@louiic does this make sense to you?

oharboe: I was trying to reduce latency at top level to subsume the 100ps network latency within the Element.

oharboe: Basically I am trying to have a globally balanced clock tree across mock-array and Element.

maliberty:
are there any ff at the level of mock-array?

oharboe:
> are there any ff at the level of mock-array?

Yes, introduced yesterday.

oharboe:
If I make the following changes to flatten the design on top of master:

<pre>$ git diff
<b>diff --git a/flow/designs/asap7/mock-array/config.mk b/flow/designs/asap7/mock-array/config.mk</b>
<b>index cf124242..71338821 100644</b>
<b>--- a/flow/designs/asap7/mock-array/config.mk</b>
<b>+++ b/flow/designs/asap7/mock-array/config.mk</b>
<font color="#2AA1B3">@@ -3,7 +3,7 @@</font> include designs/asap7/mock-array/defaults.mk
 export DESIGN_NAME            = MockArray
 export DESIGN_NICKNAME        = mock-array
 
<font color="#C01C28">-export VERILOG_FILES_BLACKBOX = designs/src/mock-array/Element.v</font>
<font color="#26A269">+# export VERILOG_FILES_BLACKBOX = designs/src/mock-array/Element.v</font>
 export VERILOG_FILES          = designs/src/mock-array/*.v
 
 export SDC_FILE               = designs/asap7/mock-array/constraints.sdc
<font color="#2AA1B3">@@ -24,9 +24,9 @@</font> export DIE_AREA  = $(shell \
   cd $(dir $(DESIGN_CONFIG)) &amp;&amp; \
   python3 -c &quot;import config; print(f&apos;{0} {0} {config.die_width} {config.die_height}&apos;)&quot;)
 
<font color="#C01C28">-BLOCKS                       = Element</font>
<font color="#26A269">+# BLOCKS                       = Element</font>
 
<font color="#C01C28">-export GDS_ALLOW_EMPTY       = Element</font>
<font color="#26A269">+# export GDS_ALLOW_EMPTY       = Element</font>
 
 export MACRO_PLACEMENT_TCL   = ./designs/asap7/mock-array/macro-placement.tcl
 
</pre>

I get the following clock tree for mock-array. Here we can see that the clock network delay is 300ps. So my thinking is that with a approperiately skewed clock tree at mock-array to account for the 100ps clock network delay in Elements, the *global*(across mock-array and Element macro) clock tree should have a network delay in the range of the flattened design, i.e. 300ps.

The "favourably skewed" clock tree for mock-array would need to have a -100ps skew where it inserts the Elements.

image

oharboe:
image

maliberty:
This all seems to be based on an optimisitic view of what CTS does.  The current OR cts tries to build a balanced H-tree of minimum delay.  It is not trying to hit some specific number.

oharboe:
More reading. What I'm hoping for doesn't seem novel... https://www.eng.biu.ac.il/temanad/files/2017/02/Lecture-8-CTS.pdf

image

maliberty:
It's not novel but also non-existent in OR.  Note that the command used in not set_clock_latency which is about modeling an existing delay not creating a constraint.

oharboe:
> It's not novel but also non-existent in OR.

Quite.

> Note that the command used in not set_clock_latency which is about modeling an existing delay not creating a constraint.

After reading more, I've started to understand that...

The next question is if set_clock_latency models the non-existant feature, and if so, how well?

maliberty:
> The next question is if set_clock_latency models the non-existant feature, and if so, how well?

I don't think it models it.

rovinski:
One thing that it could help out with however is the constraints on I/O. One of the irksome parts of IO constraints is that they are in reference to an ideal clock, whereas the signal is coming from the clock tree. So if the clock tree latency increases, it makes the I/O constraints harder to hit. Adding some skew based on the clock tree latency (whether by using a negative `set_clock_latency` or otherwise) will make the timing constraints more realistic assuming the clock tree on the outside of this block is balanced as well.

This is also a problem in commercial tools which seems to unfortunately be a mostly manual process. I think the reason being that there is no proper way for the tool to infer intent from the user.

The way I have solved this issue in the past is by simply running the block through the flow once, measuring the average clock tree insertion latency (the average amount of time it takes for the clock to reach all endpoints) and then use that number to skew the I/O constraints.

rovinski:
One problem that I believe the current CTS does not address at all is if you have multiple blocks with internal clock trees of different lengths. In commercial tools, one can specify the clock tree insertion latency per block so that CTS can build a more balanced clock tree when accounting for the internal clock trees.

This has been raised before, but it's a very low priority item because it's a rather narrow use case that hasn't come up for anyone.

oharboe:
> One thing that it could help out with however is the constraints on I/O. One of the irksome parts of IO constraints is that they are in reference to an ideal clock, whereas the signal is coming from the clock tree. So if the clock tree latency increases, it makes the I/O constraints harder to hit. Adding some skew based on the clock tree latency (whether by using a negative `set_clock_latency` or otherwise) will make the timing constraints more realistic assuming the clock tree on the outside of this block is balanced as well.
> 
> This is also a problem in commercial tools which seems to unfortunately be a mostly manual process. I think the reason being that there is no proper way for the tool to infer intent from the user.
> 
> The way I have solved this issue in the past is by simply running the block through the flow once, measuring the average clock tree insertion latency (the average amount of time it takes for the clock to reach all endpoints) and then use that number to skew the I/O constraints.

So if the clock network latency is 100ps, you add 100ps to input and subtract 100ps from output latency?

With that change, I can change the clock period from 1000ps to 500ps without getting a timing violation. Physically what I think this means is that the clock would have to arrive earlier than the signals, which should happen with a balanced tree.

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1291

The timing constraints of Element (the submacro) isn't terribly important, except to properly motivate optimization, the timing is checked at the mock-array level, as I understand it.

oharboe:
> One problem that I believe the current CTS does not address at all is if you have multiple blocks with internal clock trees of different lengths. In commercial tools, one can specify the clock tree insertion latency per block so that CTS can build a more balanced clock tree when accounting for the internal clock trees.
> 
> This has been raised before, but it's a very low priority item because it's a rather narrow use case that hasn't come up for anyone.

If I understand correctly:

The OpenROAD CTS algorithm will construct a clock tree where the clock network latency of the macros is *added* to the clock network latency of the containing macro.

The CTS algorithm could instead have *subtracted* the clock network latency of the macros from the containing macros clock tree before balancing the clock tree.

This is a dramatic improvement when macros have a large clock network latency.

oharboe:
I was pondering what would happen in a design as the clock tree network latency increases as one goes up in the macro hierarchy: at some point, the piper has to be paid.

The solution that comes to mind is to create two clock trees from the same clock pin when the clock network latency becomes intolerable, thus introducing a second clock domain and using a FIFO to synchronize between the two clock domains.

This then "resets" the clock network latency.

The new clock tree that defines the clock insertion point for the inputs/outputs should be shallow, so drive few flip flops/FIFOs.

I have created a branch where I dabbled with this and created "FIFOs at home" (using only two flip flops to connect the two clock domains) when I discovered, obvious in retrospect, that the the number of flip flops driven needs to be low to get a short clock network latency: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/compare/master...Pinata-Consulting:OpenROAD-flow-scripts:mock-array-fifo-at-home

rovinski:
> If I understand correctly:
> 
> The OpenROAD CTS algorithm will construct a clock tree where the clock network latency of the macros is _added_ to the clock network latency of the containing macro.
> 
> The CTS algorithm could instead have _subtracted_ the clock network latency of the macros from the containing macros clock tree before balancing the clock tree.

Not sure if I follow, but the CTS algorithm doesn't look inside macros/blocks at all. It only cares about balancing delay down to the macro's clock pin.
This works fine when macros have the same internal delays/latencies. It doesn't work well when they don't.

oharboe:
> > If I understand correctly:
> > The OpenROAD CTS algorithm will construct a clock tree where the clock network latency of the macros is _added_ to the clock network latency of the containing macro.
> > The CTS algorithm could instead have _subtracted_ the clock network latency of the macros from the containing macros clock tree before balancing the clock tree.
> 
> Not sure if I follow, but the CTS algorithm doesn't look inside macros/blocks at all. It only cares about balancing delay down to the macro's clock pin. This works fine when macros have the same internal delays/latencies. It doesn't work well when they don't.

Yep... #3759

PR #1283, opened 2023-07-27
makefile: sweep do-copy/step warnings under the rug

oharboe:
would show the following warnings, but only on CentOS 7, not on Ubuntu 22.04.

```
$ make test-unset-and-make-print-PATH
[deleted]
--: line 0: unset: `do-copy': not a valid identifier
--: line 0: unset: `do-step': not a valid identifier
[deleted]
```


oharboe:
@maliberty I think this is the wrong fix, #1284 is better.

PR #1302, opened 2023-08-01
util: metadata depends on finish

vvbandeira:
Change target and dependencies for metadata. This change makes sure that the flow has finished before generating the metadata and metrics check. Calling `make metadata` will always update the generated json file.


vvbandeira:
@maliberty does this address your concern from https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1282#issuecomment-1655959889?

maliberty (approved):

PR #1303, opened 2023-08-01
Cleanup make issue generation

vvbandeira:
Do not save `*_CMD, *_EXE, *_SHELL`.
Do not save things that do not exist or that have a path outside `git` root.
Do not save variables that are machine specific, e.g., `SHELL, NPROC`


vijayank88:
@vvbandeira 
For example if we are submitting issue for `make cts_issue` is that necessary to add all previous stages `odb` file part of the package?
Only `3_*.odb` is sufficient right to run cts stage. Because test case size increase more, recently I seen on of the user uploaded `2GB` file size as test case.

@maliberty Please share your thoughts as well.

vvbandeira:
@vijayank88 
Just to be clear, this PR only affects the variables saved into `vars*` files. It has nothing to do with what files get included in the `tar` file. Please start a discussion for this.

maliberty:
I have mixed feelings about including a single .odb.  It would save size with is a plus.  However it often happens that someone says they have a problem in step X but really the issue is earlier in the flow (eg macro placement causing drt problems).  Having all the prior .odb can be useful.  Perhaps we should have an option for large test cases.

PR #1313, opened 2023-08-03
mock-array: simplify Element constraints, check it at mock-array level

oharboe:


oharboe:
First get #1311 into master, then revisit simplification and probably introduce a virtual clock in Element/constraint.sdc, like at the mock-array level.

PR #1314, opened 2023-08-03
Mock array max lsb flight

oharboe:
This will be needed as we move to less than 500ps clock period


maliberty:
"This branch has conflicts that must be resolved"

oharboe:
> 

Fixed, force pushed single commit that excludes accidental OpenROAD upgrade so that it is easier to review.

oharboe:
@maliberty unrelated failures

maliberty (approved):

PR #1320, opened 2023-08-03
gf180: change tapcell config

vvbandeira:
As suggested in https://github.com/The-OpenROAD-Project/OpenROAD/issues/3760#issuecomment-1664493301


maliberty:
Please hold off on this as it doesn't work correctly as I've noted in the PR

maliberty (approved):

maliberty:
It seems we don't want top & bottom rows of tapcells so this is ok.  We can revisit if needed later.

PR #1324, opened 2023-08-04
asap7: remove uart-blocks

oharboe:
it does not add much testing value as there are other designs testing BLOCKS.

I added this design back when I was first experimenting with BLOCKS and asap7 and there were no BLOCKS examples in asap7.


oharboe:
@maliberty unrelated failures in CI

maliberty:
@vijayank88 you added uart - any thoughts on this?

vijayank88:
@maliberty 
`uart` different from `uart-blocks`. 
`uart-blocks` added by @oharboe only.
Now there are many test cases with `BLOCKS` options. So, if this test case not required mean, approve this.

oharboe:
> @maliberty `uart` different from `uart-blocks`. `uart-blocks` added by @oharboe only. Now there are many test cases with `BLOCKS` options. So, if this test case not required mean, approve this.

Can we remove all uses of uart-blocks now? I only removed the one in asap7, I was unsure if the uart-blocks in gf180 had any value.

maliberty:
@vijayank88 you did add it for gf180.  Any thoughts there?

maliberty (approved):

PR #1327, opened 2023-08-04
makefile: .lyt files no longer use absolute paths for lef-files

oharboe:
instead the path is relative to the  folder

this makes it possible to mix and match docker generated .lyt files and local build files

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -350,11 +350,11 @@ $(OBJECTS_DIR)/klayout_tech.lef: $(TECH_LEF)
 	@mkdir -p $(OBJECTS_DIR)
 	sed '/OR_DEFAULT/d' $< > $@
 
- $(OBJECTS_DIR)/klayout.lyt: $(KLAYOUT_TECH_FILE) $(OBJECTS_DIR)/klayout_tech.lef
-	sed 's,<lef-files>.*</lef-files>,$(foreach file, $(OBJECTS_DIR)/klayout_tech.lef $(SC_LEF) $(ADDITIONAL_LEFS),<lef-files>$(abspath $(file))</lef-files>),g' $< > $@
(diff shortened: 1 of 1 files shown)


stefanottili:
See https://github.com/KLayout/klayout/commit/f518d1aa1d974897cfeb1ef057057125eeceef40

oharboe:
> See [KLayout/klayout@f518d1a](https://github.com/KLayout/klayout/commit/f518d1aa1d974897cfeb1ef057057125eeceef40)

Relative paths seem to work, is there a reason to use absolute paths?

How would I use this feature? I'm not familiar with klayout details... I'm just trying to fix the absolute path issue.

stefanottili:
Matthias put this in place to allow lef files to be relative depending on the technology.
I'm not familiar with the way lyt files work with regard to technologies, but I envision it similar to the lyt file having a relative lyp file path in it. If in doubt, ask the Author.

maliberty (approved):

rovinski:
If KLayout has really added support for relative paths then I don't think there is a reason to have an lyt object file at all anymore. That was the only reason it existed, as far as I can remember

oharboe:
> If KLayout has really added support for relative paths then I don't think there is a reason to have an lyt object file at all anymore. That was the only reason it existed, as far as I can remember

Good to know!

Any hints on how to remove the .lyt file? I have two more PRs pending to mess around more with the .lyt files, so getting rid of the .lyt file would remove the need for those PRs.

See #1364 and #1363

rovinski:
Clarification: this does not get rid of the .lyt file, it gets rid of the .lyt _object_ file - the one that gets created in `$(OBJECT_DIR)` via makefile rule. You might be talking about that already, but I just wanted to clarify.

I am about 3 weeks behind on my notifications due to vacation and conferences so I am slowly making my way through them :)

All you "should" need to do is replace the absolute paths in the .lyt file with the appropriate relative paths, delete the makefile rules for `$(OBJECT_DIR)/*.lyt`, and then replace references of the lyt object with the original lyt.

PR #1330, opened 2023-08-07
Mock array two element types

oharboe:
@maliberty Heads up... this will test that io pin placement is idempotent. It will fail until #1322 + a small tweak is in master due to hold cells clogging up routing.


maliberty:
Would this be better as a ppl unit test for idempotency?

oharboe:
> Would this be better as a ppl unit test for idempotency?

Hmmm... this is an integration idempotency test... 

Maybe both?

maliberty:
What's the difference?  It is a single flow step and not an interaction between steps.

oharboe:
> What's the difference? It is a single flow step and not an interaction between steps.

The input verilog is different in Element and Element_1, but the pins are the same. I expect the same result in both cases.

PR #1331, opened 2023-08-07
Refactor generate-vars.sh, fix paths in variables

vvbandeira:
~better naming
+added excluded patterns
-removed file path modifications

### DIFF
--- a/flow/util/generate-vars.sh
+++ b/flow/util/generate-vars.sh
@@ -3,47 +3,55 @@ set -euo pipefail
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 # exclude system and CI variables
-EXCLUDED_VARS="MAKE|PYTHONPATH|PKG_CONFIG_PATH|PERL5LIB|PCP_DIR|PATH|MANPATH"
-EXCLUDED_VARS+="|LD_LIBRARY_PATH|INFOPATH|HOME|PWD|MAIL|QT_QPA_PLATFORM"
-EXCLUDED_VARS+="|OPENROAD_CMD|OPENROAD_GUI_CMD|OPENROAD_NO_EXIT_CMD"
-EXCLUDED_VARS+="|LSORACLE_CMD|YOSYS_CMD|TIME_CMD|STDBUF_CMD"
-EXCLUDED_VARS+="|SHELL|OPENROAD_EXE|YOSYS_EXE"
-EXCLUDED_VARS+="|NPROC|NUM_CORES|PUBLIC|CURDIR|ISSUE_SCRIPTS|MAKEFLAGS"
+EXCLUDED_VARS="MAKE|PERL5LIB"
+EXCLUDED_VARS+="|HOME|PWD|MAIL|QT_QPA_PLATFORM"
+EXCLUDED_VARS+="|SHELL|OPENROAD_EXE|YOSYS_EXE|RESULTS_ODB"
+EXCLUDED_VARS+="|NPROC|NUM_CORES|PUBLIC|ISSUE_SCRIPTS|MAKEFLAGS"
 EXCLUDED_VARS+="|UNSET_VARIABLES_NAMES|do-step|get_variables|do-copy"
 
-# get the root directory of the Git repository
-GIT_ROOT=$(git rev-parse --show-toplevel)
-FLOW_ROOT=${GIT_ROOT}/flow
-printf '%s\n' "$ISSUE_VARIABLES" | while read -r V;
-do
-    if [[ ${V%=*} =~ ^[[:digit:]] || ${V} != *"="* || -z ${V#*=} || ${V%=*} == *"MAKEFILE"* || ${V%=*} =~ ^(${EXCLUDED_VARS})$  ]] ; then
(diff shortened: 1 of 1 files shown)


oharboe:
We are running using ORFS, but our source code is in our private repository *outside* of ORFS.

To reproduce the problems:

1. create a folder somewhere *outside* of OpenROAD-flow-scripts, e.g. /tmp/foo
2. unzip [test.zip](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/files/12287433/test.zip)
3. Change folder to /tmp/foo
4. Modify ./orfs to point to your ORFS location
5. Run `./orfs make floorplan` from within /tmp/foo
6. Run `./orfs make floorplan_issue`
7. untar the resuting .tar.gz to some folder
8. cd to the .tar.gz unzipped folder and execute `./run*.sh`

For now, this procedure fails at step 6:

```
/home/oyvind/OpenROAD-flow-scripts//flow//util/makeIssue.sh floorplan
Creating run-me-mock-array-asap7-base.sh script
Creating vars-mock-array-asap7-base.sh/tcl script
fatal: not a git repository (or any of the parent directories): .git
make: *** [/home/oyvind/OpenROAD-flow-scripts//flow//util/utils.mk:97: floorplan_issue] Error 128
oyvind@ogre:~/test$ ./orfs make floorplan_issue
```

oharboe (changes requested):
1. Fix attached test case
2. Also, empty environment variables must be saved, because ORFS checks for the *presence* of environment variables, so an empty environment variable is *not* the same as an absent environment variable.

oharboe:
@vvbandeira Can we revert 2d9767526c8727e6cc9f49096006be794e2b107a while we wait for this to be sorted out?

vvbandeira:
@oharboe
Apologies for the delay. I have changed this PR to revert the changes and included some refactoring I was working.
I will work later on a separate PR to achieve what I need regarding the paths.

oharboe (changes requested):
This PR still breaks this test-case: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1331#issuecomment-1668911314

I think the best thing is to be a bit patient and let you work on the testcase above.

Thanks!

vvbandeira:
@oharboe seems to work for me, but on step 8 I used `./orfs ./run-me-mock-array-asap7-base.sh`
Edit: considering the version from [4f70508](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1331/commits/4f7050824474a83a8a5188e68411b2e2b4f7a53f)

oharboe:
Perhaps time to rewrite this bash script to Python? :-)

```
import os
import re

# get directory of the current script
DIR = os.path.dirname(os.path.abspath(__file__))
FLOW_ROOT = os.path.realpath(os.getenv("FLOW_HOME"))
ORFS_ROOT = os.path.realpath(os.path.join(os.getenv("FLOW_HOME"), '..'))

# exclude system and CI variables
EXCLUDED_VARS = [
    "MAKE", "MAKEFLAGS", "PERL5LIB", "QT_QPA_PLATFORM", "RESULTS_ODB",
    "PUBLIC", "ISSUE_SCRIPTS", "HOME", "PWD", "MAIL", "SHELL", "NPROC",
    "NUM_CORES", "UNSET_VARIABLES_NAMES", "do-step", "get_variables", "do-copy"
]

EXCLUDED_PATTERNS = ["_EXE$", "PATH$", "_CMD$", "\."]

ISSUE_VARIABLES = os.getenv("ISSUE_VARIABLES", '').split("\n")

with open(f"{os.sys.argv[1]}.sh", 'w') as sh_file, \
        open(f"{os.sys.argv[1]}.tcl", 'w') as tcl_file, \
        open(f"{os.sys.argv[1]}.gdb", 'w') as gdb_file:
    for VAR in ISSUE_VARIABLES:
        if '=' not in VAR:
            continue
        name, value = VAR.split('=', 1)
        
        # Various filters
        if re.match("^[0-9]", name):
            continue
        if value == "''":
            value = ''
        if "MAKEFILE" in name:
            continue
        if name in EXCLUDED_VARS:
            continue
        if any(re.search(pattern, name) for pattern in EXCLUDED_PATTERNS):
            continue
        if value.startswith('/'):
            value = value.replace(FLOW_ROOT, "${FLOW_HOME}")
            value = value.replace(ORFS_ROOT, "${FLOW_HOME}/..")

        # Output
        if name in ["GND_NETS_VOLTAGES", "PWR_NETS_VOLTAGES"]:
            sh_file.write(f"export {name}='{value}'\n")
        else:
            sh_file.write(f"export {name}=\"{value}\"\n")
        tcl_file.write(f"set env({name}) \"{value}\"\n")
        gdb_file.write(f"set env {name} {value}\n")
```

oharboe (approved):
I tested 4f7050824474a83a8a5188e68411b2e2b4f7a53f, works!

For my part, this is PR is good to go. Perhaps we can make some tweaks in the future, but at least this will generate valid issue files.

Perhaps in the future an option to archive files referring to FLOW_HOME and make them relative to the unarchived folder?

vvbandeira:
@oharboe 

I agree with converting to Python (this is a first step to try and have `util` as a single Python module) and adding more functionalities to this, but I decided to merge this PR to fix some issues.

PR #1332, opened 2023-08-08
env.sh: set up OPENROAD_EXE and YOSYS_CMD

oharboe:
this allows having multiple ORFS folders locally and build once, for instance

@gadfort Useful?


gadfort:
@oharboe this just sets them to the same value they would get then the Makefile runs. I don't think this is needed. The solution is to create my own settings.mk file, all I was point out was this used to not be required.

PR #1334, opened 2023-08-08
detail route: reporting policy

oharboe:
If DETAILED_ROUTE_ARGS is not specified, save out progress report a few iterations after the first two iterations. The first couple of iterations would produce very large .drc reports without interesting information for the user.

The idea is to have a policy that gives progress information soon without having to go spelunking in Tcl or modify configuration scripts, while not having to wait too long or generating large useless reports.

This PR in addition to https://github.com/The-OpenROAD-Project/OpenROAD/pull/3826, seems like a good default policy to me.


maliberty (approved):
You should updated mock-array and sram-64x16 as they use DETAILED_ROUTE_ARGS

oharboe:
> You should updated mock-array and sram-64x16 as they use DETAILED_ROUTE_ARGS

I dont think so, the "user" has chosen those args... good to have a test.that is not the default behavior as well...

PR #1347, opened 2023-08-11
asap7: remove TNS_END_PERCENT=100, when default 5 works fine

oharboe:
A test build to see what happens.

### DIFF
--- a/flow/designs/asap7/ibex/config.mk
+++ b/flow/designs/asap7/ibex/config.mk
@@ -14,4 +14,3 @@ export PLACE_DENSITY_LB_ADDON  = 0.20
 export ENABLE_DPO = 0
 
 export DFF_LIB_FILE           = $($(CORNER)_DFF_LIB_FILE)
-export TNS_END_PERCENT        = 100
--- a/flow/designs/asap7/jpeg/config.mk
+++ b/flow/designs/asap7/jpeg/config.mk
@@ -14,5 +14,3 @@ export CORE_MARGIN            = 2
 export PLACE_DENSITY          = 0.60
 
 export DFF_LIB_FILE           = $($(CORNER)_DFF_LIB_FILE)
-
-export TNS_END_PERCENT        = 100
--- a/flow/designs/asap7/jpeg_lvt/config.mk
+++ b/flow/designs/asap7/jpeg_lvt/config.mk
@@ -23,4 +23,3 @@ export CORE_MARGIN            = 2
 export PLACE_DENSITY          = 0.60
 
 export DFF_LIB_FILE           = $($(CORNER)_DFF_LIB_FILE)
-export TNS_END_PERCENT        = 100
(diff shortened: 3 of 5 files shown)


oharboe:
@openroadie Hi Harsh, I saw that you recently changed these files and these projects work fine without TNS_END_PERCENT=100...

OK to remove?

I discovered this while digging into some hold problems, see discussion: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/discussions/1348

openroadie:
@oharboe 100 means fix all the violating timing paths while 5 means fix only 5% of the paths (minimum would be 1). So I am not sure that removing this option would be the best call unless we are not running repair timing on these designs at all. I would suggest tightening the clock period instead. 

If TNS =100 causes more timing violations then that is something we need to look at.

oharboe:
> @oharboe 100 means fix all the violating timing paths while 5 means fix only 5% of the paths (minimum would be 1). 

Nit-picking... Actually, and I got this wrong in my post, and I edited my post to fix this. The minimum is 0, which means fix only the worst path, according to https://github.com/The-OpenROAD-Project/OpenROAD/blob/b31a3b960dfa4f0cf387e53d6218e9e8f89d3aa8/src/rsz/README.md?plain=1#L193

>So I am not sure that removing this option would be the best call unless we are not running repair timing on these designs at all. I would suggest tightening the clock period instead.

If I understand correctly: you want this to be a test-case that has negative slack and you want to observe that the negative slack is repaired fully?

> If TNS =100 causes more timing violations then that is something we need to look at.

oharboe:
@openroadie I checked

asap7 ibex, slight negative slack(a few picoseconds)
asap7 jpeg, large negative slack -397.73
asap7 jpeg_lvt, not in CI
asap7 riscv32i, slight negative slack for min and max, a few picoseconds
asap7 swerv_wrapper, not in CI

openroadie:
@oharboe For the *_lvt" tests, I did add them to thje jenkins file via a PR that has now been merged into master.

Since we are doing repair_timing post CTS and before global routing (let alone detailed routing), we cannot guarantee that all negative slacks will be fixed fully at final analysis (or even post CTS if the applied transforms are not able to reduce timing violations to 0). But wherever possible, we do want the optimization code to work on all the timing paths with negative slack (which is the goal of TNS100). The only reason the default is not 100 is because for some designs, TNS100 actually makes results worse (could be congestion/maybe we need global routing parasitics /could be something else). 

cc: @maliberty @tspyrou

oharboe:
@openroadie I see. Thanks for clarifying.

Let me see if I have understood correctly 

What a good default is depends on where you are in your project. 

For exploration setting TNS5 makes sense because I am really only interested in understandIng where the problems are and fixing the root cause in the RTL.

TNS5 also makes sense when setting up the flow and .sdc files as I am not interested in working on bogus hold congestion, just learn where the hold congestion is and its relation to the .sdc files.

If the RTL is fixed and .sdc files are set up correctly and one is only trying to eke out that last bit of performance TNS100 is approperiate.

PR #1356, opened 2023-08-15
mock-alu: fleshing out the ALU

oharboe:
- multiplier based on https://github.com/antonblanchard/vlsiffra
- add, subtract and compare optimized to share resources
- single barrel shifter for all shift operations
- unit-tests to ensure that the ALU isn't simply empty
- multiplier isn't unit-tested, not quite sure how to hook up black box tests to Chisel unit-tests.

Slightly negative slack, which doesn't matter at this point.

image

image

placement heatmap looks a bit odd, transparent at the top and right:

image

image


oharboe:
@maliberty Any thoughts on whether the placement heatmap is working as expected here? Why the blank area top and right?

image

flow/designs/src/mock-alu/src/test/resources/README.md
```diff
@@ -0,0 +1,7 @@
+ASAP7 behavioral model files
+============================
+
+The multiplier is implemented in ASAP7 using Perhaps https://github.com/antonblanchard/vlsiffra
```
oharboe, line 4: @antonblanchard FYI, the multiplier from vlsiffra worked on the first try and passes unit-tests using the ASAP7 behavioral model :-)

maliberty:
Please open the heatmap as an OR issue

flow/designs/asap7/mock-alu/config.mk
```diff
@@ -9,8 +9,8 @@ export SDC_FILE               = designs/asap7/mock-alu/constraints.sdc
 
 export PLATFORM               = asap7
 
-export PLACE_DENSITY          = 0.60
-export CORE_UTILIZATION       = 40
+export PLACE_DENSITY          = 0.75
```
maliberty, line 12: Wow!

oharboe:
> Please open the heatmap as an OR issue

What about the routing congestion map, it too has a thinner black line at the top and on the right. Same thing?

flow/designs/src/mock-alu/src/test/resources/asap7sc7p5t_AO_RVT_TT_201020.v
maliberty, line 1: Why are these in  mock-alu and not the platform?

maliberty:
> > Please open the heatmap as an OR issue
> 
> What about the routing congestion map, it too has a thinner black line at the top and on the right. Same thing?

Possibly - you can make it one issue.

flow/designs/src/mock-alu/src/test/resources/asap7sc7p5t_AO_RVT_TT_201020.v
oharboe: Good point. asap7 is gigabytes, so I should only copy some of that into ORFS?

Where can I put them? The file structure of platforms/asap7/ doesn't match the ASAP7 repository of course...

maliberty: What are they being used for in ORFS?

oharboe: I have implemented a unit-test for the MockAlu and to test the multiplier I need the behavioral logic to run everything through Verilator.

The MockAlu is a RTL that is made explicitly for ORFS to provide an interesting test-case.

Hence, I need access to the behavioral logic files of the PDK from within ORFS.

oharboe:
> > > Please open the heatmap as an OR issue
> > 
> > 
> > What about the routing congestion map, it too has a thinner black line at the top and on the right. Same thing?
> 
> Possibly - you can make it one issue.

Done:

https://github.com/The-OpenROAD-Project/OpenROAD/issues/3863
https://github.com/The-OpenROAD-Project/OpenROAD/issues/3862

flow/designs/src/mock-alu/src/test/resources/asap7sc7p5t_AO_RVT_TT_201020.v
maliberty: ORFS doesn't use Verilator or run design unit tests.  RTL verification is a separate problem.  Is there a reason this needs to be in ORFS?

oharboe: If someone else is going to try to work on the mock-alu RTL, then it is going to be pretty nasty if that is in a separate repository w.r.t. syncing versions, etc.

The mock-alu has to be in lock-step with designs/asap7/mock-alu/.

I can modify the PR not to include these files from the asap7 PDK, but rather leave instructions on how to copy them in for those that want to work on the mock-alu RTL.

maliberty: I would rather not make the repo gigabytes larger for what is likely to be a very small fraction of users.  I think instructions would be better.

oharboe:
@maliberty Ready to go. Unrelated failure.

maliberty (approved):

maliberty:
Squashed to avoid the large files being in the history

PR #1364, opened 2023-08-16
Makefile fix lyt generation outside of orfs folder

oharboe:
I just verified that 0.28.11 is required, see https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1327#issuecomment-1666364273 and https://github.com/KLayout/klayout/commit/f518d1aa1d974897cfeb1ef057057125eeceef40

0.28.11 is a bit annoying, because in Ubuntu 23.04 that I'm now using. Not by choice... The world moves on and non-OpenROAD requirements is pulling in the direction of the latest Ubuntu.

building klayout takes a long time and there are no scripts in ORFS or instructions to build and use klayout.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -361,7 +361,9 @@ $(OBJECTS_DIR)/klayout_tech.lef: $(TECH_LEF)
 	sed '/OR_DEFAULT/d' $< > $@
 
 $(OBJECTS_DIR)/klayout.lyt: $(KLAYOUT_TECH_FILE) $(OBJECTS_DIR)/klayout_tech.lef
-	sed 's,<lef-files>.*</lef-files>,$(foreach file, $(OBJECTS_DIR)/klayout_tech.lef $(SC_LEF) $(ADDITIONAL_LEFS),<lef-files>$(shell realpath --relative-to=$(RESULTS_DIR) $(file))</lef-files>),g' $< > $@
+	SC_LEF_RELATIVE_PATH="$$\(env('FLOW_HOME')\)/$(shell realpath --relative-to=$(FLOW_HOME) $(SC_LEF))"; \
+	OTHER_LEFS_RELATIVE_PATHS=$$(echo "$(foreach file, $(OBJECTS_DIR)/klayout_tech.lef $(ADDITIONAL_LEFS),<lef-files>$$(realpath --relative-to=$(RESULTS_DIR) $(file))</lef-files>)"); \
+	sed 's,<lef-files>.*</lef-files>,<lef-files>'"$$SC_LEF_RELATIVE_PATH"'</lef-files>'"$$OTHER_LEFS_RELATIVE_PATHS"',g' $< > $@
 
 $(OBJECTS_DIR)/klayout_wrap.lyt: $(KLAYOUT_TECH_FILE) $(OBJECTS_DIR)/klayout_tech.lef
 	sed 's,<lef-files>.*</lef-files>,$(foreach file, $(OBJECTS_DIR)/klayout_tech.lef $(WRAP_LEFS),<lef-files>$(shell realpath --relative-to=$(OBJECTS_DIR)/def $(file))</lef-files>),g' $< > $@
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -279,7 +279,18 @@ YOSYS_CMD               ?= $(abspath $(FLOW_HOME)/../tools/install/yosys/bin/yos
 
 KLAYOUT_CMD             ?= $(shell command -v klayout)
 
-KLAYOUT_FOUND            = $(if $(KLAYOUT_CMD),,$(error KLayout not found in PATH))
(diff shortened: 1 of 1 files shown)


vijayank88:
klayout installation checked during dependency installer script: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/etc/DependencyInstaller.sh#L48

flow/Makefile
```diff
@@ -279,7 +279,18 @@ YOSYS_CMD               ?= $(abspath $(FLOW_HOME)/../tools/install/yosys/bin/yos
 
 KLAYOUT_CMD             ?= $(shell command -v klayout)
 
-KLAYOUT_FOUND            = $(if $(KLAYOUT_CMD),,$(error KLayout not found in PATH))
+MIN_VERSION = 0.28.11
```
maliberty, line 282: @vvbandeira in order to accept this will we need to update klayout on the OR cluster (and maybe the jenkins image?)
```
klayout -v
KLayout 0.28.9
```

oharboe: @vvbandeira Ping?

flow/Makefile
```diff
@@ -361,7 +361,9 @@ $(OBJECTS_DIR)/klayout_tech.lef: $(TECH_LEF)
 	sed '/OR_DEFAULT/d' $< > $@
 
 $(OBJECTS_DIR)/klayout.lyt: $(KLAYOUT_TECH_FILE) $(OBJECTS_DIR)/klayout_tech.lef
-	sed 's,<lef-files>.*</lef-files>,$(foreach file, $(OBJECTS_DIR)/klayout_tech.lef $(SC_LEF) $(ADDITIONAL_LEFS),<lef-files>$(shell realpath --relative-to=$(RESULTS_DIR) $(file))</lef-files>),g' $< > $@
+	SC_LEF_RELATIVE_PATH="$$\(env('FLOW_HOME')\)/$(shell realpath --relative-to=$(FLOW_HOME) $(SC_LEF))"; \
```
oharboe, line 364: @rovinski You mentioned somewhere, can't find it now, so paraphrasing, that "with relative paths this preprocessing was no longer necessary".

However, this preprocessing includes other files than just the one that is in [asap7.lyt](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/1c4290036b48bd71c4eb05a9fe48951a87ba3ab1/flow/platforms/asap7/KLayout/asap7.lyt#L50)

Did you mean that we could remove `$(OBJECTS_DIR)/klayout_tech.lef` below and just leave in whatever is in `asap7.lyt` already?

Should be something like changing the two lines below to:

```
OTHER_LEFS_RELATIVE_PATHS=$$(echo "$(foreach file, $(ADDITIONAL_LEFS),<lef-files>$$(realpath --relative-to=$(RESULTS_DIR) $(file))</lef-files>)"); \
sed 's,\(<lef-files>.*</lef-files>\),\1<lef-files>'"$$SC_LEF_RELATIVE_PATH"'</lef-files>'"$$OTHER_LEFS_RELATIVE_PATHS"',g' $< > $@
```

and similarly a few lines down for `$(OBJECTS_DIR)/klayout_wrap.lyt`

This fine by me, but that could also be a refinement for a later commit.  The processing to introduce the additional lef files is still necessary so the code here won't change much.

flow/Makefile
vvbandeira: Apologies for the delay. I have been busy with other time-sensitive tasks and now I am resolving a critical issue with some of our servers. As soon as this is resolved, I will get back to you on this issue -- ETA 24-36h.

flow/Makefile
rovinski: @oharboe that's a good point. It seems that that `klayout_tech.lyt` only exists because KLayout could not properly parse some custom `LEF PROPERTY` syntax used. Specifically, marking certain vias as `OR_DEFAULT` so that GRT and DRT know which via to use, IIRC. The current solution is to simply create a copy in `$(OBJECT_DIR)` but trim out the `OR_DEFAULT` lines.

A better solution to this issue is to see if a recent version of KLayout handles this LEF syntax now, and if so, require that version. Another good solution would be to reexamine if marking vias as `OR_DEFAULT` is still required or if that is obsolete (@maliberty?)

flow/Makefile
vvbandeira: @oharboe I had done the work but forgot to restart the tests 😅
Looks like there are some issues with the path of some of the files. All the designs are failing with similar messages.

```
ERROR: RuntimeError: Unable to open file: /tmp/workspace/public_tests_all-pr_PR-1364-head/flow/results/asap7/uart/base//platforms/asap7/lef/asap7sc7p5t_28_R_1x_220121a.lef (errno=2) in Layout.read
  ./util/def2stream.py:113
```

oharboe:
Superceeded by #1440

PR #1366, opened 2023-08-16
makefile: add make preview_macro_placement

oharboe:
This allows viewing illegally laid out macros to debug manual macro placement. It also works on automatic placement.

```
make clean_floorplan preview_macro_placement
```

If I break the floorplan for mock-array, I can quickly see what is going wrong. Running the command above takes a few seconds, fast iterations:

```
$ git diff
diff --git a/flow/designs/asap7/mock-array/macro-placement.tcl b/flow/designs/asap7/mock-array/macro-placement.tcl
index aed383e5..fe26eb06 100644
--- a/flow/designs/asap7/mock-array/macro-placement.tcl
+++ b/flow/designs/asap7/mock-array/macro-placement.tcl
@@ -9,7 +9,7 @@ for {set row 0} {$row < $rows} {incr row} {
   for {set col 0} {$col < $cols} {incr col} {
     set inst [$block findInst [format "ces_%d_%d" $row $col]]
 
-    set x [expr round((($array_offset_x + (($placement_grid_x * $pitch_x) * $col)) * $units))]
+    set x [expr round((($array_offset_x + (($placement_grid_x * $pitch_x) * ($col + 1))) * $units))]
     set y [expr round((($array_offset_y + (($placement_grid_y * $pitch_y) * $row)) * $units))]
 
     # belt and suspenders check... ASAP7 macro placement must be aligned to 0.048um
```

Here we can see that the macros are positions too far to the right, outside of the diea area:

image

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -522,6 +522,8 @@ endif
 #-------------------------------------------------------------------------------
 $(eval $(call do-step,2_4_floorplan_macro,$(RESULTS_DIR)/2_3_floorplan_tdms.odb $(RESULTS_DIR)/1_synth.v $(RESULTS_DIR)/1_synth.sdc $(MACRO_PLACEMENT) $(MACRO_PLACEMENT_TCL),macro_place))
 
+$(eval $(call do-step,2_floorplan_debug_macros,$(RESULTS_DIR)/2_1_floorplan.odb $(RESULTS_DIR)/1_synth.v $(MACRO_PLACEMENT) $(MACRO_PLACEMENT_TCL),floorplan_debug_macros))
+
 # STEP 5: Tapcell and Welltie insertion
 #-------------------------------------------------------------------------------
 $(eval $(call do-step,2_5_floorplan_tapcell,$(RESULTS_DIR)/2_4_floorplan_macro.odb $(TAPCELL_TCL),tapcell))
@@ -915,6 +917,18 @@ gui_cts: gui_4_cts.odb
 gui_route: gui_5_route.odb
 .PHONY: gui_final
 gui_final: gui_6_final.odb
+
+.PHONY: preview_macro_placement
+
+ifneq ($(or $(MACRO_PLACEMENT),$(MACRO_PLACEMENT_TCL)),)
+MACRO_PREVIEW_ODB = 2_floorplan_debug_macros.odb
+else
+MACRO_PREVIEW_ODB = 2_4_floorplan_macro.odb
+endif
+
+preview_macro_placement:
+	@$(UNSET_AND_MAKE) $(RESULTS_DIR)/$(MACRO_PREVIEW_ODB)
+	@$(UNSET_AND_MAKE) gui_$(MACRO_PREVIEW_ODB)
 
 .PHONY: $(foreach file,$(RESULTS_DEF),gui_$(file))
 $(foreach file,$(RESULTS_DEF),gui_$(file)): gui_%:
--- a/flow/scripts/floorplan_debug_macros.tcl
+++ b/flow/scripts/floorplan_debug_macros.tcl
@@ -1,8 +1,8 @@
 source $::env(SCRIPTS_DIR)/load.tcl
-load_design 2_3_floorplan_tdms.odb 1_synth.sdc "Starting macro placement"
+load_design 2_1_floorplan.odb 1_synth.sdc "Debug floorplan"
 
 source $::env(SCRIPTS_DIR)/macro_place_util.tcl
 
 if {![info exists save_checkpoint] || $save_checkpoint} {
-  write_db $::env(RESULTS_DIR)/2_4_floorplan_macro.odb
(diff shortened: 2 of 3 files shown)


oharboe:
@maliberty unrelated failure

maliberty:
Since you moved all the code to a new file it is hard to see what the diff is.  Please explain how this differs from the standard target.

oharboe:
> Since you moved all the code to a new file it is hard to see what the diff is. Please explain how this differs from the standard target.

There is no change, I just needed to be able to use the exact same code from two different places.

maliberty:
Then what is the purpose of the new target?  It seems to be the same as the regular target aside from loading the wrong odb 2_1_floorplan.odb

oharboe:
> Then what is the purpose of the new target? It seems to be the same as the regular target aside from loading the wrong odb 2_1_floorplan.odb

If one is working on the floorplan, manual macro placement, before the io pin constraints, then io pin placment can fail with mysterious errors.

this separates the concern of macro placement debugging from io constraints.

maliberty:
> > Then what is the purpose of the new target? It seems to be the same as the regular target aside from loading the wrong odb 2_1_floorplan.odb
> 
> If one is working on the floorplan, manual macro placement, before the io pin constraints, then io pin placment can fail with mysterious errors.
> 
> this separates the concern of macro placement debugging from io constraints.

The initial (random) io placement doesn't depend on macro placement so how would the manual macro placement cause it to fail?  Wouldn't you just be iterating on the 2_4 step?

oharboe:
> > > Then what is the purpose of the new target? It seems to be the same as the regular target aside from loading the wrong odb 2_1_floorplan.odb
> > 
> > 
> > If one is working on the floorplan, manual macro placement, before the io pin constraints, then io pin placment can fail with mysterious errors.
> > this separates the concern of macro placement debugging from io constraints.
> 
> The initial (random) io placement doesn't depend on macro placement so how would the manual macro placement cause it to fail? Wouldn't you just be iterating on the 2_4 step?

I don't know why it fails, but it does if you have bugs in your manual macro placement or the floorplan is way off(like much too big or too small).

So the thinking is, this will enable us to come up with a floorplan and macro placement we intended, then we can run the rest of the flow.

maliberty:
It seems that you aren't concerned about the interaction between the two so much as you want to skip over IO placement.  I think it would be better to have an option to do that than a target that is tailored to a narrow use case and has caveats on its usage that aren't obvious (eg only for use with manual macro placement).

oharboe:
> It seems that you aren't concerned about the interaction between the two so much as you want to skip over IO placement. I think it would be better to have an option to do that than a target that is tailored to a narrow use case and has caveats on its usage that aren't obvious (eg only for use with manual macro placement).

Perhaps `make skip_to_manual_floorplan`?

modeled on `skip_cts` and the others.

Then afterwards the user can run `make gui_2_4_floorplan_macro.odb`

The naming could do with a bit of work...

maliberty:
> Perhaps `make skip_to_manual_floorplan`?
> 
> modeled on `skip_cts` and the others.
> 
> Then afterwards the user can run `make gui_2_4_floorplan_macro.odb`
> 
> The naming could do with a bit of work...

It that's fine but there should be a check that MACRO_PLACEMENT_TCL or MACRO_PLACEMENT is set to catch misuse.

oharboe:
> > Perhaps `make skip_to_manual_floorplan`?
> > modeled on `skip_cts` and the others.
> > Then afterwards the user can run `make gui_2_4_floorplan_macro.odb`
> > The naming could do with a bit of work...
> 
> It that's fine but there should be a check that MACRO_PLACEMENT_TCL or MACRO_PLACEMENT is set to catch misuse.

From a usability point of view, I like the current interface in this PR.

I could rename the target I have to e.g. `make gui_macro_placement` and add a check for MACRO_PLACEMENT/TCL....

maliberty:
> From a usability point of view, I like the current interface in this PR.
> 
> I could rename the target I have to e.g. `make gui_macro_placement` and add a check for MACRO_PLACEMENT/TCL....

The current gui targets don't do anything beside open a design so this would break the convention.   The name needs to include 'manual' in any case to make it use clear.

oharboe:
> > From a usability point of view, I like the current interface in this PR.
> > I could rename the target I have to e.g. `make gui_macro_placement` and add a check for MACRO_PLACEMENT/TCL....
> 
> The current gui targets don't do anything beside open a design so this would break the convention. The name needs to include 'manual' in any case to make it use clear.

preview_manual_macro_placement

?

maliberty:
> preview_manual_macro_placement
ok

oharboe:
@maliberty Added conditional code so `make preview_macro_placement` works for manual and automatic macro placement.

oharboe:
@maliberty Heads up! I force pushed to get rid of the conflict with master. The macro_place_util.tcl is now a cut and past of latest master into a separate file.

maliberty (approved):

PR #1369, opened 2023-08-16
makefile: avoid 6_report.log being written when that step fails

oharboe:
this avoids a situation where if you run `make finish` twice, and it fails the first time, it succeeds(false negative) the second time.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -475,8 +475,8 @@ $(if $(5),$(5),$(RESULTS_DIR))/$(1)$(if $(4),$(4),.odb): $(2)
 
 .PHONY: do-$(1)
 do-$(1):
-	{ $(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/$(3).tcl -metrics $(LOG_DIR)/$(1).json && cp $(LOG_DIR)/$(1).tmp.log $(LOG_DIR)/$(1).log; } 2>&1 | tee $(LOG_DIR)/$(1).tmp.log
-	rm -f $(LOG_DIR)/$(1).tmp.log
+	($(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/$(3).tcl -metrics $(LOG_DIR)/$(1).json) 2>&1 | tee $(LOG_DIR)/$(1).tmp.log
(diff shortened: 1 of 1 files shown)


flow/Makefile
```diff
@@ -475,7 +475,8 @@ $(if $(5),$(5),$(RESULTS_DIR))/$(1)$(if $(4),$(4),.odb): $(2)
 
 .PHONY: do-$(1)
 do-$(1):
-	($(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/$(3).tcl -metrics $(LOG_DIR)/$(1).json) 2>&1 | tee $(LOG_DIR)/$(1).log
+	{ $(TIME_CMD) $(OPENROAD_CMD) $(SCRIPTS_DIR)/$(3).tcl -metrics $(LOG_DIR)/$(1).json && cp $(LOG_DIR)/$(1).tmp.log $(LOG_DIR)/$(1).log; } 2>&1 | tee $(LOG_DIR)/$(1).tmp.log
+	rm -f $(LOG_DIR)/$(1).tmp.log
```
maliberty: why cp and not mv?

oharboe: The file still exists.

maliberty: How can it exist if you moved it?

oharboe: Not sure I follow, I copy and delete, I dont move it.

maliberty: copy + delete == mv ?  So why not mv?

oharboe: because the file is still open `(a&&b) | tee foo.txt`, here foo.txt is open until b has completed.

However, I'm testing what you expected now. moving an open file is fine in Linux. Will update PR  after some more testing.

maliberty: There is no need to use && - just mv it after it finishes.  If it errors the step should stop before mv is executed.

oharboe: Indeed.

oharboe: Fixed

oharboe:
@maliberty Should be good to go now, followed your suggestion on using mv, don't know why I didn't do that first.

maliberty (approved):

PR #1370, opened 2023-08-16
Use relative paths for submodules so they use the same protocol as the top one.

titan73:
My proxy only supports git over ssh so I have to clone the repo with ssh but cloning submodules fails are they are explicitly using https. This PR makes paths relative so the same protocol is used.

Same as https://github.com/The-OpenROAD-Project/OpenROAD/pull/3764 which has been merged.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
 [submodule "tools/yosys"]
 	path = tools/yosys
-	url = https://github.com/The-OpenROAD-Project/yosys.git
+	url = ../../The-OpenROAD-Project/yosys.git
 [submodule "tools/OpenROAD"]
 	path = tools/OpenROAD
 	url = ../OpenROAD.git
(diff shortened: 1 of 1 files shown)


vvbandeira (approved):

titan73:
CI pr-head and pr-merge both faith with riscv32i asap7 target with following error:
[ERROR] finish__timing__setup__ws fail test: -112.891 >= -88.1
What am I supposed to do to fix that?

maliberty:
There are currently some CI issues in the main branch.  I wouldn't worry about that as part of this PR.

titan73:
Ah ok. Btw what are exactly pr-head and pr-merge?

maliberty:
pr-head == build of the head of your branch
pr-merge == build of the head of master with your branch merged in

The former tests just your change.  The later looks for a possible interaction with the head of master which may be ahead of where you branched off from.

PR #1377, opened 2023-08-17
makefile: add OPENROAD_GUI_ARGS

oharboe:
the usecase is to run openroad in console mode.

This can be used by humans, but also for scripting.

For instance: inject a command or two and then parse the output. Convenient for quick and dirty Python scripts to plot something, for instance.

To start OpenROAD in console mode with the final result:

make OPENROAD_GUI_ARGS="" gui_final

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -273,7 +273,8 @@ export OPENROAD_EXE      ?= $(abspath $(FLOW_HOME)/../tools/install/OpenROAD/bin
 OPENROAD_ARGS            = -no_init $(OR_ARGS)
 OPENROAD_CMD             = $(OPENROAD_EXE) -exit $(OPENROAD_ARGS)
 OPENROAD_NO_EXIT_CMD     = $(OPENROAD_EXE) $(OPENROAD_ARGS)
-OPENROAD_GUI_CMD         = $(OPENROAD_EXE) -gui $(OR_ARGS)
+OPENROAD_GUI_ARGS       ?= -gui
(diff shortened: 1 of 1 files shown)


maliberty:
Is there a use case beyond just "" to get a non-gui console?  If not I would just make a parallel set of targets to gui_* like open_* as that would be clearer

oharboe:
> Is there a use case beyond just "" to get a non-gui console? If not I would just make a parallel set of targets to gui_* like open_* as that would be clearer

Not that I can think of now.

Maybe con_* is a better match for the gui_* naming convention?

maliberty:
> > Is there a use case beyond just "" to get a non-gui console? If not I would just make a parallel set of targets to gui_* like open_* as that would be clearer
> 
> Not that I can think of now.
> 
> Maybe con_* is a better match for the gui_* naming convention?

People may not know what con means so I would use the full word console_*.  I prefer open but not strongly.

oharboe:
> > > Is there a use case beyond just "" to get a non-gui console? If not I would just make a parallel set of targets to gui_* like open_* as that would be clearer
> > 
> > 
> > Not that I can think of now.
> > Maybe con_* is a better match for the gui_* naming convention?
> 
> People may not know what con means so I would use the full word console_*. I prefer open but not strongly.

Open works. However, I just thought of something. There is a corresponding set of gui_ for .def files. I take it those are obscure enough that we don't need open_* set for them too?

maliberty:
> Open works. However, I just thought of something. There is a corresponding set of gui_ for .def files. I take it those are obscure enough that we don't need open_* set for them too?

I think we can ignore them.  If a need arises its easy to add.

oharboe:
#1385 replaces this one.

PR #1386, opened 2023-08-18
Make issue fixes for gdb tcl

vvbandeira:
Fixes setting variables in tcl and gdb.

Allows use of:

```
GDB=1 ./run-me*.sh
```
@maliberty is this helpful for `gdb` or do you have another workflow in mind?


maliberty:
Usually I just 'gdb openroad' and then source the var*gdb file.  I could try it this way instead in which case we wouldn't need the .gdb file.

PR #1388, opened 2023-08-21
Recover power

openroadie:
Recover power flow improved 
- All paths are now optimized (the percentage setting may be kept for runtime reasons)
- Bad moves are undone


openroadie:
I have 2 detailed route errors. Am going to update the metrics since this is a known random instability.

cc: @maliberty @osamahammad21

vvbandeira (approved):

PR #1397, opened 2023-08-27
mock-array: IO Placement with annealing works fine

oharboe:


flow/platforms/asap7/config.mk
```diff
@@ -267,3 +267,5 @@ export IR_DROP_LAYER ?= M1
 
 # Allow empty GDS cell
 export GDS_ALLOW_EMPTY ?= fakeram.*
+
+export PLACE_PINS_ARGS = -annealing
```
eder-matheus: Do you want every asap7 design running with annealing? We have some metrics failures in other asap7 designs that are probably caused by running them with annealing.

Perhaps we should change only the mock-array design?

eder-matheus (changes requested):

flow/platforms/asap7/config.mk
oharboe: ah, this is a gaffe. i only intended to do this w mock array.

eder-matheus (approved):

maliberty (approved):

PR #1398, opened 2023-08-27
Megaboom cached netlist

oharboe:
Make it easier to play around with this design by making the source code easier to browse and not having to wait for synthesis.


flow/designs/src/megaboom/rocketchip.MegaBoomConfig.v
maliberty, line 1: Why does this need to be uncompressed?  Even an old editor like emacs can edit a gzip'ed file.  This eats disk space for minor convenience.

oharboe: works with all editors not only old ones.

i never thought to consider disk space... the temp files in ORFS dwarf this.many times over, pdks are gigabytes...

oharboe:
@maliberty There are so many other issues with megaboom at this point that I'm just shelving this PR for now until after discussions.

PR #1427, opened 2023-08-31
[Docs] Docker shell update

luarss:
@vvbandeira Addresses our concern of having "docker" targets as discussed.

@oharboe Let me know your thoughts on this change. I think the last version intended for running the GUI from within the docker image, but that assumes that the docker image has docker installed, which seems a bit weird to me.


flow/util/docker_shell
```diff
@@ -26,12 +29,15 @@ docker run -u $(id -u ${USER}):$(id -g ${USER}) \
...
+ -e OPENROAD_EXE=$OPENROAD_EXE \
+ -e KLAYOUT_CMD=$KLAYOUT_CMD \
+ -v $WORKSPACE:/OpenROAD-flow-scripts/flow \
  $DOCKER_INTERACTIVE \
  ${OR_IMAGE:-openroad/flow-centos7-builder:latest} \
  bash -c "set -ex
  mkdir /tmp/xdg-run
- . ./env.sh
- cd $WORKSPACE
+ #. ./env.sh
```
oharboe: why commented out?

oharboe: otherwise looks good to me. I'm not using this, I just shared the docker_shell script....

luarss: If you mount the docker image from the `flow` folder it seems that docker will not see any files above that directory. Hence `env.sh` will not appear in docker.

oharboe: Ok. Delete line then?

I no longer have a stake in docker_shell in ORFS, so I am good with whatever happens going forward. I am happy to answer specific  technical questions, but I dont plan on following closely.

I am glad docker_shell is finding use :-)

docs/user/DockerShell.md
```diff
@@ -1,4 +1,63 @@
...
+Also, the current working directory is mapped into the Docker image using the current user's credentials.
+
+Build docker image
+------------------
+
+First build the docker image:
+
+```
+cd OpenROAD-flow-scripts
+sudo ./setup.sh
```
vvbandeira:
```suggestion
```

flow/util/docker_shell
```diff
@@ -26,12 +29,14 @@ docker run -u $(id -u ${USER}):$(id -g ${USER}) \
  -e XAUTHORITY=$XAUTH \
  -e FLOW_HOME=/OpenROAD-flow-scripts/flow/ \
  -e MAKEFILES=/OpenROAD-flow-scripts/flow/Makefile \
- -v $WORKSPACE:`pwd` \
+ -e YOSYS_CMD=$YOSYS_CMD \
+ -e OPENROAD_EXE=$OPENROAD_EXE \
+ -e KLAYOUT_CMD=$KLAYOUT_CMD \
```
vvbandeira, line 35: @luarss, could you remind me why we need to pass these variables to docker?

docs/user/DockerShell.md
vvbandeira: For docker builds, the only requirement should be docker, which is not installed by `setup.sh`. Right?

vvbandeira (changes requested):

docs/user/DockerShell.md
luarss: Yes, you are correct. Should we add a disclaimer asking users to install Docker first?

vvbandeira: For building, we should point to `BuildWithDocker.md` as the single source for building with docker. In the `DockerShell.md`, there should be no/minimal mention of building; it should focus on running the flow using the helper script.

flow/util/docker_shell
luarss: This command sets the env variables if the user specifies it. Otherwise it sets to the default path it would have been inside the docker.

docs/user/DockerShell.md
luarss: Should we remove the first section on "First, build the docker image"?

oharboe:
LGTM

vvbandeira (approved):

PR #1433, opened 2023-09-03
makefile: remove generate_abstract dependencies

oharboe:
As long as the abstracts exist, then to regenerate an abstract is an explicit choice made by
the user, not a dependency to be checked.

The full set of dependencies for an abstract can only be checked by rebuilding the macro.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -322,11 +322,11 @@ ifeq ($(wildcard $(3)),)
 # more global concerns instead of getting mired in the details of
 # each macro.
 block := $(patsubst ./designs/$(PLATFORM)/$(DESIGN_NICKNAME)/%,%,$(dir $(3)))
-$(1) $(2) &: ./designs/$$(PLATFORM)/$$(DESIGN_NICKNAME)/block.mk
-	$$(UNSET_AND_MAKE) DESIGN_NAME=${block} DESIGN_NICKNAME=$$(DESIGN_NICKNAME)_${block} DESIGN_CONFIG=$$< generate_abstract
+$(1) $(2) &:
+	$$(UNSET_AND_MAKE) DESIGN_NAME=${block} DESIGN_NICKNAME=$$(DESIGN_NICKNAME)_${block} DESIGN_CONFIG=./designs/$$(PLATFORM)/$$(DESIGN_NICKNAME)/block.mk generate_abstract
 else
 # There is a unqiue config.mk for this Verilog module
-$(1) $(2) &: $(3)
+$(1) $(2) &:
 	$$(UNSET_AND_MAKE) DESIGN_CONFIG=$$< generate_abstract
 endif
 endef
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -327,7 +327,7 @@ $(1) $(2) &:
 else
 # There is a unqiue config.mk for this Verilog module
 $(1) $(2) &:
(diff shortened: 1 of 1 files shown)


maliberty:
> As long as the abstracts exist, then to regenerate an abstract is an explicit choice made by
the user, not a dependency to be checked.

I think that should be a non-default option.  The premise of make is that it brings everything up to date.  This would be a dangerous default.

oharboe:
> > As long as the abstracts exist, then to regenerate an abstract is an explicit choice made by
> > the user, not a dependency to be checked.
> 
> I think that should be a non-default option. The premise of make is that it brings everything up to date. This would be a dangerous default.

Why this dependency? There are LOTS of missed dependencies. What makes this one special?

The only way to check the dependencies is to invoke generate_abstract on ALL macros in BLOCKS.

I dont think more options is the way to go here. I dont see any justification(in comments) nor can I imagine one for treating one out of all the dependencies for each macro as special.

oharboe:
@maliberty Fixed DCO and a gaffe, no related failures seen in CI before I fixed the DCO.

oharboe:
@maliberty Think about it this way: would we add exactly these dependencies today?

If no, then that's pretty strong evidence they should be removed.

maliberty:
The abstract generated when you run the top level today so there must be a dependency already.  The abstract itself should depend on the files required to build it.

oharboe:
> The abstract generated when you run the top level today so there must be a dependency already.

Actually there is no such requirement. There can be a rule that only requires the files to exist without having any dependencies for a rule.

> The abstract itself should depend on the files required to build it.

This is practically impossible, unless you *always* invoke generate_abstract. Any files that go into the abstract would have to be listed from macro configuration files to verilog files to sdc constraint files, etc. Further, make checks dependencies on startup, so if you have a rule that change files as a side effect without listing those files, then make gets confused. 

What the code does today is weird: it adds effectively one out of all the dependencies for the abstract. In fact, it is a dependency that doesn't even exist when running generate_abstract directly, so it's a dummy dependency.

maliberty:
The abstract should depend on 6_final.odb, 6_final.sdc, and CDL_FILES in a similar style to all the other steps.  6_final.odb should drag in all the prior dependencies.

Side note - it looks like generate_abstract.tcl is missing a call to read_spef.  We are calling write_timing_model so the final parasitics should be read.  The spef file should also be a dependency.  Also missing is "set_propagated_clock [all_clocks]".   I'll look at fixing these two.

oharboe:
> The abstract should depend on 6_final.odb, 6_final.sdc, and CDL_FILES in a similar style to all the other steps. 6_final.odb should drag in all the prior dependencies.

Except here this is not what is going on because the depenendencies you refer to is in a make session that has not been invoked yet.

These depedencies exist only in the submake.



> 
> Side note - it looks like generate_abstract.tcl is missing a call to read_spef. We are calling write_timing_model so the final parasitics should be read. The spef file should also be a dependency. Also missing is "set_propagated_clock [all_clocks]". I'll look at fixing these two.

👍

maliberty:
"Recursive make considered harmful" strikes again.

maliberty (approved):

maliberty:
I think the flow needs some work in this area but it doesn't have to gate this PR.

oharboe:
> "Recursive make considered harmful" strikes again.

I think it is a feature here :-) When I fix the top level I cam freqeuently use out of date macros in the process.

PR #1435, opened 2023-09-04
asap7/sram-64x16: remove SYNTH_HIERARCHICAL and update_ok

oharboe:
demonstrates that SYNTH_HIERARCHICAL is not required for RTLMP_FLOW=True


maliberty:
This design has no RTL hierarchy so its not surprising it doesn't matter in this case.

oharboe:
> This design has no RTL hierarchy so its not surprising it doesn't matter in this case.

Is SYNTH_HIERARCHICAL  required for mpl2?

If so, why a separate option?

maliberty (approved):

maliberty:
It is recommended but in a case like this with no hierarchy it doesn't do anything and so can be skipped.

oharboe:
> It is recommended but in a case like this with no hierarchy it doesn't do anything and so can be skipped.

Why is it recommended with mpl2? Is this documented anywhere?

maliberty:
mpl2 is named "RTL-MP" as it uses the RTL hierarchy as a guide for macro clustering.  If you don't use SYNTH_HIERARCHICAL then the RTL is flattened by yosys removing this guidance.  It will still run but the QOR may degrade.  You can get more details in the papers https://github.com/The-OpenROAD-Project/OpenROAD/tree/master/src/mpl2#references

PR #1463, opened 2023-09-14
[DO NOT MERGE] build: link abc into yosys to improve startup error messages

oharboe:


oharboe:
It passed all the CI tests, which is what I wanted to test at this point.

PR #1479, opened 2023-09-19
makefile: add skip_place and reduce repetition in comments

oharboe:


oharboe:
@maliberty unrelated mac failure

maliberty (commented):
When would you use skip_place?

oharboe:
> When would you use skip_place?

To create eviscerated macros in bug reports and with preview_macro_placement.

maliberty:
> > When would you use skip_place?
> 
> To create eviscerated macros in bug reports and with preview_macro_placement.

Won't all the subsequent steps fail if the design is unplaced?

oharboe:
> > > When would you use skip_place?
> > 
> > 
> > To create eviscerated macros in bug reports and with preview_macro_placement.
> 
> Won't all the subsequent steps fail if the design is unplaced?

Not if you skip cts and route too 😅

maliberty (approved):

PR #1493, opened 2023-09-25
s/commercial/proprietary/

mithro:
Both `sky130` and `gf180mcu` are commercial platforms. 

The actually difference is that `GF12LP` and `TSMC65LP` are proprietary / non-open platforms.


vvbandeira (approved):

vvbandeira:
@dralabeing FYI

PR #1495, opened 2023-09-25
Change `/bin/bash` to `/usr/bin/env bash`.

mithro:
While most scripts are already using `/usr/bin/env bash` to find bash, there are still a few left which reference `/bin/bash`.

This pull request updates;
 * `util/makeIssue.sh`
 * `util/generate-vars.sh`

### DIFF
--- a/flow/util/generate-vars.sh
+++ b/flow/util/generate-vars.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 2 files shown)


mithro:
I was only looking at stuff under `flow` directory.

Looks like there are still plenty of references directly to `/bin/bash` in other top level directories;
```
dev_env.sh:#!/bin/bash                                                                        
etc/DependencyInstaller.sh:#!/bin/bash                                                        
etc/DockerHelper.sh:#!/bin/bash                                                               
flow/util/docker_shell:#!/bin/bash                                                            
setup.sh:#!/bin/bash  
```

vvbandeira (approved):

PR #1496, opened 2023-09-25
Disable make's implicit rules.

mithro:
By default `make` has a large number of "default implicit" rules, see https://www.gnu.org/software/make/manual/html_node/Implicit-Rules.html and https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html

These significantly slow down make runs and increase the noise in output (particularly with `-d` / `--debug=all` on).

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -115,6 +115,10 @@ endif
 #
 # ==============================================================================
 
+# Disable make's implicit rules
+MAKEFLAGS += --no-builtin-rules
+.SUFFIXES:
+
 #-------------------------------------------------------------------------------
 # Default target when invoking without specific target.
 .DEFAULT_GOAL := finish


mithro:
This seems to be failing with the following;
```
ld: warning: directory not found for option '-F/Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/System/Library/Frameworks'
ERROR: Something is wrong with flag 'flagfile' in file 'Users/corentinl/work/or-tools/build_make/_deps/absl-src/absl/flags/parse.cc'. One possibility: file 'Users/corentinl/work/or-tools/build_make/_deps/absl-src/absl/flags/parse.cc' is being linked both statically and dynamically into this executable. e.g. some files listed as srcs to a test and also listed as srcs of some shared lib deps of the same test.
CMake Error at /usr/local/Cellar/cmake/3.27.5/share/cmake/Modules/GoogleTestAddTests.cmake:112 (message):
  Error running test executable.

    Path: '/Users/runner/work/OpenROAD-flow-scripts/OpenROAD-flow-scripts/tools/OpenROAD/build/src/mpl2/test/cpp/mpl2_test'
    Result: 1
    Output:
      

  /usr/local/Cellar/cmake/3.27.5/share/cmake/Modules/GoogleTestAddTests.cmake:225 (gtest_discover_tests_impl)


make[2]: *** [src/mpl2/test/cpp/mpl2_test] Error 1
make[2]: *** Deleting file `src/mpl2/test/cpp/mpl2_test'
make[1]: *** [src/mpl2/test/cpp/CMakeFiles/mpl2_test.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

mithro:
It looks like the `testInstaller-macos` is also failing at HEAD - https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/actions/runs/6286271733/job/17069534052

flow/Makefile
```diff
@@ -115,6 +115,10 @@ endif
 #
 # ==============================================================================
 
+# Disable make's implicit rules
+MAKEFLAGS += --no-builtin-rules
```
oharboe, line 119: Sweet!

vvbandeira (approved):

PR #1521, opened 2023-10-03
Make eqy optional

vvbandeira:


maliberty:
ready?

maliberty:
@vvbandeira is there some replacement plan?

PR #1522, opened 2023-10-03
env: update, PATH no longer needs to be set

vvbandeira:


maliberty:
@vvbandeira is this ready?

PR #1542, opened 2023-10-10
env: only add openroad to the PATH

vvbandeira:


maliberty:
@vvbandeira is this waiting for something?

PR #1543, opened 2023-10-11
makefile: add do-gds

oharboe:
separate the concerns of checking dependencies and making the gds file.

Example use-case: make a wafer thin Bazel wrapper on top of ORFS to get artifact support for macros that can be in the hundreds of megabytes range.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -376,8 +376,12 @@ $(OBJECTS_DIR)/lib/merged.lib:
 # Pre-process KLayout tech
 # ==============================================================================
 $(OBJECTS_DIR)/klayout_tech.lef: $(TECH_LEF)
+	$(UNSET_AND_MAKE) do-klayout_tech
+
+.PHONY: do-klayout_tech
+do-klayout_tech:
 	@mkdir -p $(OBJECTS_DIR)
-	cp $< $@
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty Unrelated failures. Good to go for my part.

flow/Makefile
```diff
@@ -718,6 +732,7 @@ $(eval $(call do-copy,5_route,4_cts.sdc,,.sdc))
 
 .PHONY: do-route
 do-route:
+	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)
```
maliberty, line 735: How could you get to these later steps without even generating these directories?  The input files are coming from $(RESULTS_DIR) so how could it not exist?

oharboe: A bit overeager w copy and paste. I can remove the results dir.

reports and logs are trickier. If I use Bazel, then I only have files in the results folder when e.g. I get the synthesis artifact.

maliberty: ok

oharboe: Do you want me to fix it? I will probably, over the next month, find some more nits to fix...

maliberty: I suppose it doesn't hurt and can wait if you wish.

maliberty (approved):

PR #1550, opened 2023-10-12
util/bazel: example of wafer thin Bazel layer on top of ORFS

oharboe:
Sharing an example of how to use Bazel with ORFS. This is an example usage of the `do-foo` rules in the ORFS Makefile.

Also, it should provide some useful context for future feature requests/bug-reports.

### DIFF
--- a/flow/util/bazel/.gitignore
+++ b/flow/util/bazel/.gitignore
@@ -1,0 +1,7 @@
+aes
+bazel-*
+build/
+vars-*
+run-*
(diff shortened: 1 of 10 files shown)


maliberty:
I'm not sure it makes sense to merge this.  The OR team is not supporting Bazel and wouldn't test, support problems, or update this flow.

oharboe:
table for now, possibly to be resurrected later.

PR #1555, opened 2023-10-12
enable repair_antennas after global route

eder-matheus:


maliberty:
Sorry, I had to put through a revert so you'll have to resolve the conflicts in flow/Makefile.

eder-matheus:
@maliberty I've fixed the conflict, but I see an overlap error in the secure-ci. I'll check it and let you know when I have it fixed.

eder-matheus:
@maliberty CI is green after the updates in recover_power. Only ca53 and bp_single are failing, but it is unrelated to the changes in this PR.

flow/scripts/global_route.tcl
```diff
@@ -27,6 +27,18 @@ global_route -guide_file $env(RESULTS_DIR)/route.guide \
...
+    -right $::env(CELL_PAD_IN_SITES_DETAIL_PLACEMENT)
+repair_antennas -iterations 5
+check_placement -verbose
+
+# Reset padding to zero, as it interfers with recover power and generate
+# overlaps
+set_placement_padding -global \
+    -left 0 \
+    -right 0
+
```
eder-matheus: @maliberty I had to reset the placement padding after running repair_antennas because it interferes with recover_power, resulting in overlaps. Do you have an idea of why it may be happening?

maliberty: No - that seems odd. recover_power doesn't do any placement afaik so I don't see why it would matter. What sort of overlaps?

maliberty:
@vvbandeira does the renaming of steps have an impact on metrics or CI in general?

flow/scripts/global_route.tcl
eder-matheus: This is just one of the overlaps I see when not resetting the padding to zero. If I remove the recove_power call, the overlaps go away.

image

Oddly, the master of the two cells does not change after recover_power, but the origin of the left cell changes; it is shifted to the right, generating the overlap. If I understood correctly, recover_power should not change instance positions, only the master.

maliberty: It sounds like a bug as it is not supposed to move anything.  Please look into it.

vvbandeira:
@eder-matheus, please check for conflicts and adapt:
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/util/genMetrics.py

CI metrics should not be affected.

flow/scripts/global_route.tcl
eder-matheus: I've found that `replaceCell` function calls opendp when using groute parasitics:
image

I believe `legalCellPos` tries to legalize the cell if it is out of the die, but since the cell is already in a valid place, the position is not changing. However, it adds the `padLeft` to this position, resulting in the overlaps.
image

I will add a check to avoid running the remaining function when the legalized position is equal to the new position, but we should revisit this function in the future.

eder-matheus:
> @eder-matheus, please check for conflicts and adapt: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/util/genMetrics.py
> 
> CI metrics should not be affected.

@vvbandeira genMetrics.py updated.

eder-matheus:
@maliberty CI is green now. It should be safe to merge this PR.

maliberty (approved):

PR #1557, opened 2023-10-13
makefile: add do-gds

oharboe:
separate the concerns of checking dependencies and making the gds file.

Example use-case: make a wafer thin Bazel wrapper on top of ORFS to get artifact support for macros that can be in the hundreds of megabytes range.

There is a problem with testing do-wrapped-gdsoas as only gf12 uses WRAP_LEFS and gf12 is under NDA, but the code is written with the assumption that `$(WRAPPED_GDSOAS)` contains zero or more files.

W.r.t. the do-foo variants, only `$(WRAPPED_GDSOAS)` needs to be updated.

`$(WRAPPED_LEFS)` and `$(WRAPPED_LIBS)` also generates multiple targets, but this happens during synthesis. There is no "do-synth" as we can scrape by without a "do-synth" as it is the *first* target to be run and when making a  wafer thin Bazel wrapper, the Makefile target dependency checks do not cause a problem for the first target because we always want to run the full synth target within Bazel and generate all the dependencies of synth.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -376,8 +376,12 @@ $(OBJECTS_DIR)/lib/merged.lib:
 # Pre-process KLayout tech
 # ==============================================================================
 $(OBJECTS_DIR)/klayout_tech.lef: $(TECH_LEF)
+	$(UNSET_AND_MAKE) do-klayout_tech
+
+.PHONY: do-klayout_tech
+do-klayout_tech:
 	@mkdir -p $(OBJECTS_DIR)
-	cp $< $@
+	cp $(TECH_LEF) $(OBJECTS_DIR)/klayout_tech.lef
 
 KLAYOUT_ENV_VAR_IN_PATH_VERSION = 0.28.11
 KLAYOUT_VERSION = $(shell $(KLAYOUT_CMD) -v 2>/dev/null | grep 'KLayout' | cut -d ' ' -f2)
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty @tspyrou Obviously I'm going to need some help with testing the gf12 target...

flow/Makefile
```diff
@@ -844,35 +855,51 @@ clean_abstract:
...
-	        -rd config_file=$(FILL_CONFIG) \
-	        -rd seal_file="" \
-	        -rd out_file=$@ \
-	        -rd tech_file=$(OBJECTS_DIR)/klayout_wrap.lyt \
-	        -rd layer_map=$(GDS_LAYER_MAP) \
-	        -r $(UTILS_DIR)/def2stream.py) 2>&1 | tee $(LOG_DIR)/6_merge_$(basename $(notdir $@)).log
+	for target in $(WRAPPED_GDSOAS); do \
+		($(TIME_CMD) $(KLAYOUT_CMD) -zz \
+		-rd design_name=$(basename $(notdir $$target)) \
+		-rd in_def=$(OBJECTS_DIR)/def/$(notdir $${target:$(STREAM_SYSTEM_EXT)=def}) \
```
oharboe: I tried to test this w/`make DESIGN_CONFIG=designs/gf12/bp_dual/config.mk do-wrapped-gdsoas` and modifying designs/gf12/bp_dual/config.mk to use asap7. Obviously that testing was quite limited as it failed, but I was able to fix some basic syntax errors.

@maliberty Could one of the non-NDA PDKs be modified to use WRAPPED_GDSOAS just so as to make testing easier?

oharboe: The implementation was made more complicated by the fact that centos 7 is using make prior to 4.3. In 4.3, "&::" is supported, which would have made the implementation cleaner.

oharboe: @maliberty I'm removing support for WRAP_LEFS.

oharboe:
I have removed support for WRAP_LEFS for do-gds. WRAP_LEFS should be retired #1561

flow/Makefile
maliberty: Why?

oharboe:
@maliberty Simply don't touch WRAP_LEFS for the do-gds case, so gf12 now should work for normal builds. 

This should be good to go as long as CI doesn't turn up anything.

This code is fine as is when, at some point, the need for is obviated WRAP_LEFS/LIBS https://github.com/The-OpenROAD-Project/OpenROAD/issues/4131

flow/Makefile
oharboe: I'm removing support for WRAP_LEFS for the do-gds case. For normal builds it is still there unchanged.

If someone wants do-gds for gf12, they either have to write support WRAP_LEFS or pay back the technical debt so that WRAP_LEFS is no longer needed.

oharboe:
@maliberty Unrelated CI failures

maliberty (approved):

oharboe:
@maliberty Thanks!

PR #1558, opened 2023-10-13
makefile: create required output directories for do-floorplan/place/route/finish

oharboe:
Bazel use-case: support executing these stages using only the artifacts from the previous stage and do not rely on any side effects, such as having creating log folders.


oharboe:
@maliberty Restore this commit, I think it was only reverted because it caused a conflict when needing to revert the do-wrapped-gds commit in #1556.

Also, it no longer creates the results folder, as it already exists given that each of the do- steps uses results from a previous stage.

maliberty (approved):

PR #1562, opened 2023-10-15
Makebash env

### LINKED ISSUE #1497: `make bash` should put something in the prompt
### Description

When you type `make bash` you get a bash shell with all the environment setup as the Makefile would execute things. 

However, it is very easy to forget that you are inside this environment as there is no indication. 

There is also no indication that you have successfully entered the environment.

### Suggested Solution

Things like `conda` and `virtualenv` generally put something into the bash prompt to indicate that you are inside the environment.

### Additional Context

_No response_

luarss:
@maliberty  Fixes https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1497

`source env.sh` will setup the bashrc and the necessary variable, which is set during Makefile if `make bash` is called. Also ensures this is only ever called once. 

image

Upon running `make bash` you can see the change in environment and exiting it returns to normal. 
image


flow/Makefile
```diff
@@ -1050,7 +1050,7 @@ yosys:
 # Drop into a bash shell with all environment variables, useful for debugging
 .PHONY: bash
 bash:
-	bash
+	export MAKEFILE_ENV=1; bash
```
rovinski: Why not just do:

```
bash:
  bash --init_file $(FLOW_HOME)/../env.sh
```

And avoid touching bashrc?

rovinski (commented):
Disagree with the methodology. Scripts should not modify a user's bashrc.

flow/Makefile
luarss: Neat trick, thanks. I am using the MAKEFILE_ENV variable still to identify that `make bash` is called, is that okay?

rovinski: Well #1497 is only asking to fix the prompt for `make bash`. Maybe it's just easier to do the following?

```
bash:
  bash --init_file <(echo '...')
```

This uses a pipe redirect so there is no external file needed. the `...` is the `PS1=...` string with all the correct character escapes.

luarss: Fixed, should be cleaner now.

maliberty (approved):

maliberty:
Note that if you use PROMPT_COMMAND (as I do) then this doesn't work but it is likely good enough.

PR #1594, opened 2023-11-03
makefile: add ABSTRACT_SOURCE variable for generate-abstract

oharboe:
This can be used to generate an abstract using for instance 2_floorplan stage, which allows quickly previewing floorplan changes at the top level.

When I tested it with MegaBoom, I could view the floorplan after mpl2 in minutes instead of days.

I tried out putting all pins on the left edge of the macros and changing aspect ratio according to tip from @louiic, resulting floorplan preview:

image

### DIFF
--- a/flow/scripts/generate_abstract.tcl
+++ b/flow/scripts/generate_abstract.tcl
@@ -1,8 +1,13 @@
 source $::env(SCRIPTS_DIR)/load.tcl
-load_design 6_final.odb 6_final.sdc "Starting generation of abstract views"
 
-read_spef $::env(RESULTS_DIR)/6_final.spef
-set_propagated_clock [all_clocks]
+set stem [expr {[info exists ::env(ABSTRACT_SOURCE)] ? $::env(ABSTRACT_SOURCE) : "6_final"}]
+
+load_design $stem.odb $stem.sdc "Starting generation of abstract views"
+
+if {[file exists $::env(RESULTS_DIR)/$stem.spef]} {
+  read_spef $::env(RESULTS_DIR)/$stem.spef
+  set_propagated_clock [all_clocks]
+}
 
 puts "Starting generation of abstract views"
 write_timing_model $::env(RESULTS_DIR)/$::env(DESIGN_NAME).lib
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty Unrelated docs error.

flow/scripts/generate_abstract.tcl
```diff
@@ -1,16 +1,21 @@
...
-read_spef $::env(RESULTS_DIR)/6_final.spef
-set_propagated_clock [all_clocks]
+set stem [expr {[info exists ::env(ABSTRACT_SOURCE)] ? $::env(ABSTRACT_SOURCE) : "6_final"}]
+
+load_design $stem.odb $stem.sdc "Starting generation of abstract views"
+
+if {[file exists $::env(RESULTS_DIR)/$stem.spef]} {
+  read_spef $::env(RESULTS_DIR)/$stem.spef
+  set_propagated_clock [all_clocks]
+}
```
maliberty: you should call estimate_parasitics otherwise

flow/scripts/generate_abstract.tcl
```diff
@@ -1,16 +1,21 @@
...
 
-read_spef $::env(RESULTS_DIR)/6_final.spef
-set_propagated_clock [all_clocks]
+set stem [expr {[info exists ::env(ABSTRACT_SOURCE)] ? $::env(ABSTRACT_SOURCE) : "6_final"}]
+
+load_design $stem.odb $stem.sdc "Starting generation of abstract views"
+
+if {[file exists $::env(RESULTS_DIR)/$stem.spef]} {
+  read_spef $::env(RESULTS_DIR)/$stem.spef
+  set_propagated_clock [all_clocks]
```
maliberty, line 17: propagated clocks make sense in any step cts or later

flow/scripts/generate_abstract.tcl
oharboe: Can you give a bit more detail?

Like?

```
else {
   estimate_parasitics
}
```

maliberty: see gui.tcl

oharboe:
@maliberty unrelated error in CI. Ready for review after fixing requests for change.

maliberty (approved):

maliberty:
pr-merge seems to be a git hang

PR #1596, opened 2023-11-06
do-place and do-route now produces artifacts even in the case of failure

oharboe:
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/discussions/1593


flow/scripts/detail_place.tcl
```diff
@@ -7,7 +7,15 @@ source $::env(PLATFORM_DIR)/setRC.tcl
 set_placement_padding -global \
     -left $::env(CELL_PAD_IN_SITES_DETAIL_PLACEMENT) \
     -right $::env(CELL_PAD_IN_SITES_DETAIL_PLACEMENT)
-detailed_placement
+if {[info exists env(WRITE_ON_FAIL)] && $::env(WRITE_ON_FAIL)} {
+  set result [catch {detailed_placement} errMsg]
+  if {$result != 0} {
+    write_db $::env(RESULTS_DIR)/3_5_place_dp.odb
```
maliberty: I disagree with the approach of writing the final odb file as if it is succeeded.  If you want to write 3_5_place_db-FAILED.odb that would make more sense to me.

oharboe: That does not work and defeats the purpose of this PR: Bazel must have the same artifact file name, even if placement failed. Checking for valid placement is a separate step afterwards. This separates the concern of producing an artifact to be used or inspected and checking for failure.

maliberty: If you write the success artifact how will make know that the step is actually failed?

flow/Makefile
```diff
@@ -651,10 +651,41 @@ $(eval $(call do-copy,3_place,3_5_place_dp.odb,))
...
+	fi
+	$(UNSET_AND_MAKE) do-3_4_place_resized
+	$(UNSET_AND_MAKE) WRITE_ON_FAIL=1 do-3_5_place_dp
+	if [ $$? -ne 0 ]; then \
+		cp $(RESULTS_DIR)/3_5_place_dp.odb $(RESULTS_DIR)/3_place.odb ; \
+		$(UNSET_AND_MAKE) do-3_place.sdc ; \
+		exit 1 ; \
+	fi
+	$(UNSET_AND_MAKE) do-3_place do-3_place.sdc
+	echo >$(RESULTS_DIR)/place.ok 1
```
oharboe: @maliberty This file contains success/failure

flow/scripts/detail_place.tcl
oharboe: implemented via place.ok and route.ok files.

flow/Makefile
maliberty: Make will not read the contents of the file on the next invocation and will think this step succeeded.

oharboe: Indeed:

1. do-xxxx is not compatible w normal make target checking, you have to pick one or the other. This is fine.
2. I need to add check-place to do-cts and check-route to do-finish

flow/Makefile
```diff
@@ -657,10 +659,42 @@ $(eval $(call do-copy,3_place,3_5_place_dp.odb,))
...
+# placement fails. In this case $(RESULTS_DIR)/3_place.odb and $(RESULTS_DIR)/3_place.sdc
+# are produced, such that an artifact can be created for inspection purposes.
+#
+# The check-place target subsequently will fail if
+# placement did not complete successfully, because $(RESULTS_DIR)/place.ok
+# is set to 0.
+#
+# This means that build systems, such as Bazel, that require a non-zero
+# exit code and the same artifacts to be produced every time, or it
+# will not publish the artifacts, can use this target.
```
maliberty, line 674: @QuantamHD how do your Bazel setup work?  Presumably you would have the same issue so I'm wondering if you solved it this way?

QuantamHD: I break my flow into steps (placement, cts, drt, etc.). Run them one at a time, and if a step fails I just rerun the tcl script for that stage which references outputs from the previous successful stage.

Long story short if a step fails I don't get artifacts either.

oharboe: I see.

My main use-cases for Bazel is that I want artifacts for projects that have hours of running time so that I can examine failed builds quickly. This allows me to turn around a tweak to my scripts in a short amount of time or run `make issue` so that I can easily send feature request/issues to the community(frequently using deltaDebug.py...)

This PR gives me artifacts for failed buidls for global place, detailed place, global route and detailed route.

Later, I suspect that I'll find it worth adding artifacts for failed builds to CTS.

Note! We are also using bazel_rules_hdl, but to create a Verilator artifact.

maliberty:
Please address "This branch has conflicts that must be resolved"

oharboe:
> Please address "This branch has conflicts that must be resolved"

Done.

oharboe:
@maliberty  Unrelated failures in CI. Anything missing on this one?

maliberty:
Would you explain further why a failed run must produce the same artifacts as a passing one?  I'm still not excited about this change.

oharboe:
> Would you explain further why a failed run must produce the same artifacts as a passing one? I'm still not excited about this change.

By design, Bazel considers a run that does not produce a promised file a failed step. So, what is needed to "impedance match" ORFS with Bazel is an implementation that honors this contract.

I believe this is a design requirement that this there for good reasons after long considerations and other failed alternatives.

I suspect that the reason why is that the dependency and artifact checking can then be handled 100% within Bazel, which probably improves performance and simplifies testing.

maliberty:
It is a failed step.  Is Bazel unable to record any result from a failed step?

oharboe:
> It is a failed step. Is Bazel unable to record any result from a failed step?

My understanding is no. You don't get artifacts stored on the server that you can download for a failed step.

oharboe:
> > It is a failed step. Is Bazel unable to record any result from a failed step?
> 
> My understanding is no. You don't get artifacts stored on the server that you can download for a failed step.

If you try to run a failed step, then Bazel will redo that step, which is exactly what I didn't want. It is possible that you could go spelunking on the server to find a failed artifact, but that would defeat the whole purpose of artifacts for failed global/local place/route, these artifacts are required on the local machine.  Emphemral output folders that are quickly overwritten are no good.

maliberty:
Can we split the functionality between the ORFS & Bazel layers?  I would like to see ORFS write a step-failed.odb file.  In the Bazel layer you can rename that to step.odb if that's what you require.  No side file would be required in either case.

flow/Makefile
```diff
@@ -607,6 +609,7 @@ $(RESULTS_DIR)/2_floorplan.sdc: $(RESULTS_DIR)/2_1_floorplan.odb
 .PHONY: do-floorplan
 do-floorplan:
 	mkdir -p $(LOG_DIR) $(REPORTS_DIR)
+	touch $(LOG_DIR)/2_3_floorplan_tdms.log
```
maliberty, line 612: Why?

flow/Makefile
```diff
@@ -658,10 +661,47 @@ $(eval $(call do-copy,3_place,3_5_place_dp.odb,))
...
+	touch $(LOG_DIR)/3_2_place_iop.log
+	touch $(LOG_DIR)/3_3_place_gp.log
+	touch $(LOG_DIR)/3_4_place_resized.log
+	touch $(LOG_DIR)/3_5_place_dp.log
+	echo >$(RESULTS_DIR)/place.ok 0
+	$(UNSET_AND_MAKE) do-3_1_place_gp_skip_io do-3_2_place_iop
+	@$(UNSET_VARS); \
+	$(SUB_MAKE) do-3_3_place_gp; \
+	if [ $$? -ne 0 ]; then \
+		cp $(RESULTS_DIR)/3_3_place_gp-failed.odb $(RESULTS_DIR)/3_place.odb ; \
```
maliberty, line 688: This whole thing should be moved to Bazel not ORFS

maliberty (changes requested):

oharboe:
broke out #1619 and #1613, which I think should be fine to go in and moved the rest into a private.mk file that will be in the MegaBoom project.

PR #1605, opened 2023-11-08
update grt congestion iterations

### LINKED ISSUE #1604: Reduce default global routing effort back to 30
### Description

@eder-matheus  In ccb312ade8 https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/9422c8d4f51a446f3aaaf0560b883418269573ab/flow/scripts/global_route.tcl#L27 was changed from 30 to 50.

### Suggested Solution

Instead of changing the default value to 50, update only those example designs that need more global routing effort with global route arguments to increase effort.

Otherwise, it's a slippery slope... Every time routing is too hard for some design, the default will be increased...

### Additional Context

_No response_

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1604

Running CI to check what designs need increasing the number of iterations.


maliberty (approved):
No design required this change?

PR #1612, opened 2023-11-10
makefile: move foo.tmp.log to foo.log in the case of failure

oharboe:


oharboe:
@maliberty unrelated CI failures

maliberty (approved):

PR #1637, opened 2023-11-22
generate_abstract: improve comment

oharboe:
Generating abstract views

reads better than

Starting generation of abstract views

### DIFF
--- a/flow/scripts/generate_abstract.tcl
+++ b/flow/scripts/generate_abstract.tcl
@@ -4,7 +4,9 @@ set stem [expr {[info exists ::env(ABSTRACT_SOURCE)] ? $::env(ABSTRACT_SOURCE) :
 
 set design_stage [lindex [split [file tail $stem] "_"] 0]
 
-load_design $stem.odb $stem.sdc "Starting generation of abstract views"
+set generating_comment "Generating abstract views"
+
+load_design $stem.odb $stem.sdc $generating_comment
 
 if {$design_stage >= 6 && [file exists $::env(RESULTS_DIR)/$stem.spef]} {
   read_spef $::env(RESULTS_DIR)/$stem.spef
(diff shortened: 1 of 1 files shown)


flow/scripts/generate_abstract.tcl
```diff
@@ -17,7 +19,7 @@ if {$design_stage >= 4} {
   set_propagated_clock [all_clocks]
 }
 
-puts "Starting generation of abstract views"
+puts $generating_comment
```
maliberty, line 22: It seems redundant to print it twice (since we are cleaning things up)

oharboe: I wanted a single concern for this PR, to fix the text.

I don't understand why load_design takes this parameter or under what circumstances it is printed:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/c72ffd7dcd317c337df2535314289f8d0a593467/flow/scripts/load.tcl#L2

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/c72ffd7dcd317c337df2535314289f8d0a593467/flow/scripts/load.tcl#L33

maliberty: The standalone variable is leftover from an old mode where we ran the entire flow in a single invocation of OR (see run_all.tcl).  We haven't used it for years so I doubt it still works.  The theory was that you could save the write/read overhead between steps when running the whole flow.

In practice standalone is never set today so this else won't happen.

oharboe: Ripping out the standalone stuff is a separate concern, not related to this PR.

Should it be done?

maliberty (approved):

flow/scripts/generate_abstract.tcl
maliberty: Its a nice idea so there was always the thought to resuscitate it someday.  I've approved this and we can table the rest for now.

PR #1639, opened 2023-11-23
makefile: remove standalone vestiges

oharboe:
remove a few lines of the disused standalone capability to clarify the code


flow/scripts/load.tcl
```diff
@@ -1,37 +1,33 @@
 proc load_design {design_file sdc_file msg} {
```
maliberty: You should remove the msg arg as it is no longer used (and update the callers)

oharboe:
@maliberty unrelated failures

maliberty (approved):

PR #1641, opened 2023-11-23
asap7/riscv32i: use non-default thresholds for clusters' max number of std cells in mpl2

AcKoucher:
Test-CI passed.

_For documentation:_ The changes in [#4206](https://github.com/The-OpenROAD-Project/OpenROAD/pull/4206) caused some degradation in the critical timings paths for asap7/riscvi32. This degradation happens as a result of a bad choice of macro location inside the only macro cluster generated for this design. As it is a small design, we end up having two "large" std cells cluster when running mpl2 multilevel autoclustering engine, making it hard for mpl2 make a good supposition for the location of std cells placement.

This PR is to ensure we split these two large std cells clusters and mpl2 can come up with a better location for these macros.

With the changes here, [#4206](https://github.com/The-OpenROAD-Project/OpenROAD/pull/4206) no longer causes degradation.


maliberty:
Would you run 'make update_ok' to adjust the metrics for this improvement.

maliberty (approved):

PR #1645, opened 2023-11-24
docs: user: Fix YOSYS_CMD

dnltz:
Fix command not found when trying to export the
yosys binary path and remove "shell" from the
sub-command.

$ export YOSYS_CMD=$(shell command -v yosys)
Command 'shell' not found, did you mean:
...


vvbandeira (approved):
Thanks for catching this.

PR #1650, opened 2023-11-27
Pinata fix gnd pwr syntax nit

oharboe:
The env var contains `""` (two double quotes)  instead of an empty string as intended
by these statements.

Fix these statements so that this incorrect idiom is not copied
and pasted elsewhere


maliberty:
Missed asap7/swerv_wrapper

oharboe:
> Missed asap7/swerv_wrapper

Fixed.

maliberty (approved):

PR #1652, opened 2023-11-27
cts: add CTS_ARGS env var to override CTS arguments

oharboe:


flow/scripts/cts.tcl
```diff
@@ -24,21 +24,24 @@ proc save_progress {stage} {
...
-                      -sink_clustering_max_diameter $cluster_diameter \
-                      -balance_levels
+  lappend cts_args -distance_between_buffers "$::env(CTS_BUF_DISTANCE)"
+}
+
+if {[info exist ::env(CTS_ARGS)]} {
+  set cts_args $::env(CTS_ARGS)
 }
 
+puts "clock_tree_synthesis [join $cts_args " "]"
```
maliberty, line 41: Why print the command?  We don't elsewhere.

oharboe: This is the same as we do in global route.

Copy and paste actual args from log into config.mk and adjust.

maliberty:
Note the collision course with https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1649

flow/scripts/cts.tcl
maliberty: Where in global_route.tcl do you see this?

oharboe: Seems like I was a bit quick. There is no print in global route, but otherwise it is the same concept behind the impmementation.

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/8c90610291910fe89e1394cbe3a0db7b5d4d067b/flow/scripts/global_route.tcl#L27

maliberty: The implementation is fine, its the puts that I was questioning.

oharboe: Found it. In detailed route, the arguments are output to the log, so I thought I'd do the same here for the same reason. To set DETAILED_ROUTE_ARGS, copy and paste from the log and modify in config.mk:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/8c90610291910fe89e1394cbe3a0db7b5d4d067b/flow/scripts/detail_route.tcl#L66

maliberty: Looks like you did that one as well: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/b8c7c7097fa6079c9fb3af397d3dca8ac412330a

I can live with it but I don't like the inconsistency in the flow.

maliberty (approved):

flow/scripts/cts.tcl
oharboe: agreed, which is why I wrote it this way, consistent w detailed route where the arguments can be fully overridden.

PR #1678, opened 2023-12-04
Adjusted ihp-sg13g2 stdcell power stripe width to match standard cells

Syndace:
The stdcell power stripe width of `0.48` does not match the ihp-sg13g2 standard cells. By looking in the LEF file, I found `0.44` to be the correct width. Maybe @KrzysztofHerman could confirm this?


KrzysztofHerman:
@Syndace, yes the value of 0.44 is the correct one. Thank you for contribution.

maliberty (approved):

maliberty:
Thanks that should make the routing easier!

PR #1680, opened 2023-12-04
scripts: fix mpl2 options / include the ones that were missing

AcKoucher:
I put the commands in the order we see them in the docs.

I'm opening this as a draft, because I'm changing bus_planning_flag to an actual flag in #[4339](https://github.com/The-OpenROAD-Project/OpenROAD/pull/4339) which needs to be merged first.


oharboe:
@maliberty @AcKoucher Would you consider #1683 as an alternative path forward?

AcKoucher:
@oharboe I wasn't aware of #1683. Apparently the approach there would be much cleaner. If @maliberty agrees I'll close this.

PR #1683, opened 2023-12-05
rtlmp: add RTLMP_ARGS consistent with global/place/cts

oharboe:
RTLMP has a lot of environment variables.

This change makes it possible to add and remove any current and future RTLMP argument, consistent with how this is done in global route, placement and cts.

The next step, once this is merged, would be to prune the disused RTLMP_ environment variables to reduce repetition and expose the OpenROAD interface directly through RTLMP_ARGS, rather than create an ORFS environment variable version of every current and future RTLMP argument.

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/74c112eb3ab5a3fc8c703b428aa7386bd0cb3da1/flow/scripts/global_route.tcl#L32

### DIFF
--- a/flow/scripts/macro_place_util.tcl
+++ b/flow/scripts/macro_place_util.tcl
@@ -117,11 +117,14 @@ if {[find_macros] != ""} {
         append additional_rtlmp_args " -fence_uy $env(RTLMP_FENCE_UY)"
     }
 
+    set all_args $additional_rtlmp_args
 
-    puts "Call Macro Placer $additional_rtlmp_args"
+    if { [info exists ::env(RTLMP_ARGS)] } {
+      set all_args $::env(RTLMP_ARGS)
+    }
 
-    rtl_macro_placer \
-                 {*}$additional_rtlmp_args
+    puts "rtl_macro_placer [join $all_args " "]"
+    rtl_macro_placer {*}$all_args
 
     puts "Delete buffers for RTLMP flow..."
     remove_buffers


maliberty (approved):

maliberty:
While I'm fine with this change, this

> The next step, once this is merged, would be to prune the disused RTLMP_ environment variables to reduce repetition and expose the OpenROAD interface directly through RTLMP_ARGS, rather than create an ORFS environment variable version of every current and future RTLMP argument.

concerns me.  It would make this inconsistent with all the other tools that present individual options.   Also the autotuner needs discrete knobs to turn so this would impair the ability to use it.  No user is required to use the options so I'm not clear of the harm.

oharboe:
> While I'm fine with this change, this
> 
> > The next step, once this is merged, would be to prune the disused RTLMP_ environment variables to reduce repetition and expose the OpenROAD interface directly through RTLMP_ARGS, rather than create an ORFS environment variable version of every current and future RTLMP argument.
> 
> concerns me. It would make this inconsistent with all the other tools that present individual options. Also the autotuner needs discrete knobs to turn so this would impair the ability to use it. No user is required to use the options so I'm not clear of the harm.

The problem is repetition and no unit-tests. Every option has to be repeated in ORFS and documented whereas it is already documented and should be unit-tested in OpenROAD.

It is no arduous task for Autotuner to articulate a string of OpenROAD arguments and articulate it as a command line argument.

maliberty:
Unit tests are an orthogonal concern (that I share).

What rule is there for how many arguments can have per switch vars and how many is too many?  It seems a bit arbitrary.

oharboe:
> Unit tests are an orthogonal concern (that I share).
> 
> What rule is there for how many arguments can have per switch vars and how many is too many? It seems a bit arbitrary.

Unsure what you are asking.

Generally, I think that the role of ORFS is to make common cases w.r.t. these parameters easy and the more obscure cases possible.

maliberty:
In detail_route.tcl we have a fair number of individual arguments but you haven't proposed to do the same there.  Is the only reason to do it in mpl2 is the number of arguments?

oharboe:
> In detail_route.tcl we have a fair number of individual arguments but you haven't proposed to do the same there. Is the only reason to do it in mpl2 is the number of arguments?

mpl2 is newer and the parameters are added but not used, but really the same idea applies. Make ORFS thinner.

PR #1684, opened 2023-12-05
makefile: fix escaping problem with GDS_ALLOW_EMPTY

oharboe:
Now works, used to fail:

```
$ make  GDS_ALLOW_EMPTY=\(a\|b\) print-SUB_MAKE
SUB_MAKE = make  GDS_ALLOW_EMPTY=\(a\|b\) --no-print-directory DESIGN_CONFIG=./designs/nangate45/gcd/config.mk
$ make "GDS_ALLOW_EMPTY=(a|b)" print-SUB_MAKE
SUB_MAKE = make  GDS_ALLOW_EMPTY=\(a\|b\) --no-print-directory DESIGN_CONFIG=./designs/nangate45/gcd/config.mk
```


oharboe:
@maliberty unrelated errors in dependencyinstaller tests

flow/scripts/escape.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+sed -e 's/ /\\ /g' -e 's/(/\\(/g' -e 's/|/\\|/g' -e 's/)/\\)/g'
```
maliberty, line 2: Do [].*^$\ need escaping?

oharboe: Yes, but I fought a bit with make, bash and sed w.r.t. escaping hell and lost...

I don't see these other chars used by ORFS variables.

Can we merge this? It is a step forward...

To test further escape rules, run something like:

```
$ make  "GDS_ALLOW_EMPTY=(a|b)" test-unset-and-make-print-GDS_ALLOW_EMPTY
GDS_ALLOW_EMPTY = (a|b)
```

maliberty (approved):

PR #1687, opened 2023-12-06
detailed route: remove ORFS vars that are unused

oharboe:
Detailed route allows passing any option directly to the detailed_route openroad command via the DETAILED_ROUTE_ARGS environment variable. This allows config.mk to pass any option to detailed routing and also external tools, like Autotuner.

The ORFS variables that are in detail_route.tcl now are used by the PDK to provide defaults or from config.mk project files.

This commit was written after a discussion about what ORFS should expose and serves as an example of some careful "Bonasai pruning".

This commit follows the principle that simple common usage should be simple, complicated or uncommon use-cases should be possible.

### DIFF
--- a/flow/scripts/detail_route.tcl
+++ b/flow/scripts/detail_route.tcl
@@ -10,14 +10,7 @@ set_propagated_clock [all_clocks]
 
 set_thread_count $::env(NUM_CORES)
 
-if { [info exists ::env(PRE_DETAIL_ROUTE_TCL)] } {
-  source $::env(PRE_DETAIL_ROUTE_TCL)
-}
-
 set additional_args ""
-if { [info exists ::env(dbProcessNode)]} {
-  append additional_args " -db_process_node $::env(dbProcessNode)"
-}
 if { [info exists ::env(OR_SEED)]} {
   append additional_args " -or_seed $::env(OR_SEED)"
 }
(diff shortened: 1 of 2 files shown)


rovinski:
I think that removing some of these might break the private CI.

oharboe:
> I think that removing some of these might break the private CI.

Gotit. In which case, those are examples of variables in use, ideally a test-case can be added to public CI. This would both be good for grepping(it is found to be in use) and for testing...

flow/scripts/detail_route.tcl
```diff
@@ -10,14 +10,7 @@ set_propagated_clock [all_clocks]
 
 set_thread_count $::env(NUM_CORES)
 
-if { [info exists ::env(PRE_DETAIL_ROUTE_TCL)] } {
-  source $::env(PRE_DETAIL_ROUTE_TCL)
-}
-
 set additional_args ""
-if { [info exists ::env(dbProcessNode)]} {
```
maliberty, line 18: Must keep

flow/scripts/detail_route.tcl
```diff
@@ -40,9 +33,6 @@ if { [info exists ::env(VIA_IN_PIN_MAX_LAYER)]} {
 if { [info exists ::env(DISABLE_VIA_GEN)]} {
   append additional_args " -disable_via_gen"
 }
-if { [info exists ::env(REPAIR_PDN_VIA_LAYER)]} {
```
maliberty: Must keep

maliberty (changes requested):

maliberty:
I don't see an easy way to setup a meaningful public test.

maliberty (approved):

PR #1703, opened 2023-12-15
DO NOT MERGE makefile: break out synthesis and .sdc copy dependencies

oharboe:
redoing synthesis is only necessary if ABC_CLOCK_PERIOD_IN_PS changed in the SDC_FILE.

This change allows Bazel ORFS layer to add a dependency on redoing synthesis when ABC_CLOCK_PERIOD_IN_PS changes, but skip synthesis(which can take hours for large designs) when other parts of the .sdc file changes

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -442,9 +442,8 @@ $(WRAPPED_LIBS):
 # |____/ |_| |_| \_| |_| |_| |_|_____|____/___|____/
 #
 .PHONY: synth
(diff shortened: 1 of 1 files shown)


oharboe:
Wrong approach. SDC_FILE should generate a ps.txt file in objects folder and then synthesis should depend on ps.txt. That way, Bazel can depend on ps.txt and synthesis is only re-run if the clock period changed.

PR #1704, opened 2023-12-15
makefile: synthesis does not need versions.txt

oharboe:


oharboe:
@maliberty unrelated failures

flow/Makefile
```diff
@@ -442,8 +442,7 @@ $(WRAPPED_LIBS):
 # |____/ |_| |_| \_| |_| |_| |_|_____|____/___|____/
 #
 .PHONY: synth
-synth: versions.txt \
```
maliberty, line 445: Something needs to trigger the generation of this file.  Removing it here means it wont' be generated by default

oharboe: What is it used for?

maliberty: It makes it easy to know what versions of tools someone is using.

oharboe: in make issue?

oharboe: > It makes it easy to know what versions of tools someone is using.

Except it doesn't do that... Only if the "synth" target is invoked explicitly is this file written... so `make issue` is likely to have the *wrong* version.

If I run  `make nuke && make DESIGN_CONFIG=designs/asap7/gcd/config.mk floorplan`, then no versions.txt is written.

maliberty: It is a quite old construct so perhaps it is no longer attached to the right target.   Attaching it to make issue seems reasonable.

oharboe: Then this PR is good to go...

```
$ ls versions.txt
ls: cannot access 'versions.txt': No such file or directory
$ make DESIGN_CONFIG=designs/asap7/gcd/config.mk floorplan_issue
[deleted]
$ ls versions.txt
versions.txt
$ cat versions.txt
Yosys 0.13+15 (git sha1 bc027b2ca, gcc 12.3.0-1ubuntu1~23.04 -fPIC -Os)
openroad v2.0-11432-gead3236e5
KLayout 0.28.11
$ echo >versions.txt blah
$ cat versions.txt
blah
$ make DESIGN_CONFIG=designs/asap7/gcd/config.mk floorplan_issue
[deleted]
$ cat versions.txt
Yosys 0.13+15 (git sha1 bc027b2ca, gcc 12.3.0-1ubuntu1~23.04 -fPIC -Os)
openroad v2.0-11432-gead3236e5
KLayout 0.28.11
$ 
```

maliberty (approved):

PR #1707, opened 2023-12-16
Updated ihp-sg13g2's fill.json

Syndace:
The `fill.json` for ihp-sg13g2 seemed to just be a copy of sky130hd's for the most part. This pr adjusts the rules to actually follow ihp-sg13g2's design rules:

- ihp-sg13g2 has a GDS layer/datatype specifically for filler shapes: \<layer\>/22
- ihp-sg13g2 recommends "brick"-shaped fillers, i.e. not perfect squares. The exact shapes used now are from IHP's design rule recommendations.
- Adjusted the space to fill/non-fill according to IHP's design rules.
- Removed `dir` from all layers, since that seems to be unused now.

I couldn't figure out what the `outline` & `space_to_outline` stuff does. Before merging, I would like to:

- Confirm that `dir` is actually unused now
- Understand what `outline` is and adjust the rules accordingly

### DIFF
--- a/flow/platforms/ihp-sg13g2/fill.json
+++ b/flow/platforms/ihp-sg13g2/fill.json
@@ -1,113 +1,106 @@
 {
     "layers" : {
         "Metal1" : {
-            "layer":                    8,
-            "name":                    "Metal1",
-            "dir":                     "V",
-            "datatype":                 0,
-            "space_to_outline":         70,
+            "layer": 8,
+            "datatype": 0,
+            "name": "Metal1",
+            "space_to_outline": 60,
             "non-opc": {
-                "datatype":             0,
-                "width":                [2.00, 1.00, 0.58, 0.3],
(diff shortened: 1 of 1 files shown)


maliberty:
dir is unused - probably left over from when we used klayout to do fill.

The outline is the outer edge and is a layer with a big rectangle.  In some PDKs there is a spacing rule to the outline that is different.  However I don't see it being used now either which is more surprising.  If you don't have such a rule I wouldn't worry about it.

Syndace:
Thank you, I have updated the rules accordingly. I don't think IHP has a "minimum distance from the outline" rule, especially given that the design is wrapped in a seal ring macro before tapeout, which contains a bit of margin. The outline is the `EdgeSeal.bnd` layer I think, so `39/4` should be correct.

Feel free to merge now.

maliberty (approved):

PR #1720, opened 2023-12-22
makefile: support synthesizing in Bazel only on changed clock period

oharboe:
There are many iterations on .sdc file where synthesis, which can take hours on large designs, is not required.

These changes allow the Bazel layer on ORFS not to resynthesize when the clock period did not change inside the .sdc file


oharboe:
@maliberty Needed by https://github.com/The-OpenROAD-Project/megaboom/pull/7

oharboe:
@maliberty unrelated gf180 failures.

maliberty (approved):

PR #1722, opened 2023-12-22
synthesis: add SYNTH_GUT to gut design

oharboe:
this is useful for quickly creating mock abstracts for faster iterations.

### DIFF
--- a/flow/scripts/synth.tcl
+++ b/flow/scripts/synth.tcl
@@ -5,8 +5,13 @@ if { [info exist ::env(SYNTH_HIERARCHICAL)] && $::env(SYNTH_HIERARCHICAL) == 1 &
   source $::env(SYNTH_STOP_MODULE_SCRIPT)
 }
 
+if { [info exist ::env(SYNTH_GUT)] && $::env(SYNTH_GUT) == 1 } {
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty unrelated gf180 failures.

flow/scripts/synth.tcl
```diff
@@ -5,8 +5,13 @@ if { [info exist ::env(SYNTH_HIERARCHICAL)] && $::env(SYNTH_HIERARCHICAL) == 1 &
   source $::env(SYNTH_STOP_MODULE_SCRIPT)
 }
 
+if { [info exist ::env(SYNTH_GUT)] && $::env(SYNTH_GUT) == 1 } {
+  hierarchy -check -top $::env(DESIGN_NAME)
+  delete $::env(DESIGN_NAME)/c:*
```
maliberty, line 12: What does c:* match?

oharboe: all cells, which quickly optimizes away everything else, including macros. Added comment.

I was going to delete wires too, but then I ran into a yosys snag and as it turns out, it's enough to delete the cells, optimization takes care of the rest...

https://github.com/YosysHQ/yosys/issues/4082

maliberty (approved):

PR #1737, opened 2024-01-02
Makefile floorplan sdc canonical form

oharboe:
This speeds up builds because it avoids unecessary rebuilds and also removes any source dependencies that SDC_FILE has, such as a utility.tcl are not then also source file dependencies for floorplan.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -490,9 +490,15 @@ do-yosys: yosys-dependencies
 $(RESULTS_DIR)/1_1_yosys.v: $(SDC_FILE_CLOCK_PERIOD)
 	$(UNSET_AND_MAKE) do-yosys
 
-$(RESULTS_DIR)/1_synth.sdc: $(SDC_FILE)
-	mkdir -p $(REPORTS_DIR)
-	cp $(SDC_FILE) $(RESULTS_DIR)/1_synth.sdc
+$(RESULTS_DIR)/1_synth.sdc: $(SDC_FILE) $(RESULTS_DIR)/1_1_yosys.v
+	$(UNSET_AND_MAKE) do-synth-sdc
+
+.PHONY: do-synth-sdc
+do-synth-sdc:
+	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)
+	cp $(RESULTS_DIR)/1_1_yosys.v $(RESULTS_DIR)/1_synth.v
+	cp $(SDC_FILE) $(RESULTS_DIR)/1_1_yosys.sdc
+	$(OPENROAD_CMD) $(SCRIPTS_DIR)/write-sdc.tcl
 
 .PHONY: do-synth
 do-synth:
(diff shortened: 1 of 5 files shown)


flow/Makefile
```diff
@@ -490,9 +490,15 @@ do-yosys: yosys-dependencies
...
-	cp $(SDC_FILE) $(RESULTS_DIR)/1_synth.sdc
+$(RESULTS_DIR)/1_synth.sdc: $(SDC_FILE) $(RESULTS_DIR)/1_1_yosys.v
+	$(UNSET_AND_MAKE) do-synth-sdc
+
+.PHONY: do-synth-sdc
+do-synth-sdc:
+	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR)
+	cp $(RESULTS_DIR)/1_1_yosys.v $(RESULTS_DIR)/1_synth.v
+	cp $(SDC_FILE) $(RESULTS_DIR)/1_1_yosys.sdc
+	$(OPENROAD_CMD) $(SCRIPTS_DIR)/write-sdc.tcl
```
maliberty, line 501: Why not do this at the start of floorplan.tcl?

maliberty (approved):

maliberty:
I didn't mean to approve, just comment.

flow/Makefile
oharboe: Because this is the *output* of the synthesis stage. 

I don't want to rebuild the floorplan if there's no change to the canonical 1_synth.sdc.

oharboe:
> I didn't mean to approve, just comment.

Is it necessary to load the .v file or can I just load and write out the .sdc file?

oharboe:
> > I didn't mean to approve, just comment.
> 
> Is it necessary to load the .v file or can I just load and write out the .sdc file?

```
--- a/flow/scripts/write-sdc.tcl
+++ b/flow/scripts/write-sdc.tcl
@@ -1,3 +1,3 @@
 source $::env(SCRIPTS_DIR)/load.tcl
-load_design 1_1_yosys.v 1_1_yosys.sdc
+read_sdc $::env(SDC_FILE)
```

Apparently not... "[ERROR STA-0201] no network has been linked."

flow/Makefile
maliberty: It isn't an output of yosys so its debatable if it is a synthesis output.  If you are going to run floorplan next then this duplicates all the parsing and sta that will happen in that stage.

Should this be a side target if you are not planning to run beyond synth?

oharboe:
@maliberty I understand the concern about increased running times, but if we respect the interface of 'synth' stage, then SDC_FILE comes in and 1_synth.sdc comes out.

Are you suggesting we change that interface?

maliberty:
Today we just do 
```
$(RESULTS_DIR)/1_synth.sdc: $(SDC_FILE)
	mkdir -p $(REPORTS_DIR)
	cp $(SDC_FILE) $(RESULTS_DIR)/1_synth.sdc
```
which requires nothing more than a cp.  You are suggesting a new and more expensive flow to produce it but I'm not clear why.

oharboe:
> Today we just do
> 
> ```
> $(RESULTS_DIR)/1_synth.sdc: $(SDC_FILE)
> 	mkdir -p $(REPORTS_DIR)
> 	cp $(SDC_FILE) $(RESULTS_DIR)/1_synth.sdc
> ```
> 
> which requires nothing more than a cp. You are suggesting a new and more expensive flow to produce it but I'm not clear why.

I see. The reason why is that I don't want to rerun floorplan when there are no changes in the canonical sdc.

However, I have enough going on, so I'm going to close this PR for now and resurrect it when the dust settles.

PR #1739, opened 2024-01-02
checkpoint: cleanup, remove vestiges of checkpoint

oharboe:
this variable is never set anywhere, so this conditional code is a vestige of something or other.

### DIFF
--- a/flow/scripts/cts.tcl
+++ b/flow/scripts/cts.tcl
@@ -125,10 +125,8 @@ if { [info exists ::env(POST_CTS_TCL)] } {
   source $::env(POST_CTS_TCL)
 }
 
-if {![info exists save_checkpoint] || $save_checkpoint} {
-  if {[info exists ::env(GALLERY_REPORT)]  && $::env(GALLERY_REPORT) != 0} {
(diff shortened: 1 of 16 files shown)


maliberty (approved):

rovinski:
This was part of the `run_all.tcl` script removed in #1639. The idea was that it could source every step and run through completely without having to read/write files between steps. The risk is that if the flow crashed or errored out, then nothing would be saved, so the `save_checkpoint` variable was an option to keep the intermediate files in order to better resume after a failure.

That workflow wasn't the default method and no one really knew about it because it only worked best if the flow was expected to fully complete without error.

Just adding context :)

maliberty:
In theory it saves some runtime but as nobody was using it (or testing it) I felt it was time to retire it.  It also used to give different results due to the sorting of the def files between steps.  With the switch to .odb that would have been less of an issue.

rovinski:
I never used it either 😆

maliberty:
@oharboe please resolve the conflicts (presumably with your last PR)

oharboe:
> @oharboe please resolve the conflicts (presumably with your last PR)

Fixed.

PR #1752, opened 2024-01-13
genMetrics: less conditional code, less repetition

oharboe:
Also, what exactly motivates yosys to output 'top ' here isn't entirely clear.

This paves the way for a followup pull request

### DIFF
--- a/flow/util/genMetrics.py
+++ b/flow/util/genMetrics.py
@@ -249,12 +249,10 @@ def extract_metrics(cwd, platform, design, flow_variant, output, hier_json):
                        'Number of cells: +(\S+)',
                        rptPath + '/synth_stat.txt')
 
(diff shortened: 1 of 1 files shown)


oharboe:
@vvbandeira unrelated CI error, this error is on master.

flow/util/genMetrics.py
```diff
@@ -251,16 +251,10 @@ def extract_metrics(cwd, platform, design, flow_variant, output, hier_json):
 
     hierarchicalSynth = os.environ.get('SYNTH_HIERARCHICAL')
```
vvbandeira: This is no longer needed, it was only used in the if/else statement. Could you remove it?

vvbandeira (changes requested):

vvbandeira (approved):

PR #1771, opened 2024-01-23
util: add design and platform files to make issue

vvbandeira:


vvbandeira:
Including all the files from `DESIGN_DIR` and `PLATFORM_DIR` can make the tar file size almost double in some cases. This option saves only the `config.mk` files and allows the user to set a variable to save all files for corner cases where it might be useful.

maliberty (approved):

PR #1778, opened 2024-01-25
make issue: default to full issue

oharboe:
There are very few small uncomplicated test-cases these days.

The biggest size saver is to manually prune .odb files that are not needed, which has to be done manually anyway.

### DIFF
--- a/flow/util/makeIssue.sh
+++ b/flow/util/makeIssue.sh
@@ -92,10 +92,10 @@ fi
 echo "Using $COMPRESS to compress tar file"
 
 # Save all files inside design and platform folders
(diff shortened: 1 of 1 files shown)


maliberty:
If its a public PDK then including a copy is generally not needed.  If its a private PDK then including a copy is generally the wrong thing to do (unless there is a 3-way NDA in place).

oharboe:
> If its a public PDK then including a copy is generally not needed. If its a private PDK then including a copy is generally the wrong thing to do (unless there is a 3-way NDA in place).

So this is a correctness, not size concern?

PR #1780, opened 2024-01-25
[NOT READY FOR REVIEW] Fifo upgrade

oharboe:
@tspyrou I have created a "mock CPU" where there is a single clock in, an asynchronous FIFO between the pins of the CPU and the core. The core has many flip-flops.

How can I articulate a .sdc file that gives me two clock trees?

I want one clock tree for the core side of the async FIFO and one for the other on the pins side. All the pins are at the bottom, so the async FIFO is at the bottom and there is a really short path from the clock pin to the async FIFO.

```
make DESIGN_CONFIG=designs/asap7/fifo_oneclock/config.mk cts
```

image

Here we can see that it doesn't meet timing because the clock latency of the CPU side is added to the other side of the FIFO:

image


flow/designs/asap7/fifo_oneclock/constraint.sdc
```diff
@@ -0,0 +1,26 @@
+# https://gist.github.com/brabect1/7695ead3d79be47576890bbcd61fe426
+#
```
oharboe: @tspyrou This is the .sdc file that I would like to have "upgraded" so that it creates two clock trees for the mock CPU core: one clock tree on the "external" side of the async FIFO and one on the CPU core side of the FIFO.

tspyrou: @oharboe I want to make sure I understand. Do you want to create two additional clocks, one which is outside of the fifo for the input signals and one which is outside of the fifo for the output signals?

oharboe:
@tspyrou Now with two nets, 'clk' and 'clk_uncore'. 'uncore' is used for the clock that is external to the CPU, generally peripheral buses, etc. for Chipyard MegaBoom.

image
image

oharboe:
@tspyrou This constraint is wrong. This flipflop on the core clock has had time to settle. The flip flop from the read pointer for the output fifo is on the uncore clock and the read pointer won't point to a fifo output that hasn't settled.

image

PR #1787, opened 2024-02-01
Added aarch64 architecture support for Ubuntu 22.04

liav-vorak:
The following has been done:

- Modified the Klayout tools installation process to dynamically clone and build it from source since there are no prebuilt binaries available for the "aarch64" architecture.

- The modification has been developed and tested successfully on an AWS Graviton (aarch64) instance using the following specifications: Instance type - t4g.xlarge (4 vCPU, 16 GiB Memory), Storage - 128 GiB.

Integration with Other Changes:

This pull request is associated with another changes in the OpenROAD  repository. Specifically, adjustments were made in the DependancyInstaller.sh script to accommodate dependencies for the aarch64 architecture. Further details and the link to the related pull request will be provided in the comments.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -86,20 +86,29 @@ _installUbuntuPackages() {
     if _versionCompare $1 -ge 23.04; then
         apt-get install klayout python3-pandas
     else
-        if [[ $1 == 20.04 ]]; then
-            klayoutChecksum=15a26f74cf396d8a10b7985ed70ab135
-        else
-            klayoutChecksum=db751264399706a23d20455bb7624264
-        fi
+        arch=$(uname -m)
         lastDir="$(pwd)"
         # temp dir to download and compile
         baseDir=/tmp/installers
         mkdir -p "${baseDir}"
-        cd ${baseDir}
-        wget https://www.klayout.org/downloads/Ubuntu-${1%.*}/klayout_${klayoutVersion}-1_amd64.deb
-        md5sum -c <(echo "${klayoutChecksum} klayout_${klayoutVersion}-1_amd64.deb") || exit 1
-        dpkg -i klayout_${klayoutVersion}-1_amd64.deb
-        cd ${lastDir}
+        cd "${baseDir}"
+        if [[ $arch == "aarch64" ]]; then
+            echo "Installing KLayout for aarch64 architecture"
+            installDir="/usr/local/bin"
+            git clone https://github.com/KLayout/klayout.git
+            cd klayout
+            ./build.sh -bin "${installDir}"
+        else
+            if [[ $1 == 20.04 ]]; then
+                klayoutChecksum=15a26f74cf396d8a10b7985ed70ab135
+            else
+                klayoutChecksum=db751264399706a23d20455bb7624264
+            fi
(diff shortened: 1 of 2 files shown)


etc/DependencyInstaller.sh
```diff
@@ -86,20 +86,33 @@ _installUbuntuPackages() {
...
-        cd ${baseDir}
-        wget https://www.klayout.org/downloads/Ubuntu-${1%.*}/klayout_${klayoutVersion}-1_amd64.deb
-        md5sum -c <(echo "${klayoutChecksum} klayout_${klayoutVersion}-1_amd64.deb") || exit 1
-        dpkg -i klayout_${klayoutVersion}-1_amd64.deb
-        cd ${lastDir}
+        cd "${baseDir}"
+        if [[ $arch == "aarch64" ]]; then
+        if [ ! -f /usr/local/bin/klayout ]; then
+            echo "Installing KLayout for aarch64 architecture"
+            installDir="/usr/local/bin"
```
rovinski: I don't think this should default to `/usr/local/bin`, it should default to where the other packages from source are built.

etc/DependencyInstaller.sh
```diff
@@ -86,20 +86,33 @@ _installUbuntuPackages() {
...
+            git clone https://github.com/KLayout/klayout.git
+            cd klayout
+            ./build.sh -bin "${installDir}"
```
rovinski: Doesn't KLayout have several additional dependencies that need to be installed in order to get it to build?

liav-vorak: Hi @rovinski 
Thanks for your comment. Indeed, KLayout has dependencies that require installation, the necessary changes have already been pushed․ Regarding the installDir="/usr/local/bin" path defined for installation of Klayout, I included the PREFIX in a manner consistent with the approach used in the OpenROAD/etc/DependencyInstallers.sh file during the installation of various dependencies, however Klayout installation is done when the script is first time invoked from setup.sh with "--base" option which according to the help message is not compatible with "--prefix" option, so the installation will take place at "/usr/local" directory. Please let me know if there are any open questions that I can address.

Regards
Lia

etc/DependencyInstaller.sh
```diff
@@ -86,20 +99,34 @@ _installUbuntuPackages() {
...
+        klayoutPrefix=${PREFIX:-"/usr/local"}
         mkdir -p "${baseDir}"
-        cd ${baseDir}
-        wget https://www.klayout.org/downloads/Ubuntu-${1%.*}/klayout_${klayoutVersion}-1_amd64.deb
-        md5sum -c <(echo "${klayoutChecksum} klayout_${klayoutVersion}-1_amd64.deb") || exit 1
-        dpkg -i klayout_${klayoutVersion}-1_amd64.deb
-        cd ${lastDir}
+        cd "${baseDir}"
+        if [[ $arch == "aarch64" ]]; then
+        if [ ! -f ${klayoutPrefix}/klayout ]; then
```
vvbandeira: Looks to be missing an indentation level.

vvbandeira (changes requested):
It's just a tiny nit about indentation. Otherwise, it looks good. I want to wait until I can deploy a CI machine with aarch64 so we can adequately test the PR and make sure it will not break silently in the future.

vvbandeira (approved):

PR #1793, opened 2024-02-08
[DO NOT REVIEW] CTS improvements CI test

oharboe:


oharboe:
@precisionmoon FYI, enabling new CTS and taking it for a spin to see what happens.

oharboe:
@vvbandeira I want to test a draft pull request + some additional changes on latest master ORFS. Is that not allowed?

oharboe:
@precisionmoon mock-array will need some adjustments, but other than that the new CTS feature could be enabled by default.

precisionmoon:
> @precisionmoon mock-array will need some adjustments, but other than that the new CTS feature could be enabled by default.

Thank you very much for your help with validation.

PR #1794, opened 2024-02-09
mock-array: use clock tree insertion latency of macros to reduce skew

oharboe:
From 400 to 250ps clock period.

There is a tiny negative slack, not a bad thing in testing, but perhaps that will go away once CTS improvements are done.


oharboe:
@precisionmoon FYI

flow/designs/asap7/mock-array/Element/config.mk
```diff
@@ -43,3 +43,5 @@ export PLACE_PINS_ARGS = -annealing
 
 export GND_NETS_VOLTAGES      =
 export PWR_NETS_VOLTAGES      =
+
+export CTS_ARGS = -insertion_delay -sink_clustering_enable -balance_levels -distance_between_buffers 60
```
maliberty, line 47: Why are you adding -distance_between_buffers 60 which seems unrelated.

oharboe: See implementation of CTS_ARGS. This is copy, paste & modify of the default CTS args.

maliberty: That is not in the default args:

```
if {[info exist ::env(CTS_BUF_DISTANCE)]} {
  lappend cts_args -distance_between_buffers "$::env(CTS_BUF_DISTANCE)"
}
```

oharboe: I copied and pasted from the logs, so, I guess it is...

maliberty: Apparently it is in the asap7 platform config (which I've never noticed).

maliberty (approved):

PR #1796, opened 2024-02-10
Mock array no insbuf

oharboe:


flow/designs/asap7/mock-array/Element/config.mk
```diff
@@ -43,3 +43,7 @@ export PLACE_PINS_ARGS = -annealing
 
 export GND_NETS_VOLTAGES      =
 export PWR_NETS_VOLTAGES      =
+
+export CTS_ARGS = -insertion_delay -sink_clustering_enable -balance_levels -distance_between_buffers 60
```
maliberty, line 47: You appear to have mixed your other PR into this one.  This serializes the merge process.

maliberty:
What is the use model here?  Its not slow so I'm not sure how it relates to archictural eval.

oharboe:
> What is the use model here? Its not slow so I'm not sure how it relates to archictural eval.

This answers the question of what the performance is in the limit that all buffers are removed. There are some superfluous buffers today.

For arch exploration, the superfluous buffers do not have to be removed in real designs, it can be sufficient to have a lower bound to bottom out on discussions

maliberty:
If by "superfluous buffers" you are referring to the bug I would just fix that rather than add this option.  If you think the i/o buffers are superfluous that is a different idea and I think a bad one.  You will not get a meaningful timing model with a feedthrough.

oharboe:
@maliberty If https://github.com/The-OpenROAD-Project/OpenROAD/pull/4654 can take us down to 1 buffer, then that's good enough, I think.

PR #1801, opened 2024-02-13
DO NOT MERGE mock automatic update of clock insertion latency

oharboe:
@maliberty @tspyrou A mock implementation of automatic adjustment of clock instertion latency.

To build:

```
make DESIGN_CONFIG=designs/asap7/mock-array/config.mk
```

### DIFF
--- a/flow/scripts/cts.tcl
+++ b/flow/scripts/cts.tcl
@@ -73,6 +73,11 @@ if {[info exist ::env(CTS_SNAPSHOTS)]} {
   save_progress 4_1_pre_repair_hold_setup
 }
 
+write_timing_model latency.lib
+# FIXME! How do I find the average clock insertion latency?
+set max_latency [exec bash -c {python -c "import re; print(re.search('max_clock_tree_path.*?cell_fall\(scalar\).*?values\(\"(.*?)\"\)', open('latency.lib').read(), re.DOTALL).group(1))"}]
+set_clock_latency [expr $max_latency] [get_clocks {clock_vir}]
+
 puts "Repair setup and hold violations..."
 
 # process user settings
--- a/flow/.gitignore
+++ b/flow/.gitignore
@@ -2,3 +2,4 @@ settings.mk
 vars.sh
 vars.gdb
 vars.tcl
+latency.lib
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -748,6 +748,8 @@ route: $(RESULTS_DIR)/5_route.odb \
 #-------------------------------------------------------------------------------
 $(eval $(call do-step,5_1_grt,$(RESULTS_DIR)/4_cts.odb $(FASTROUTE_TCL) $(PRE_GLOBAL_ROUTE),global_route))
 
+$(RESULTS_DIR)/5_1_grt.sdc: $(RESULTS_DIR)/5_1_grt.odb
+
 # SEP 2: Filler cell insertion
 # ------------------------------------------------------------------------------
 $(eval $(call do-step,5_2_fillcell,$(RESULTS_DIR)/5_1_grt.odb,fillcell))
@@ -762,7 +764,7 @@ endif
 
 $(eval $(call do-copy,5_route,5_3_route.odb))
 
-$(eval $(call do-copy,5_route,4_cts.sdc,,.sdc))
+$(eval $(call do-copy,5_route,5_1_grt.sdc,,.sdc))
 
 .PHONY: do-route
 do-route:
@@ -833,7 +835,7 @@ endif
 
 $(eval $(call do-copy,6_1_fill,5_route.sdc,,.sdc))
 
-$(eval $(call do-copy,6_final,5_route.sdc,,.sdc))
+$(RESULTS_DIR)/6_final.sdc: $(RESULTS_DIR)/6_final.odb
 
 $(eval $(call do-step,6_report,$(RESULTS_DIR)/6_1_fill.odb $(RESULTS_DIR)/6_1_fill.sdc,final_report,.log,$(LOG_DIR)))
 
--- a/flow/designs/asap7/mock-array/Element/config.mk
+++ b/flow/designs/asap7/mock-array/Element/config.mk
@@ -45,3 +45,5 @@ export GND_NETS_VOLTAGES      =
 export PWR_NETS_VOLTAGES      =
 
 export CTS_ARGS = -insertion_delay -sink_clustering_enable -balance_levels -distance_between_buffers 60
+
+export DONT_BUFFER_PORTS = 1
--- a/flow/designs/asap7/mock-array/Element/constraints.sdc
+++ b/flow/designs/asap7/mock-array/Element/constraints.sdc
@@ -5,9 +5,9 @@ set cols [expr {[info exists ::env(MOCK_ARRAY_COLS)] ? $::env(MOCK_ARRAY_COLS) :
 set clk_name clock
 set clk_port_name clock
 set clk_period 300
-set clk_imax_pct 0.75
-set clk_imin_pct 0.10
-set clk_omax_pct 0.80
+set clk_imax_pct 0.20
+set clk_imin_pct 0.20
+set clk_omax_pct 0.20
 set clk_omin_pct 0.10
 
 set clk_port [get_ports $clk_port_name]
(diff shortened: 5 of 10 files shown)


flow/scripts/cts.tcl
```diff
@@ -73,6 +73,11 @@ if {[info exist ::env(CTS_SNAPSHOTS)]} {
   save_progress 4_1_pre_repair_hold_setup
 }
 
+write_timing_model latency.lib
+# FIXME! How do I find the average clock insertion latency?
+set max_latency [exec bash -c {python -c "import re; print(re.search('max_clock_tree_path.*?cell_fall\(scalar\).*?values\(\"(.*?)\"\)', open('latency.lib').read(), re.DOTALL).group(1))"}]
+set_clock_latency [expr $max_latency] [get_clocks {clock_vir}]
```
maliberty, line 79: What is the purpose of this?  You have a clock tree so why set clock latency at all when you can use propagated clocks?

oharboe: Otherwise you get a ton of hold cells.

oharboe: recall: Louii and My adviced this approach last summer and you/tom approved...

oharboe: without this I also get setup violations

tspyrou, line 79: If this is the expected external clock latency (from the perspective of this block) then it makes sense. If it is clock latency within the block, that should be picked up automatically from set_propagated_clock.

oharboe: in a sense it is both. Cho's CTS changes will deliver the clock earlier than the flipflops at the next level up by this much

oharboe: and the clock insertion latency within the macro should be about the same.

oharboe:
@rovinski A rough prototype of auromatically propagating clock latency.

rovinski:
@oharboe can you elaborate on the issue you are seeing and what you are trying to fix? It isn't clear to me from the discussion so far.

One issue that can come up is that reasonable I/O constraints with an ideal clock can become unreasonable constraints after adding clock propagation delay. In many cases, the designer would like to set the I/O constraints relative to the clock tree latency, because it is assumed that the inputs and outputs are coming from/going to endpoints with similar clock tree latencies. This is especially true in array-based architectures.

As an example, let's say that the output delay on a pin is 500 ps with an ideal clock. The clock period is 1000 ps, so there is 500 ps available to hit the constraint. But after clock propagation, the clock latency might be 1000ps. That means it is now impossible to hit the constraint. But in reality you would like to offset the constraint by 1000ps to maintain the original intention.

It would be nice if there was a clean way to handle this in SDC, but to the best of my knowledge, there isn't. In a previous project, I simply ran the implementation once, collected the average clock latency, and then added that number to the constraints for each I/O pin.

But in hindsight, a better way to do this might be to use the ideal latency up until cts, then once cts is complete and the average latency is known, use that number as an offset for all the I/O.

It seems like that might be what you're trying to do here, but I am not positive if `set_clock_latency` is the right approach. I am not sure how OpenSTA handles it, but I believe that `set_clock_latency` usually is ignored once clocks are propagated.

An approach I might suggest is to get the clock latency, hopefully through Tcl commands, and then directly add that to the I/O constraints through `set_input_delay` / `set_output_delay` `-add_delay`. In this case, I would consider the clock latency to be (min clock latency + max clock latency)/2. In a more stringent approach, I think you would only need to consider the min/max clock latencies for the endpoints which are directly connected to an I/O.

oharboe:
@rovinski From what you write, it seems to me that you understood my intent. You add some color w.r.t. what OpenSTA supports and how to express the constraints post CTS.

I think all that is needed here is to come up with one or more valid ways to express this and then test and make sure that OpenSTA supports all those approaches that should be supported. Minimally one though..... 

I am fumbling in the dark, because I dont know how to express it in .sdc and I dont know what effect these statements should have and I cant distunguish if features are missing, broken or working as intended in OpenSTA.

oharboe:
@rovinski Trying to articulate with the vocabulary you use:

1. i want to specify input/output delay constraints with ideal clock
2. after CTS, prior to timing repair, the input output delays should be adjusted by (max/min clock insertion latency)/2.
3. again after route and final, where clock insertion latency changes again, the input output latency should be adjusted as in 2 above.

The only way I know how to articulate this, where it works, is with the virtual clock in mock array, though I am sure there are other.

rovinski:
Understood. So I think there are a few different ways to accomplish this notion:

1. One of the built-in approaches is to use `set_input_delay -reference_pin $pin` where `$pin` is the clock pin of the register that is connected to the I/O.
    * Upside is that it doesn't require finding the clock tree latency, it is built into the command by adding the latency from the clock root to the register pin. Essentially if you start with a 500ps constraint before CTS, you will have a 500ps constraint after CTS too.
    * Downside is that if the clock tree is very unbalanced, this will end up hiding the problem. If you have 1500ps of clock latency to one register but 500ps of clock latency to the other registers, it will hide it.
    
image

2. Another approach is `set_network_latency -source -$avg_clock_latency` after CTS. What this does is basically try to cancel out the average clock tree latency, but still maintain the skew between registers.
    * Upside: this solves the problem in (1).
    * Downside: it requires collecting the average clock latency. It also may cause timing reports to look a bit weird because your clock latencies effectively disappear. Your timing reports will show clock latency to the register, then subtract out the average latency, leaving just the difference from the average. You may like this, you may not. I think it's a personal preference.

image

3. A third approach is to `set_input_delay -add -$avg_clock_latency [all_inputs]` after CTS. What this does is instead of trying to cancel out the average clock tree latency as in (2), it adds the average clock tree latency to the required time.
    * Upside: also solves the problem in (1).
    * Downside: requires collecting the average clock latency. Adds clock latency to the constraints so it may make the original pre-CTS constraints unclear in the timing report. This is a personal preference vs. (2)

image


I have used (3) before in the past. Anyways, it looks like your strategy is on the right path. The main thing is to extract the average clock latency (which should be able to be done from a Tcl command I hope... I'll look into that). And then add the clock latency given approach 1 or 2, whichever is your preference.

I will start looking more into this.

oharboe:
@rovinski In your summary above, you dont mention that the clock insertion latency changes significantly between CTS, route and final stages, so adjustment for average latency of min. max must be applied multiple time. If I understand correctly, the constraints only affect the results during repair timing during CTS. Get the input latency wrong and you get a ton of hold cells. You dont mention set_clock_latency and virtual clock as mock-array uses today as an alternative, is there a reason to exclude it?

I get the impression that you have dismissed 1., makes sense to me as taking skew into account is critical.

rovinski:
> adjustment for average latency of min. max must be applied multiple time.

That's fair. In commercial tools I haven't typically seen the clock tree latency change dramatically after construction, but OR could be different. It shouldn't be dramatically different from pre-route to post route though... that means there is bad parasitics correlation which would be a separate problem.

> If I understand correctly, the constraints only affect the results during repair timing during CTS. Get the input latency wrong and you get a ton of hold cells.

Currently in ORFS, ~~CTS is the last stage which any timing optimizations are applied~~. But these constraints are still used in the check during final reporting.

> You dont mention set_clock_latency and virtual clock as mock-array uses today as an alternative, is there a reason to exclude it?

It's not an exhaustive list, I'm sure there are other ways to get the same effect. Although, I'm not what that approach is? `set_clock_latency` only works before clocks are propagated. `set_propagated_clock` will overwrite a `set_clock_latency` and vice versa.

> I get the impression that you have dismissed 1., makes sense to me as taking skew into account is critical.

I think approach 1 could work if the clock tree is usually balanced. In designs with lots of macros it usually isn't the case, and I believe approach 1 would hide skew problems in the timing report.

oharboe:
set_clock_latency is used with a virtual clock, which is not propagated. read .sdc file for mock array for details.

maliberty:
> Currently in ORFS, CTS is the last stage which any timing optimizations are applied. 

This is not true since we added post-grt fixing.  https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/067099e79f308b77cb7f031f37d2f9ca2ac25b7b/flow/scripts/global_route.tcl#L72

flow/scripts/cts.tcl
tspyrou: The propagated clocks work at the top level but when scoped on the block itself the top level latency is pushed into the blocks with set_clock_latency.

oharboe:
Restore later if still relevant

PR #1807, opened 2024-02-17
Flow: Use Yosys/abc native functions for dont_use

### LINKED ISSUE #1502: Update scripts to use Ethan's new `dont_use` functionality
### Description

@QuantamHD landed a change into both Yosys and ABC to better support `dont_use` functionality. OpenROAD-flow-scripts should probably be updated to use that functionality.

 * Yosys change - https://github.com/YosysHQ/yosys/pull/3892
 * ABC change - https://github.com/berkeley-abc/abc/pull/239

The equivalent change in Silicon compiler can be found @ https://github.com/siliconcompiler/siliconcompiler/pull/1998/files

### Suggested Solution

_No response_

### Additional Context

One downside is that it means OpenROAD-flow-scripts will depend on having a new enough version of Yosys and ABC.

rovinski:
* Change the Makefile target for `DONT_USE_LIBS` to `DECOMPRESSED_LIBS`
  * Removes the call to markDontUse.py, replaces it with either gunzip or ln
* Switch yosys/abc to use the "dont use" flags available
* Switch all uses of `DONT_USE_SC_LIB` to `SC_LIB`
* Add missing `DFF_LIB_FILE` var to asap7

Closes #1502

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -313,10 +313,12 @@ endif
 WRAPPED_LEFS = $(foreach lef,$(notdir $(WRAP_LEFS)),$(OBJECTS_DIR)/lef/$(lef:.lef=_mod.lef))
 WRAPPED_LIBS = $(foreach lib,$(notdir $(WRAP_LIBS)),$(OBJECTS_DIR)/$(lib:.lib=_mod.lib))
 export ADDITIONAL_LEFS += $(WRAPPED_LEFS) $(WRAP_LEFS)
-export LIB_FILES += $(WRAP_LIBS) $(WRAPPED_LIBS)
 
-export DONT_USE_LIBS   = $(patsubst %.lib.gz, %.lib, $(addprefix $(OBJECTS_DIR)/lib/, $(notdir $(LIB_FILES))))
-export DONT_USE_SC_LIB ?= $(firstword $(DONT_USE_LIBS))
+export LIB_FILES += $(WRAP_LIBS) $(WRAPPED_LIBS)
+export DECOMPRESSED_LIBS = $(patsubst %.lib.gz, %.lib, $(addprefix $(OBJECTS_DIR)/lib/, $(notdir $(LIB_FILES))))
+
+# Assume the first lib file is the standard cell lib
+export DECOMPRESSED_SC_LIB ?= $(firstword $(DECOMPRESSED_LIBS))
 
 # Stream system used for final result (GDS is default): GDS, GSDII, GDS2, OASIS, or OAS
 STREAM_SYSTEM ?= GDS
@@ -371,15 +373,19 @@ versions.txt:
 # Pre-process libraries
 # ==============================================================================
 
-# Create temporary Liberty files which have the proper dont_use properties set
-# For use with Yosys and ABC
+# Unzip any compressed Liberty files
 .SECONDEXPANSION:
-$(DONT_USE_LIBS): $$(filter %$$(@F) %$$(@F).gz,$(LIB_FILES))
-	@mkdir -p $(OBJECTS_DIR)/lib
-	$(UTILS_DIR)/markDontUse.py -p "$(DONT_USE_CELLS)" -i $^ -o $@
+$(DECOMPRESSED_LIBS): $$(filter %$$(@F) %$$(@F).gz,$(LIB_FILES))
+	mkdir -p $(OBJECTS_DIR)/lib
+	if [[ $^ == *.gz ]]; then \
+	  gunzip -c $^ > $@; \
+	else \
+	  ln -s $$(realpath --relative-to=$(dir $@) $^) $@; \
+	fi
 
-$(OBJECTS_DIR)/lib/merged.lib:
-	$(UTILS_DIR)/mergeLib.pl $(PLATFORM)_merged $(DONT_USE_LIBS) > $@
+# Create merged Liberty file for use with Yosys and ABC
+$(OBJECTS_DIR)/lib/merged.lib: $(DECOMPRESSED_LIBS)
+	$(UTILS_DIR)/mergeLib.pl $(PLATFORM)_merged $^ > $@
 
 # Pre-process KLayout tech
 # ==============================================================================
@@ -480,7 +486,7 @@ $(SDC_FILE_CLOCK_PERIOD): $(SDC_FILE)
 	echo $(ABC_CLOCK_PERIOD_IN_PS) > $@
 
 .PHONY: yosys-dependencies
-yosys-dependencies: $(DONT_USE_LIBS) $(WRAPPED_LIBS) $(DONT_USE_SC_LIB) $(DFF_LIB_FILE) $(VERILOG_FILES) $(CACHED_NETLIST) $(LATCH_MAP_FILE) $(ADDER_MAP_FILE)
+yosys-dependencies: $(DECOMPRESSED_LIBS) $(WRAPPED_LIBS) $(DECOMPRESSED_SC_LIB) $(DFF_LIB_FILE) $(VERILOG_FILES) $(CACHED_NETLIST) $(LATCH_MAP_FILE) $(ADDER_MAP_FILE)
 
 .PHONY: do-yosys
 do-yosys: yosys-dependencies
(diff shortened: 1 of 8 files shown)


maliberty:
Ready for review?

rovinski:
Not yet, ASAP7 needs fixing

rovinski:
I went down a bit of a rabbit hole. It appears that `markDontUse.py` was fulfilling multiple duties. Not only was it altering the `.lib` properties, but it was also uncompressing any Gzipped files first. Neither Yosys nor abc support native decompression, so this is still required. ~~ASAP7 is the only platform with Gzipped Liberty files right now.~~ GF180 uses gzipped files too.

But, there is also a step which does a merging of the liberty files. I wondered if this step is still required, and as far as I can tell it is. The main step that requires all the `LIB_FILES` is the `abc` command from inside `yosys`. This command allegedly _does_ support multiple `.lib` files because you can repeat the `-liberty` with new files and each one gets pushed to a vector. This gets translated to abc via multiple `read_lib` commands in the generated abc script. This does not work as intended, because abc apparently overwrites or at least only looks at the most recent .lib file that was read.

```
ABC: ** cmd error: aborting 'source /home/rovinski/OpenROAD-flow-scripts/flow/scripts/abc_speed.script'
ABC: ** cmd error: aborting 'source <abc-temp-dir>/abc.script'
ABC: Error: Cannot find buffer gate in the library.
```

abc _does_ read the ASAP7 BUFINV liberty, but it wasn't last, and it complains about all the other libraries not having a buffer.

Weirdly enough, it looks like abc supports specifically 1-2 lib files.

```
Usage: read_lib [-SG float] [-M num] [-dnuvwh] [-X cell_name] <file> <file2>
                   reads Liberty library from file
        -S float : the slew parameter used to generate the library [default = 0.00]
        -G float : the gain parameter used to generate the library [default = 0.00]
        -M num   : skip gate classes whose size is less than this [default = 0]
        -X name  : adds name to the list of cells ABC shouldn't use. Flag can be passed multiple times
        -d       : toggle dumping the parsed library into file "*_temp.lib" [default = no]
        -n       : toggle replacing gate/pin names by short strings [default = no]
        -u       : toggle setting unit area for all cells [default = no]
        -v       : toggle writing verbose information [default = yes]
        -w       : toggle writing information about skipped gates [default = no]
        -h       : prints the command summary
        <file>   : the name of a file to read
        <file2>  : the name of a file to read (optional)
```
But if I try using any more than 2 files it prints the usage message instead.

So the conclusion is that while we can eliminate modifying the `.lib` `dont_use` properties, we still need to decompress files, and we still need to merge libs into one standard cell `.lib`.

It will have to stay this way until abc adds support for reading >2 liberty files. It is also problematic that yosys allows passing of multiple liberty files to abc because it is very misleading.

maliberty:
@QuantamHD perhaps you can ask Alan about this abc behavior.

flow/scripts/sc/tools/openroad/sc_apr.tcl
```diff
@@ -27,13 +27,17 @@ set inputs [list]
...
 if {$sc_step == "or_synth"} {
-    # Pre-synthesis: mark dont-use cells in liberty files, and merge them.
+    # Pre-synthesis: decompress liberty files, and merge them.
     foreach f [split $::env(LIB_FILES)] {
-        exec $::env(UTILS_DIR)/markDontUse.py -p $::env(DONT_USE_CELLS) -i $f -o "../../[file tail $f]-mod.lib"
+        if {[file extension $f] == .gz} {
+            exec gunzip -c $f > "../../[file tail $f]-mod.lib"
+        } else {
+            file link -symbolic $f "../../[file tail $f]-mod.lib"
+        }
```
rovinski, line 36: @gadfort can you review this please? I don't know how to test.

oharboe: @maliberty We went down a rabbithole og symlinks and tcl. TL;DR symlinks cant be made to work w makefiles and tcl file timestamps are truncated to whole seconds.

Does that cause a proble here?

rovinski: This is linking the .lib file which should usually come directly from the manufacturer/vendor. Generally it's not expected to be changed by the user, maybe only during a `git pull` (and rarely at that). If it does matter, it can just be a `cp` command instead.

flow/Makefile
```diff
@@ -371,15 +373,19 @@ versions.txt:
...
-$(DONT_USE_LIBS): $$(filter %$$(@F) %$$(@F).gz,$(LIB_FILES))
-	@mkdir -p $(OBJECTS_DIR)/lib
-	$(UTILS_DIR)/markDontUse.py -p "$(DONT_USE_CELLS)" -i $^ -o $@
+$(DECOMPRESSED_LIBS): $$(filter %$$(@F) %$$(@F).gz,$(LIB_FILES))
+	mkdir -p $(OBJECTS_DIR)/lib
+	if [[ $^ == *.gz ]]; then \
+	  gunzip -c $^ > $@; \
+	else \
+	  ln -s $$(realpath --relative-to=$(dir $@) $^) $@; \
+	fi
```
gadfort, line 384: there is one additional issue here.
You still need to use the mark don't use script to ensure the yosys call to map dffs doesn't use any FFs that are marked as don't use.
Mark don't use also does some additional cleanup which has caused ABC to crash before on other PDKs.

In order to actually get rid of this, you will need:
1. yosys support for adding don't use to the dff call in that script (otherwise it's a loss of functionality).
2. abc and yosys some improvements to their liberty parsers to not choke on pdks. https://github.com/siliconcompiler/siliconcompiler/blob/55d8ff05d42f55ac45970641a51f0817a9918b3f/siliconcompiler/tools/yosys/prepareLib.py#L43 https://github.com/siliconcompiler/siliconcompiler/blob/55d8ff05d42f55ac45970641a51f0817a9918b3f/siliconcompiler/tools/yosys/prepareLib.py#L51 https://github.com/siliconcompiler/siliconcompiler/blob/55d8ff05d42f55ac45970641a51f0817a9918b3f/siliconcompiler/tools/yosys/prepareLib.py#L58

flow/scripts/sc/tools/openroad/sc_apr.tcl
gadfort: the sym link is fine here.
Although I'm not sure if any of this code is actually functional at this point

flow/scripts/synth.tcl
```diff
@@ -53,7 +53,7 @@ if {[info exist ::env(LATCH_MAP_FILE)]} {
 if {[info exist ::env(DFF_LIB_FILE)]} {
   dfflibmap -liberty $::env(DFF_LIB_FILE)
 } else {
-  dfflibmap -liberty $::env(DONT_USE_SC_LIB)
+  dfflibmap -liberty $::env(DECOMPRESSED_SC_LIB)
```
gadfort, line 56: needs dont_use

flow/scripts/sc/tools/openroad/sc_apr.tcl
rovinski: Okay, will keep it as is. Thanks.

rovinski:
I opened https://github.com/YosysHQ/yosys/pull/4219 to fix the dont_use issue with `dfflibmap`. It works on my system. @quantamHD maybe you can help review?

I also see an easy path to supporting multiple libs in `dfflibmap`. This would eliminate one half of the need to do lib merging, but as mentioned [above](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1807#issuecomment-1950926389) ABC needs to fix its multiple lib parsing. I'm not sure I want to try that... it looks like a lot of ABC PRs just get stuck.

flow/scripts/synth.tcl
rovinski: I opened https://github.com/YosysHQ/yosys/pull/4219 to fix, but it may take a bit to get approved then upstreamed.

rovinski:
I think for right now because of everything @gadfort mentioned I am going to revert this PR to only adding the `dont_use` flags in the script on top of the current method of marking them in lib. This will be duplicate, but it will be in preparation for eventually removing liberty merging / preprocessing if the upstream is ever fixed.

rovinski:
I've decided to break this PR down into smaller pieces and only merge the ones that make sense.
#1809 covers using the dont_use flags for abc which is originally what #1502 was about.
When https://github.com/YosysHQ/yosys/pull/4219 is merged and then downstreamed to ORFS, then `dfflibmap` can be updated and dont_use modification in .lib can be totally removed. The other library prep steps will have to remain for the time being.

PR #1809, opened 2024-02-20
syn: Use the dont_use flags in the abc pass

rovinski:
This doesn't remove the lib modification in the `markDontUse.py` utility script due to lack of a `dont_use` flag for `dfflibmap` as of right now, among other reasons pointed out in #1807.

This partially replaces #1807.

Future PRs will switch functionality from utility scripts to yosys when support is upstreamed.


rovinski:
This still closes #1502.

maliberty (approved):

PR #1834, opened 2024-03-12
allow empty cells

habibayassin:


vvbandeira:
@maliberty 
For context: there are error messages about empty GDS today in our CI, but only on the new scripted pipeline does it throw an error -- which is the expected behavior as the `def2stream.py` script exits with error code 42 (number of cells with empty GDS).

maliberty:
What is an example design with this error?

habibayassin:
> What is an example design with this error?

aes-hybrid gf180
Example run: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Public%2Fpublic_tests_all-pr/detail/PR-1709-head/161/pipeline/141/

maliberty:
That design is a bit of an odd case and you could put this option in its config.mk.  Ideally we would handle the gds there too but it `Argument list too long` and I haven't thought it worth addressing.

maliberty (approved):

PR #1852, opened 2024-03-18
designs: add rob design

oharboe:
rob - reorder buffer from Chipyard MegaBoom


oharboe:
@maliberty This is an interesting test case to run through synthesis only to verify that the handoff between Yosys and openroad is working. Is there a way to create a synthesis only test-case?

maliberty:
>  Is there a way to create a synthesis only test-case?

No.  

I'm guessing this is related to https://github.com/The-OpenROAD-Project/OpenROAD/issues/4809.  It seems better to formulate that issue as a yosys test (no \$check cells in the final netlist)..

flow/scripts/synth.tcl
```diff
@@ -14,6 +14,8 @@ if { [info exist ::env(SYNTH_GUT)] && $::env(SYNTH_GUT) == 1 } {
 
 # Generic synthesis
 synth -top $::env(DESIGN_NAME) {*}$::env(SYNTH_ARGS)
+# Get rid of indigestibles
+chformal -remove
```
oharboe, line 18: @maliberty Purge indigestibles... Since yosys is sensitivt to initial conditions, even changing comments can affect quality of results, stripping unused information could require updating the metadata that ORFS uses to check for quality of results regressions.

:crossed_fingers:

maliberty: This PR is now a mix of things.  I'm not sure we want to add the design but this command seems useful.

oharboe: Indeed. For now I just wanted to check that nothing else goes wrong and that this design passes floorplan...

PR #1856, opened 2024-03-19
cts: remove unused CTS_CLUSTER_DIAMETER and CTS_CLUSTER_SIZE

oharboe:


maliberty:
They are used by the autotuner so I don't view them as unused.

oharboe:
> They are used by the autotuner so I don't view them as unused.

Autotuner is changing these options, but the options are no longer connected to anything.

maliberty:
@precisionmoon I see you disconnected these options in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/9cb596fd9eb50f5c0c13a1b26ffb25a77289802a - what was the motivation?  The options still seem to exist in the cts code base.

precisionmoon:
These options are still valid.  They are no longer required.  A message is
issued when they are used though about possible sub-optimal solution.

maliberty:
The autotuner might still be able to find a better than default value, no?

precisionmoon:
Yes.

maliberty:
Then I think we should reconnect those values in ORFS even if we undocument them for users [or document them as for ML only].  (@luarss FYI)

PR #1872, opened 2024-03-21
AutoTuner fixes

vvbandeira:


maliberty:
@luarss FYI

luarss:
I think it would be good to fix the version of ray and tensorboard version that worked for reproducibility. I can do this in a future pr.

vvbandeira:
> I think it would be good to fix the version of ray and tensorboard version that worked for reproducibility. I can do this in a future pr.

A couple of hours later they released Ray `3.10.0` which broke `ax` so I merged #1874

PR #1874, opened 2024-03-22
AutoTuner: lock all package versions to avoid breakage

vvbandeira:


vvbandeira:
FYI @maliberty @luarss

PR #1887, opened 2024-03-26
gitignore: update

oharboe:


maliberty:
Is there a reason this test can't write into a results/ dir as other tests do?

oharboe:
> Is there a reason this test can't write into a results/ dir as other tests do?

will fix

PR #1890, opened 2024-03-28
sky130hd-microwatt: increase halos/channels for mpl2 flow & turn off boundary penalty

AcKoucher:
This should unblock either the boundary penalty fix for mpl2 and the gpl routability changes.

### DIFF
--- a/flow/designs/sky130hd/microwatt/config.mk
+++ b/flow/designs/sky130hd/microwatt/config.mk
@@ -22,8 +22,11 @@ export ADDITIONAL_LEFS  = $(wildcard $(microwatt_DIR)/lef/*.lef)
 export ADDITIONAL_LIBS = $(wildcard $(microwatt_DIR)/lib/*.lib)
 
 export SYNTH_HIERARCHICAL = 1
-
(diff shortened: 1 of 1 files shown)


AcKoucher:
Running locally with master showed no metrics failures.

eder-matheus:
@maliberty Arthur made this change because the halos were too small, considering the size of the die. I believe microwatt was on the edge of failing during the global routing with high congestion. Changing the halos creates more space for the routing near the macros. Note that the previous halo value was a default value for all designs in the PDK.

@AcKoucher, could you paste a figure showing the new macro placement after this change?

maliberty:
I restarted pr-merge as it looks like a network failure

AcKoucher:
image

AcKoucher:
I opened [#4867](https://github.com/The-OpenROAD-Project/OpenROAD/issues/4867) to investigate how to improve results and run time for this design. As discussed in R&D synch, it's a challenging case for mpl2 due to large macros, pins in all boundaries, few metal layers...

eder-matheus (approved):

PR #1897, opened 2024-04-01
sky130hd-microwatt: design specific configs to allow mpl2 and gpl changes

AcKoucher:
New attempt to to do what was done in #1890 and reverted  in #1891  due to antenna repair crash.
1. Increase the halos
2. Set mpl2 SA boundary penalty to zero

Now we're manually (and temporarily) skipping antenna repair to avoid the crash.

Also using the latest OR to avoid some other failing after merging.


eder-matheus:
@maliberty I think temporarily disabling repair_antennas here makes sense since the repair_antennas crash is holding two different updates. I hope we'll have the new antenna checker in the next one or two weeks, so this design will have the repair disabled for a short period.

Also, I intend to add a CI check for antenna violations (see https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1899) after having the new antenna checker.

maliberty (approved):

PR #1911, opened 2024-04-03
add target clock latency metric to reports

tspyrou:
This can be used to get the top level target clock latency automatically for use in blocks.
The next step would be to figure out a way to push this value down into the mock element for a future run.


flow/scripts/report_metrics.tcl
```diff
@@ -152,6 +152,23 @@ proc report_metrics { stage when {include_erc true} {include_clock_skew true} }
       set path_delay -1
       set path_slack 0
     }
+
+    set reg_to_reg_critical_path [lindex [find_timing_paths -from [all_registers] -to [all_registers]] 0]
+    if {$reg_to_reg_critical_path != ""} {
+      set path_delay [sta::format_time [[$reg_to_reg_critical_path path] arrival] 4]
+      set path_slack [sta::format_time [[$reg_to_reg_critical_path path] slack] 4]
```
maliberty: Why it is overriding the values computed above?

maliberty:
Is this something that is supposed to be a metric in the json or just a report?

flow/scripts/report_metrics.tcl
tspyrou: I just fixed that, it was a paste error.

tspyrou:
> Is this something that is supposed to be a metric in the json or just a report?

It depends on the use model for @oharboe. I don't think this is a generally useful metric but it is useful for pushing the value into sub blocks. Ideally I would enhance this to create a value for each possible target clock in a multi clock design. I thought I would start with this so we can work out the methodology. That enhancement is trivial to implement. If the clock name is in the metric I doubt we would want it in the json. We could put max latency across or clocks. This code will only give the latency of the critical path's target clock.

tspyrou:
@maliberty This is a tiny subset of the ability to push an SDC down to a sub block instance.

oharboe (changes requested):
From reading the script I believe "target clock latency" gets the longest clock insertion latency for a flip flop at the mock-array level. I'm not sure what mock-array/Element should do with this information: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/33d43e861a4971fc4ffc1f8102f6871b1a32edf8/flow/designs/asap7/mock-array/Element/constraints.sdc#L21

I'm a bit unclear on what is being calculated and what mock-array/Element should do with this information. Can you update mock-array/Element to show what should happen in mock-array/Element?

Once what is going on here is clarified, I have an idea on how to wrap this PR up. It is a step in the right direction and valuable in and of itself, even if there is room for future improvement(automatically pushing information down to mock-array/Element):



Update the reporting to list all macros with actual values for the mock-array/Element level and the recommended values for mock-array/Element/constraints.sdc.

The user can then copy and paste these values into the constraints.sdc files for the macros.

Add an environment variable MACRO_INSERTION_LATENCY_TOLERANCE, default "inf" or 1e+30. If the macro insertion latency for the macros are out of acceptable range, fail report_metrics with an error.

This sets up a second best situation to automatic propagation down to mock-array/Element of this information: automatically checked + actionable error, but manually updated.

tspyrou:
I am not sure where the 70 came from for the virtual clock. 

This latency is focused on the latency for the propagate clock. You would add a line to the SDC for element with
set_clock_latency $value_from_mock_array [get_clocks clock]

Startpoint: ces_6_1 (rising edge-triggered flip-flop clocked by clock)
Endpoint: ces_6_5 (rising edge-triggered flip-flop clocked by clock)
Path Group: clock
Path Type: max

oharboe:
> I am not sure where the 70 came from for the virtual clock.
> 
> This latency is focused on the latency for the propagate clock. You would add a line to the SDC for element with set_clock_latency $value_from_mock_array [get_clocks clock]
> 
> Startpoint: ces_6_1 (rising edge-triggered flip-flop clocked by clock) Endpoint: ces_6_5 (rising edge-triggered flip-flop clocked by clock) Path Group: clock Path Type: max

70 is the clock insertion latency within the mock-array/Element.

rovinski:
@tspyrou I'm not sure if the clock latency is useful information for pushing into the blocks? When doing timing allocation I would usually want to work bottom-up from the blocks to the top-level. This seems orthogonal to the solution requested in https://github.com/The-OpenROAD-Project/OpenROAD/issues/4655

maliberty:
I don't see a consensus that this is useful.

tspyrou:
> @tspyrou I'm not sure if the clock latency is useful information for pushing into the blocks? When doing timing allocation I would usually want to work bottom-up from the blocks to the top-level. This seems orthogonal to the solution requested in [The-OpenROAD-Project/OpenROAD#4655](https://github.com/The-OpenROAD-Project/OpenROAD/issues/4655)

@rovisnki I think the request from @oharboe is to automate the pushing of the clock latency from the top level into the element so that the element level run will not add hold buffers which are not needed in the contest of the incoming clock from the top. I do see that the hold closes more easily with the new SDC as measured by the post global routing repair timing report.

tspyrou:
@oharboe @rovinski I updated the report_metrics script to report the critical setup and the critical hold path's clock latency. We could find the worst latency across all paths but that would be runtime prohibitive.

tspyrou:
image

PR #1921, opened 2024-04-09
[Autotuner] Working params for asap7

luarss:
@vvbandeira Committing some changes to `autotuner.json` for the following designs in asap7.

- jpeg, jpeg_lvt
- ibex
- aes
- aes-block
- mock-array
- mock-alu
- gcd

Key parameters I adjusted/removed
- `EQUIVALENCE_CHECK`: eqy bug
- `_FR_LAYER_ADJUST`: due to routing failures 
- `_PINS_DISTANCE`: due to pin placement failures.
- `CORE_UTILISATION`: the upper limit is either equal to golden or somewhere near.


flow/scripts/cts.tcl
```diff
@@ -79,14 +79,14 @@ append_env_var additional_args SKIP_PIN_SWAP -skip_pin_swap 0
...
 if {[info exists ::env(SKIP_CTS_REPAIR_TIMING)] == 0 || $::env(SKIP_CTS_REPAIR_TIMING) == 0} {
-  if { [info exists ::env(EQUIVALENCE_CHECK)] } {
+  if {[info exists ::env(EQUIVALENCE_CHECK)] && $::env(EQUIVALENCE_CHECK) == 1} {
       write_eqy_verilog 4_before_rsz.v
   }
 
   puts "repair_timing [join $additional_args " "]"
   repair_timing {*}$additional_args
 
-  if { [info exists ::env(EQUIVALENCE_CHECK)] } {
```
vvbandeira, line 90: Instead of modifying the `cts.tcl` which applies to all designs including the ones outside of AT, you could modify the `distributed.py` by adding `EQUIVALENCE_CHECK=0` to the `make_command` in the `openroad` function, for example.

vvbandeira (changes requested):
As we talked about in the meeting. Removing the `_PINS`, `_FR`, and `_SDC` variables is fine for now, but they should be present in at least one CI testcase later.

flow/scripts/cts.tcl
luarss: So `EQUIVALENCE_CHECK` is meant more as a flag rather than a switch?

vvbandeira: If you are using a definition where "flag" is true/false and "switch" can have multiple options, then yes it is a flag. If that is not what you meant, please let me know.

luarss: I am just clarifying on the use of `EQUIVALENCE_CHECK` because I see in `config.mk` files they are defined as `export ... ?= 1`, but in the code `cts.tcl` we just check for the existence of such a variable, so it seems that the code treats it more like a flag.

luarss: Code checks for existence of `EQUIVALENCE_CHECK`, so even if it is set to zero I think the logic should still execute?

maliberty: As it is off by default I don't think we need to worry about explicitly turning it off.

vvbandeira: > I am just clarifying on the use of `EQUIVALENCE_CHECK` because I see in `config.mk` files they are defined as `export ... ?= 1`, but in the code `cts.tcl` we just check for the existence of such a variable, so it seems that the code treats it more like a flag.

Apologies, I think I confused the diffs. I think is ok to add the check `== 1`.



> As it is off by default I don't think we need to worry about explicitly turning it off.

Jack is facing an issue with one design that has `eqy` check enabled, but is failing a lot during AT. I suggested to disable `eqy` for AT runs while we investigate the issue.

flow/designs/asap7/jpeg/autotuner.json
```diff
@@ -8,21 +8,21 @@
...
         "type": "int",
         "minmax": [
-            30,
-            100
+            0,
+            0
         ],
-        "step": 1
+        "step": 0
     },
```
vvbandeira: Let's use the approach I mentioned in the other comment instead of adding this to the search space. Adding an entry here would increase the number of dimensions in the search space -- which is suboptimal.

vvbandeira (changes requested):

flow/designs/asap7/jpeg/autotuner.json
maliberty: There is no need to auto tune the use of eqy as it doesn't improve the results, it only verifies correctness.  I suggest disabling it when autotuning.

luarss: Thanks for the clarification @vvbandeira @maliberty. Updated accordingly

vvbandeira (approved):

luarss:
@vvbandeira I recently found that _FR_LAYER_ADJUST can be made a small value for flow completion. Let me adjust it before merging.

vvbandeira:
> @vvbandeira I recently found that _FR_LAYER_ADJUST can be made a small value for flow completion. Let me adjust it before merging.

@luarss, do you have an ETA for the update?

vvbandeira (approved):

PR #1937, opened 2024-04-14
gha: add rules table as pr body

habibayassin:


.github/workflows/github-actions-update-rules.yml
```diff
@@ -1,5 +1,6 @@
 name: Create draft PR for updated rules
 on:
+  push:
```
vvbandeira, line 3: Was this for testing only or should it be kept?

vvbandeira: @habibayassin ping

flow/util/updateRules.py
```diff
@@ -43,7 +43,7 @@
     design = dirList[2]
     test = '{} {}'.format(platform, design)
     dataFile = os.path.join(designsDir, runFilename)
-    if os.path.exists(dataFile) and (platform != 'sky130hd_fakestack' or platform != 'src'):
+    if os.path.exists(dataFile) and (platform != 'sky130hd_fakestack' and platform != 'src' and platform not in ["intel22", "intel16", "gf12", "gf55", "tsmc65lp"]):
```
vvbandeira:
```suggestion
    platform_list = []
    platforms_path = 'platforms'
    for item in os.listdir():
      item_path = os.path.join(platforms_path, item)
      if os.path.isdir(item_path):
        platform_list.append(item)
    if os.path.exists(dataFile) and platform in platform_list:
```

vvbandeira: @habibayassin 
Please test the suggestion. This way is better so we do not depend on hardcoded strings for the platform names.

flow/util/updateRules.py
```diff
@@ -34,6 +34,15 @@
 
 runFilename = f'rules-base.json'
 
+platform_list = []
+platforms_path = 'platforms'
+for item in os.listdir():
```
vvbandeira:
```suggestion
platforms_path = 'platforms'
for item in os.listdir(platforms_path):
```

flow/util/updateRules.py
```diff
@@ -34,6 +34,15 @@
...
+    if item == 'sky130hd_fakestack':
+        continue
+    item_path = os.path.join(platforms_path, item)
+    if os.path.isdir(item_path):
```
vvbandeira:
```suggestion
    if os.path.isdir(os.path.join(platforms_path, item)):
```

flow/util/genRuleFile.py
```diff
@@ -311,9 +315,9 @@ def gen_rule_file(design_dir, update, tighten, failing, variant, golden_metrics=
     parser.add_argument('dir', help='Path to the design directory.')
     parser.add_argument(
         '-v',
```
vvbandeira:
```suggestion
        '-p',
```

.github/workflows/github-actions-update-rules.yml
vvbandeira: Remove and we can move with the merge.

vvbandeira (changes requested):

flow/util/genRuleFile.py
```diff
@@ -328,9 +332,9 @@ def gen_rule_file(design_dir, update, tighten, failing, variant, golden_metrics=
...
-        '--variant',
+        '--platform',
         default='base',
-        help='Flow variant [default="base"].')
+        help='Flow platform [default="asap7"].')
```
vvbandeira: I thought these changes had been reverted. Why are we removing `variant` and adding `platform` argument? There are other places where we use this script and it still uses the `variant` argument.

vvbandeira (changes requested):

PR #1945, opened 2024-04-17
sg13g2: Use decoupling cells as fill cells

hpretl:
Currently, only the fill cells `sg13g2_fill_[1|2|4|8]` are used. Instead of `sg13g2_fill_[4|8]` the supply decouling cells `sg13g2_decap_]4|8]` shall be used.

### DIFF
--- a/flow/platforms/ihp-sg13g2/config.mk
+++ b/flow/platforms/ihp-sg13g2/config.mk
@@ -27,7 +27,7 @@ sg13g2_dfrbp_2
 
 
 # Define fill cells
-export FILL_CELLS = sg13g2_fill_1 sg13g2_fill_2 sg13g2_fill_4 sg13g2_fill_8
+export FILL_CELLS = sg13g2_fill_1 sg13g2_fill_2 sg13g2_decap_4 sg13g2_decap_8
 # -----------------------------------------------------
 #  Yosys
 #  ----------------------------------------------------


maliberty (approved):

maliberty:
@KrzysztofHerman any concerns?

rovinski:
This is generally not preferable because decap cells are leaky. You should only be adding decap as needed, not adding it by default, because it can increase leakage power unnecessarily. OpenROAD unfortunately doesn't have a decap insertion function at this point. If it did, that would be the place to add decap.

maliberty:
Perhaps it should be a config option?

@rovinski it shouldn't be too hard to add such an step.  Do you have a sense of what user controls are needed?

rovinski:
@maliberty usually decaps should be added after IR drop analysis and inserted next to instances with the worst IR drop. A loose spec off the top of my head would be:

* list of decap cells to use
* maximum amount of decap to insert (in liberty cap units), or
* voltage margin to fix up to
  * e.g. if worst IR drop is 0.15 V, and the user specifies 0.10 V, the tool would try to insert decap around all failing instances to ensure the worst IR drop is no more than 0.10 V

maliberty:
@rovinski the last choice would require a dynamic IR drop calculation.  Currently we only look at static IR so we would have to look at the max cap option.  Is that useful?

hpretl:
Good point about the increased leakage, definitely true for a large digital chip. I come from the mixed-signal world where you often want to maximize decap for crosstalk reason. 

Maybe a user setting controlling the percentage of fill w/o decap and fill w/ decap would be a simple first step?

rovinski:
> @rovinski the last choice would require a dynamic IR drop calculation. Currently we only look at static IR so we would have to look at the max cap option. Is that useful?

Yes, that is actually the status quo of how the proprietary solutions do it, although it's a bit tedious from the designer standpoint. It requires a bit of trial and error because the designer doesn't generally know how many fF of cap is required to fix an IR issue.

I would definitely take a simple solution over no solution. Static IR drop hotspots correlate highly with dynamic IR hotspots. If the routine could add a user-specified amount of cap around the instances with the largest static IR drop, that would be a very good first-cut solution.

rovinski:
Added https://github.com/The-OpenROAD-Project/OpenROAD/issues/4970 to track

maliberty:
For now I'll merge this as leakage is less serious than IR drop related failure.  When we have https://github.com/The-OpenROAD-Project/OpenROAD/issues/4970 we can revisit

PR #1947, opened 2024-04-18
flow: slight global_place_skip_io skipping logic if there are io constraints

oharboe:


oharboe:
@maliberty Ready for review. I found a lot of mistakes in how HAS_IO_CONSTRAINTS was specified, copy & paste?

I think the only robust solution is to get rid of HAS_IO_CONSTRAINTS altogether and have the openroad binary check if there are any io constraints and if so, skip.

Can openroad be modified to get rid of reliance on the config files here and figure out if it should skip this step or not?

oharboe:
@maliberty Had to update_ok two cases that had broken HAS_IO_CONSTRAINTS check.

I suppose you have to run this on private CI and probably update_ok a few broken cases there too...

oharboe:
@maliberty Good to go for my part.

maliberty (approved):

PR #1965, opened 2024-04-23
mock-array: Element now has register to register paths

oharboe:
Investigate build times and results.


oharboe:
@maliberty @tspyrou Some register to register paths within mock-array/Element. Next I will look at adding an asynchronous FIFO to the mock-array level.

flow/designs/src/mock-array/src/main/scala/MockArray.scala
```diff
@@ -66,8 +66,9 @@ class MockArray(width: Int, height: Int, singleElementWidth: Int)
     // Registered routing paths
     //  left <-> down
     //  up <-> right
-    (io.outs.asSeq zip io.ins.asSeq.reverse.map(RegNext(_))).foreach {
-      case (a, b) => a := b
+    (io.outs.asSeq zip (io.ins.asSeq ++ Seq(io.ins.asSeq.head))
+      .sliding(2).toSeq.reverse.map(_.map(RegNext(_)))).foreach {
+      case (a, b) => a := RegNext(b(0) + b(1))
```
maliberty, line 71: What does that say?  Its not very readable to me.

maliberty (approved):

flow/designs/src/mock-array/src/main/scala/MockArray.scala
oharboe: It's quite readable, if you are used to functional programming :-) 

it puts inputs into registers, sums up pairs of inputs and puts the result into a register and assigns the output of the registers to outputs.

the point is: there is some non-trivial amount of register  output -> fog of logic -> register paths, which is useful when studying constraints.sdc files.

PR #1967, opened 2024-04-23
mock-array:  use CTS_BUF_DISTANCE to make mock-array less brittle

oharboe:
@precisionmoon @maliberty CTS_ARGS had to be used while -insertion_delay was introduced, but now CTS_ARGS is no longer needed as CTS_BUF_DISTANCE alone adequately changes options.

With this change, mock-array is less brittle when CTS and CTS_ARGS is updated.


maliberty:
Please resolve the conflicts

oharboe:
@maliberty fixed. Did a force push since it is a single line change.

PR #1968, opened 2024-04-24
tests: simpler to maintain test_genElapsedTime.py

oharboe:
python knows how to diff lists and present them nicely. This avoids exactly replicating the number of tabs, which means formatting can be changed without updating tests. Better.


oharboe:
@maliberty Some unrelated CI error

PR #1970, opened 2024-04-25
asap7: less repetition

oharboe:


oharboe:
@maliberty unrelated microwatt DRC errors

maliberty (approved):

PR #1977, opened 2024-04-26
platforms: introduce a platforms/common folder

oharboe:
these are files that could be platform specific, but are not today.


oharboe:
@maliberty Don't know what to do about the security error in CI:

```
ERROR: File name is blocked: flow/platforms/common/lcu_kogge_stone.v
```

rovinski:
`src` is already a common folder. Not sure why this is necessary? If it becomes platform-specific, then it could be moved to that specific platform.

oharboe:
> `src` is already a common folder. Not sure why this is necessary? If it becomes platform-specific, then it could be moved to that specific platform.

I view the designs/ as distinct from the ORFS build system(Makefile, scripts and platforms), it is part of the test cases, not part of what I need to use ORFS for my own project.

maliberty (approved):

PR #1978, opened 2024-04-26
[DO NOT MERGE] scripts: add a histogram to show endpoint slack for register to register paths

oharboe:
Fetches data using .tcl and plots using Python matplotlib

source scripts/histogram.tcl

### DIFF
--- a/flow/scripts/histogram.py
+++ b/flow/scripts/histogram.py
@@ -1,0 +1,26 @@
+import matplotlib.pyplot as plt
+import sys
+import os
+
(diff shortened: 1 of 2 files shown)


oharboe:
@maliberty Is there a better way to do this? Can I run python scripts from within the GUI when it is in tcl mode?

maliberty:
I don't see a better way at this point

oharboe:
> I don't see a better way at this point

Good to know at least.

PR #1985, opened 2024-05-01
platforms: use path relative to the FLOW_HOME

vvbandeira:
For private PDKs we use a different PLATFORM_HOME which does not contain the "common" folder.


vvbandeira:
@oharboe FYI in case it has some side effects on your setup.

oharboe (approved):

PR #1986, opened 2024-05-01
riscv32i-mock-sram: use PDN_TCL that we generally use in a BLOCKS scenario

oharboe:


oharboe:
@maliberty It passed tests, but "continuous-integration/jenkins/pr-merge" is stuck. Are the build servers down? Backlog?

maliberty (approved):

PR #1991, opened 2024-05-05
makefile: reduce surprises, NUM_CORES can now be specified on command line

oharboe:
`NUM_CORES=1234  make print-NUM_CORES` now prints 1234

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -243,7 +243,7 @@ export RTLMP_BLOCKAGE_FILE ?= $(OBJECTS_DIR)/rtlmp/partition.txt.blockage
 #-------------------------------------------------------------------------------
 ifeq (,$(strip $(NUM_CORES)))
   # Linux (utility program)
-  NUM_CORES := $(shell NUM_CORES 2>/dev/null)
+  NUM_CORES := $(shell nproc 2>/dev/null)
 
   ifeq (,$(strip $(NUM_CORES)))
     # Linux (generic)
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -241,26 +241,24 @@ export RTLMP_RPT_FILE ?= partition.txt
 export RTLMP_BLOCKAGE_FILE ?= $(OBJECTS_DIR)/rtlmp/partition.txt.blockage
 
 #-------------------------------------------------------------------------------
-ifeq (, $(strip $(NPROC)))
+ifeq (,$(strip $(NUM_CORES)))
   # Linux (utility program)
-  NPROC := $(shell nproc 2>/dev/null)
+  NUM_CORES := $(shell NUM_CORES 2>/dev/null)
 
-  ifeq (, $(strip $(NPROC)))
+  ifeq (,$(strip $(NUM_CORES)))
     # Linux (generic)
-    NPROC := $(shell grep -c ^processor /proc/cpuinfo 2>/dev/null)
+    NUM_CORES := $(shell grep -c ^processor /proc/cpuinfo 2>/dev/null)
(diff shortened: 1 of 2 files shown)


maliberty:
You could already specify NPROC so now there are two different ways to do the same thing.

oharboe:
> You could already specify NPROC so now there are two different ways to do the same thing.

Yes and the second didn't behave like the first, now it does.

Should one be retired?

oharboe:
Having thought about it: NPROC is the number of threads/CPUs in the system, NUM_CORES is the number of cores ORFS should use.

These are distinct concerns, so I think both variables should be kept.

maliberty:
The only purpose of nproc is to set num cores so it isn't really useful by itself.

maliberty:
I think these should just be collapsed into one var

oharboe:
> I think these should just be collapsed into one var

If I want to use the number of available cores in some reporting or policies, both could be useful, would be my counterargument.

oharboe:
I think the question of whether to collapse these variables is out of scope for this PR.

maliberty:
> I think the question of whether to collapse these variables is out of scope for this PR.

I think have duplicate settable variables is confusing so I think the question is relevant.

> If I want to use the number of available cores in some reporting or policies, both could be useful, would be my counterargument.

We don't do that.  If we did, having NUM_CORES be the number of running threads wouldn't match your idea.  At a minimum things would need renaming and only one should be user settable.

oharboe:
I am happy to eliminate NPROC, will do in this PR 
these things are easily changed when needed.

oharboe:
Seems to work:

```
oyvind@small-cigar:~/OpenROAD-flow-scripts/flow$ make print-NUM_CORES
NUM_CORES = 20
oyvind@small-cigar:~/OpenROAD-flow-scripts/flow$ make NUM_CORES=10 print-NUM_CORES
NUM_CORES = 10
oyvind@small-cigar:~/OpenROAD-flow-scripts/flow$ NUM_CORES=10 make print-NUM_CORES
NUM_CORES = 10
oyvind@small-cigar:~/OpenROAD-flow-scripts/flow$ echo >>settings.mk export NUM_CORES=10
oyvind@small-cigar:~/OpenROAD-flow-scripts/flow$ make print-NUM_CORES
NUM_CORES = 10
oyvind@small-cigar:~/OpenROAD-flow-scripts/flow$ rm settings.mk 
oyvind@small-cigar:~/OpenROAD-flow-scripts/flow$ make print-NUM_CORES
NUM_CORES = 20
```

maliberty (approved):

flow/Makefile
```diff
@@ -241,26 +241,24 @@ export RTLMP_RPT_FILE ?= partition.txt
 export RTLMP_BLOCKAGE_FILE ?= $(OBJECTS_DIR)/rtlmp/partition.txt.blockage
 
 #-------------------------------------------------------------------------------
-ifeq (, $(strip $(NPROC)))
+ifeq (,$(strip $(NUM_CORES)))
   # Linux (utility program)
-  NPROC := $(shell nproc 2>/dev/null)
+  NUM_CORES := $(shell NUM_CORES 2>/dev/null)
```
rovinski: `nproc` is a bin util. It can't be replaced here.

oharboe: Oops. Well spotted. Fixed.

oharboe:
@maliberty Should be good to go now, fixed gaffe spotted by @gadfort

maliberty:
I think you mean spotted by @rovinski

PR #1999, opened 2024-05-09
util: makeIssue.sh: Use DESIGN_CONFIG env var for passing the path to the config

lpawelcz:
`makeIssue.sh` specifies the path to the design config as `$DESIGN_DIR/config.mk`.
This will be incorrect for designs which use config that has name different than `config.mk`. We can use `DESIGN_CONFIG` env var to not have to make assumptions about the config file name.


maliberty (approved):
Thanks

PR #2006, opened 2024-05-14
Makefile creature comforts

oharboe:


maliberty:
Would you give an example of what this enables?

oharboe:
> Would you give an example of what this enables?

See commit comments for example use cases.

maliberty (approved):

PR #2007, opened 2024-05-14
scripts: introduce log_cmd utility for consistency and less reptition

oharboe:


maliberty (approved):
https://github.com/The-OpenROAD-Project/OpenROAD/issues/3539 is a better solution but this will do until then.

PR #2013, opened 2024-05-16
Fix `ABC contains local modifications` error when building Docker image

eszpotanski:
This PR fixes an errors appearing when OpenROAD is built, first locally and then in Docker image.

It can be reproduced on Ubuntu 23.04 with commands:
```
sudo ./setup.sh
./build_openroad.sh --local --latest --clean-force && ./build_openroad.sh --latest --clean-force
```

Yosys uses `git diff-index` to detect changes in ABC, which did not update indexes before check ([code](https://github.com/The-OpenROAD-Project/yosys/blob/a1bb0255d654ffd0d8503577274f24b349a42995/Makefile#L798)). Received error:
```
> [4/4] RUN ./build_openroad.sh --no_init --local --threads 16:
78.96 [ 88%] Building techlibs/quicklogic/ql_dsp_macc_pm.h
79.02 [ 89%] Building techlibs/sf2/synth_sf2.o
79.02 [ 89%] Building techlibs/xilinx/synth_xilinx.o
79.27 [ 89%] Building techlibs/xilinx/xilinx_dffopt.o
79.35 [ 99%] Building yosys-config
79.35 [ 99%] Building abc/abc-0cd90d0
79.38 ERROR: ABC contains local modifications! Set ABCREV=default in Yosys Makefile!
79.38 make: *** [Makefile:795: abc/abc-0cd90d0] Error 1
79.38 make: *** Waiting for unfinished jobs....
82.53 make: Leaving directory '/OpenROAD-flow-scripts/tools/yosys'
------
Dockerfile.builder:16
--------------------
  14 |     # RUN cat tools/yosys/Makefile | grep ABCREV
  15 |     # RUN cd tools/yosys/abc; git status; cd -
  16 | >>> RUN ./build_openroad.sh --no_init --local --threads ${numThreads}
  17 |
--------------------
ERROR: failed to solve: process "/bin/sh -c ./build_openroad.sh --no_init --local --threads ${numThreads}" did not complete successfully: exit code: 2
```
Fixed by adding `git update-index --refresh` before building Yosys.

### DIFF
--- a/build_openroad.sh
+++ b/build_openroad.sh
@@ -273,6 +273,12 @@ __local_build()
             set -u
         fi
 
+        YOSYS_ABC_PATH=tools/yosys/abc
+        if [[ -d "${YOSYS_ABC_PATH}/.git" ]]; then
+            # update indexes to make sure git diff-index uses correct data
+            git --work-tree=${YOSYS_ABC_PATH} --git-dir=${YOSYS_ABC_PATH}/.git update-index --refresh
+        fi
+
         echo "[INFO FLW-0017] Compiling Yosys."
         ${NICE} make install -C tools/yosys -j "${PROC}" ${YOSYS_ARGS}
 


oharboe:
@eszpotanski To reproduce:

```
./build_openroad.sh --local --latest --openroad-args -DCMAKE_BUILD_TYPE=DEBUG
```

Then:

```
./build_openroad.sh --local --latest
```

gives:

```
CMake Error: Error: generator : Ninja
Does not match the generator used previously: Unix Makefiles
Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.

CMake Error at /home/oyvind/ORFS2/dependencies/share/cmake-3.24/Modules/FetchContent.cmake:1592 (message):
  CMake step for googletest failed: 1
  /home/oyvind/ORFS2/dependencies/share/cmake-3.24/Modules/FetchContent.cmake:1744:EVAL:2 (__FetchContent_directPopulate)
  /home/oyvind/ORFS2/dependencies/share/cmake-3.24/Modules/FetchContent.cmake:1744 (cmake_language)
  /home/oyvind/ORFS2/dependencies/share/cmake-3.24/Modules/FetchContent.cmake:1958 (FetchContent_Populate)
  third-party/gtest/CMakeLists.txt:8 (FetchContent_MakeAvailable)


-- Configuring incomplete, errors occurred!
See also "/home/oyvind/ORFS2/tools/OpenROAD/build/CMakeFiles/CMakeOutput.log".
See also "/home/oyvind/ORFS2/tools/OpenROAD/build/CMakeFiles/CMakeError.log".
```

build_openroad.sh
```diff
@@ -273,6 +273,11 @@ __local_build()
             set -u
         fi
 
+        if [[ -d "tools/yosys/abc/.git" ]]; then
+            # update indexes to make sure git diff-index uses correct data
+            git -C tools/yosys/abc update-index --refresh
```
oharboe: ancient versions of git doesn't have the -C option, so this would be a problem on CentOS 7. 

@maliberty Do you care about CentOS 7 anymore?

maliberty: We are in the process of leaving centos7 behind but have no fully gotten there yet.

maliberty:
Wouldn't it be better to upstream the yosys fix rather than work around it here?

.dockerignore
```diff
@@ -31,3 +31,6 @@ core*
 # git
 .git
 .gitignore
+
+# local dependencies
+dependencies/
```
maliberty: Where does this come from in the build process?

eszpotanski: `dependencies` directory is created by [setup.sh](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/setup.sh#L20)

eszpotanski:
Master branch of Yosys already have this issue fixed, whole ABC handling was refactored and now it is a proper submodule checked with `git submodule status` ([code](https://github.com/The-OpenROAD-Project/yosys/blob/master/Makefile#L790)), so this `git update-index --refresh` is kind of temporary solution until next Yosys update in ORFS.

Also, this `ABC contains local modifications!` error appears in ORFS, as it caused by coping files to Docker, where we end up with technically different files (with the same content) then already indexed ones, Yosys alone works fine.

@maliberty if it is not convincing for you, I will prepare changes in Yosys.

oharboe:
@eszpotanski How about breaking out one concern per PR? That way the easy fixes can go in and the rest can be worked on separately.

eszpotanski:
@oharboe Sure, I created https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2015 with `dependencies/` in .dockerignore and I'll edit this one to contain only fix for `git diff-index`

maliberty (approved):
I still feel this should be fixed in the upstream but will accept it until that can happen.

maliberty:
> Master branch of Yosys already have this issue fixed

In which commit is this fixed upstream?  We are on 0.40 and in process to update to 0.41.

maliberty:
The title of this PR is out of date

eszpotanski:
https://github.com/The-OpenROAD-Project/yosys/commit/b37e8d52ec5f8b5886aa3da8508c7980263024ec fixes it by introducing ABC as submodule and removing `git diff-index` usage.

Title had been changed to: Fix `ABC contains local modifications` error when building Docker image

build_openroad.sh
eszpotanski: `-C ...` changed to `--work-tree` and `--git-dir`

PR #2018, opened 2024-05-18
mpl2: MACRO_PLACE_HALO x and y are now separate

oharboe:
same as for classic macro placement.


oharboe:
@AcKoucher @maliberty FYI

maliberty (approved):

PR #2020, opened 2024-05-19
[DO NOT MERGE] Mock array mpl2

oharboe:
@AcKoucher Gave it a go, but it fails in global routing, DRC report shown.

Just a thought: it should be possible to automatically detect which macros can be routed by abutment and in which orientation. The pins and nets have to line up....

image


AcKoucher:
@maliberty After some debugging it looks like there's another function from the original implementation - `SACoreSoftMacro::alignMacroClusters()` - that, each time SA for a given outline is done, it tries to push the macros in the direction of that current outline's boundaries based on some threshold. I need some time to understand if that is actually good for eliminating notches (which is, based on the code comments, the intentions of this function). 

@oharboe Either way, I don't think Mock-array need to wait for this investigation. There's a temporary change that can be done.

oharboe:
@AcKoucher No rush on getting mock-array working with mpl2. I would rather wait for a real fix that to hack config.mk files and options. The fewer options tweaks the better. 

If many mysteriou options must be fine tuned, then that undermines the "automatic" part of mpl2, it becomes little better than manual placement.

AcKoucher:
@oharboe Thanks for the clarification. In this case, it wouldn't be a user config thing so I went ahead and opened [#5140](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5140).

oharboe:
@AcKoucher Works w/latest master.

I didn't study the change in metadata results, the floorplan looks good to me, so I suspect just some sort of change in initial conditions for other stages.

```
[WARN] Overwriting Tag finish__runtime__total
[WARN] Overwriting Tag finish__cpu__total
[WARN] Overwriting Tag finish__mem__peak
[INFO] synth__design__instance__area__stdcell pass test: 15646.73174 <= 18363.81
[INFO] constraints__clocks__count pass test: 1.0 == 1.0
[INFO] placeopt__design__instance__area pass test: 119959.0 <= 137948.0
[INFO] placeopt__design__instance__count__stdcell pass test: 10691.0 <= 12335.0
[INFO] detailedplace__design__violations pass test: 0.0 == 0.0
[INFO] cts__design__instance__count__setup_buffer pass test: 0.0 <= 1073.0
[INFO] cts__design__instance__count__hold_buffer pass test: 0.0 <= 1073.0
[ERROR] detailedroute__route__wirelength fail test: 76149.0 <= 73578.0
[INFO] detailedroute__route__drc_errors pass test: 0.0 <= 0.0
[ERROR] finish__timing__setup__ws fail test: -78.1376 >= -6.93
[INFO] finish__design__instance__area pass test: 119974.0 <= 137966.0
[INFO] finish__timing__drv__setup_violation_count pass test: 28.0 <= 536.0
[INFO] finish__timing__drv__hold_violation_count pass test: 0.0 <= 100.0
[ERROR] finish__timing__wns_percent_delay fail test: -49.411146 >= -10.0
Failed metadata checks: 3 out of 14
```

oharboe:
A quick look: the io_lsb_* group lower right is placed at the bottom instead of maybe in the middle?

image

oharboe:
Magic option, seems to be required still or the macros are not centered:

```
-export RTLMP_BOUNDARY_WT=0
```

Results in this placement, which will cause too much congestion on the left, I believe.

image

AcKoucher:
@oharboe Setting the boundary weight to zero, for the current approach, will still be needed.

The position of the pin groups would probably be related to PPL.

oharboe:
@AcKoucher Interesting. io_lsbs_0 is an output that is driven by a flipflop. The input to that flipflop comes from 


image

The input to that flipflop comes from the upper right corner of the design :exploding_head: 

image

maliberty:
It looks like there isn't enough room on the side to put that ff where it wants to do .

oharboe:
> It looks like there isn't enough room on the side to put that ff where it wants to do .

Which makes me worry that the mock-array test case is broken in some sense.

These flip flops, by design should be immediately right of the rightmost Element in each row with the io_lsbs_*  pins for each row to the right of the flip flops...

Merits further study....

oharboe:
@maliberty @AcKoucher On master, things look as expected. The io_lsbs_* are centered on the middle of the right side and the flip flops are on the right side of the array between the io_lsbs_* pins.


image

AcKoucher:
Mpl2 does generate a temporary location for the std cells in the lower left corner due to way that the sequence pair representation works.

image

However afaik this temporary location is not used.

oharboe:
I'm unsure what you mean, so maybe I'm talking past you here. I ran this PR through the full flow when I identified the flip flops in the lower [right corner](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2020#issuecomment-2121167498).

oharboe:
Follow the rest of these developments in #2059

PR #2025, opened 2024-05-22
[Autotuner] Installation/ Sanity checks

luarss:
- Make etc/DependencyInstaller.sh install Autotuner by default.
- Runtime sanity check. If the parameter is not a flow variable, raise a Warning <- should this be an error?

@vvbandeira


oharboe:
@luarss @maliberty Does this lock down the versions required in the system install for Python?

That would be very bad as it is going to be version hell for other non-OpenROAD requirements on that same machine...

If this is the case, I would recommend not installing into the system but using venv instead. That way autotuner can lock down versions of modules without worrying about non-OpenROAD python requirements.

luarss:
@oharboe I see. Tentatively I have thought of creating a “supported Python versions” list where we will route dependency installation for corresponding Python versions, but  an easy way out is also to make Autotuner optional like you mentioned

oharboe:
> @oharboe I see. Tentatively I have thought of creating a “supported Python versions” list where we will route dependency installation for corresponding Python versions, but an easy way out is also to make Autotuner optional like you mentioned

Are you familiar with venv?

It is designed to easily and robustly solve exactly the problem you are facing, unless I misunderstand.

With venv you create a "light weight installation"(takes at most a few seconds to set up) of Python where you can specify exactly the versions of modules that you want today using requirements.txt without disturbing the system installation.

luarss:
@oharboe You mean something like this?

```bash
#!/bin/bash

venv_name="my_env"
python3 -m venv "$venv_name"
source "$venv_name/bin/activate"
pip install -r requirements.txt
deactivate
```

oharboe:
> @oharboe You mean something like this?
> 
> ```shell
> #!/bin/bash
> 
> venv_name="my_env"
> python3 -m venv "$venv_name"
> source "$venv_name/bin/activate"
> pip install -r requirements.txt
> deactivate
> ```

Yes.

flow/Makefile
```diff
@@ -441,6 +441,17 @@ $(WRAPPED_LIBS):
 	mkdir -p $(OBJECTS_DIR)/lib
 	sed 's/library(\(.*\))/library(\1_mod)/g' $(filter %$(notdir $(@:_mod.lib=.lib)),$(WRAP_LIBS)) | sed 's/cell(\(.*\))/cell(\1_mod)/g' > $@
 
+# Print variables
+.PHONY: printvars
+printvars:
+	@echo "Printing all variables to VARS.txt"
+	@rm -rf VARS.txt
```
oharboe: why -r? that is recursive, this is a file.

oharboe (changes requested):
nit

oharboe (changes requested):
Write a unittest with mocking for python code so that it can be regression and implementation tested easily and add to CI.

github copilot/chatgpt will do 95% of the work in my experience.

etc/DependencyInstaller.sh
```diff
@@ -34,6 +34,14 @@ _installCommon() {
     else
         pip3 install --no-cache-dir --user -U $pkgs
     fi
+
+    # Autotuner reqs (use venv)
+    local autotuner_pkgs=" $(cat ./etc/requirements.txt)"
```
luarss: don't put in dependency installer.  this forces the user to download autotuner deps. 

Setup a separate script just for autotuner reqs.  -> called optionally by the user. Put inside `autotuner` folder.

etc/DockerHelper.sh
```diff
@@ -66,6 +66,7 @@ _setup() {
         "dev" )
             fromImage="${FROM_IMAGE_OVERRIDE:-$osBaseImage}"
             cp tools/OpenROAD/etc/DependencyInstaller.sh etc/InstallerOpenROAD.sh
+            cp tools/AutoTuner/requirements.txt etc/requirements.txt
```
luarss: Don't need this

flow/Makefile
```diff
@@ -449,6 +449,17 @@ $(WRAPPED_LIBS):
 	mkdir -p $(OBJECTS_DIR)/lib
 	sed 's/library(\(.*\))/library(\1_mod)/g' $(filter %$(notdir $(@:_mod.lib=.lib)),$(WRAP_LIBS)) | sed 's/cell(\(.*\))/cell(\1_mod)/g' > $@
 
+# Print variables
+.PHONY: printvars
+printvars:
```
luarss: Do we still need this?

tools/AutoTuner/src/autotuner/distributed.py
```diff
@@ -328,6 +331,66 @@ def read_tune_ax(name, this):
...
+            # Regex to parse all variables with "``" identifiers that are uppercase.
+            flow_variables = set(match 
+                                for line in f
+                                for match in re.findall(DOCS_UPPERCASE_REGEX, line)
+                                if match.isupper())
+            for line in f:
+                matches = re.findall(regex, line)
+                for match in matches:
+                    if not match.isupper(): continue
+                    flow_variables.add(match)
```
luarss: Remove docs bc they don't have the latest

.github/workflows/github-actions-cron-test-installer.yml
```diff
@@ -45,6 +45,12 @@ jobs:
           sudo service docker restart
       - name: Run installer
         run: |
+          if [[ ${{ matrix.os }} == "ubuntu20.04" ]]; then
+            sudo rm -rf /usr/share/dotnet
+            sudo rm -rf /opt/ghc
+            sudo rm -rf "/usr/local/share/boost"
+            sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+          fi
```
vvbandeira: remove, this action is no longer supported

.gitignore
```diff
@@ -90,3 +90,7 @@ docs/build
 GitGuide.md
 mainREADME.md
 build
+
+# Autotuner
+autotuner_env/
+etc/requirements.txt
```
vvbandeira: Why add this?

docs/user/InstructionsForAutoTuner.md
```diff
@@ -97,7 +106,7 @@ For Global Routing parameters that are set on `fastroute.tcl` you can use:
   - Layer adjustment for layer NAME. This will create a copy of `_FR_FILE_PATH` and modify the layer adjustment only for the layer NAME.
 * `_FR_GR_SEED`
   - Global route random seed. This will create a copy of `_FR_FILE_PATH` and modify the global route random seed.
-
+
```
vvbandeira: remove

docs/user/InstructionsForAutoTuner.md
```diff
@@ -97,7 +97,7 @@ For Global Routing parameters that are set on `fastroute.tcl` you can use:
...
```
vvbandeira:
```suggestion
```

tools/AutoTuner/installer.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+local autotuner_pkgs=" $(cat ./requirements.txt)"
+local venv_name="autotuner_env"
+python3 -m venv $venv_name
+source "$venv_name/bin/activate"
+pip3 install -U $autotuner_pkgs
```
vvbandeira:
```suggestion
local venv_name="autotuner_env"
python3 -m venv $venv_name
source "$venv_name/bin/activate"
pip3 install -U --user -r requirements.txt
```

tools/AutoTuner/installer.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
```
vvbandeira, line 2: I think we need to add code here to consider that this can be called from ORFS root as well as AT's root.

vvbandeira (changes requested):

tools/AutoTuner/installer.sh
luarss: I don't think it should be writing to the user's site-packages, it should be just in the virtual environment no?

tools/AutoTuner/installer.sh
```diff
@@ -0,0 +1,20 @@
...
+# Get the directory where the script is located
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+cd $script_dir
```
vvbandeira:
```suggestion
script_dir="$(dirname "${BASH_SOURCE[0]}")"
cd $script_dir
```

tools/AutoTuner/installer.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Check if the Python version is 3.8
+required_version="3.8"
+python_version=$(python3.8 --version 2>&1)
+if [[ $python_version != "Python $required_version"* ]]; then
+    echo "Error: Python $required_version is required. Current version: $python_version"
+    exit 1
+fi
```
vvbandeira:
```suggestion
required_version="3.8"
if command -v python${required_version} &> /dev/null ; then
    echo "Found python${required_version}"
else
    echo "Error: Python ${required_version} is required."
    exit 1
fi
```

vvbandeira (changes requested):

tools/AutoTuner/installer.sh
```diff
@@ -0,0 +1,21 @@
...
+if command -v python${required_version} &> /dev/null ; then
+    echo "Found python${required_version}"
+else
+    echo "Error: Python ${required_version} is required."
+    exit 1
+fi
+
+# Define the virtual environment name
+venv_name="autotuner_env"
+python3.8 -m venv "$script_dir/$venv_name"
```
vvbandeira:
```suggestion
python${required_version} -m venv "$script_dir/$venv_name"
```

vvbandeira (approved):

flow/designs/asap7/gcd/autotuner.json
```diff
@@ -8,22 +8,6 @@
         ],
         "step": 0
     },
-    "CORE_UTILIZATION": {
```
oharboe, line 11: Is this change in this PR intentional?

tools/AutoTuner/installer.sh
```diff
@@ -0,0 +1,21 @@
...
+# Get the directory where the script is located
+script_dir="$(dirname "${BASH_SOURCE[0]}")"
+cd $script_dir
+
+# Check if the Python version is 3.8
+required_version="3.8"
+if command -v python${required_version} &> /dev/null ; then
+    echo "Found python${required_version}"
+else
+    echo "Error: Python ${required_version} is required."
```
oharboe: What OS ships *specifically* with Python 3.8?

oharboe: Doesn't work on Ubuntu 23.10

oharboe: Shouldn't this be 3.8 or newer?

flow/designs/asap7/gcd/autotuner.json
luarss: Yes, it is. The rationale behind this removal is the that if the golden flow doesn't use it, autotuner will not use it either

tools/AutoTuner/src/autotuner/distributed.py
```diff
@@ -353,6 +396,10 @@ def parse_config(config, path=os.getcwd()):
                       'fully supported, ignoring _SYNTH_FLATTEN parameter.')
         # Default case is VAR=VALUE
         else:
+            # Sanity check: ignore all flow variables that are not tunable
+            if key not in flow_variables:
+                print(f'[ERROR TUN-0017] Variable {key} is not tunable.')
```
oharboe: unlike in OpenROAD, this error number is not enforced. Either drop the OpenROAD numbering scheme or reserve it within OpenROAD. Probably drop it as no standalone tools/ORFS doesn't use OpenROAD error numbering scheme.

oharboe (changes requested):
Questions...

tools/AutoTuner/src/autotuner/distributed.py
luarss: Vitor and I had a discussion, we will probably keep the numbering but reserve the TUN- abbreviation for OpenROAD.

tools/AutoTuner/installer.sh
luarss: We settled on 3.8 because Ubuntu22 is prepacked with 3.10, Ubuntu20 with 3.8, and Centos7 with 3.6, so it happens to be the middle ground. Going forward in future PR we can consider compatibility upwards to at least Python3.10

oharboe: You cant install 3.8 on all environments, if you require *minimum* 3.8, this can be tested on other environments.

3.8 is by now  quite dated...

vvbandeira: @luarss 
We have dropped support for Centos7 since it reached its EOL.
Let's remove these checks and if we reach another compatibility issue we can think how to best address.

vvbandeira (changes requested):

tools/AutoTuner/installer.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
vvbandeira:
```suggestion
#!/usr/bin/env bash
```

tools/AutoTuner/src/autotuner/distributed.py
```diff
@@ -349,10 +392,14 @@ def parse_config(config, path=os.getcwd()):
             elif key == "_PINS_DISTANCE":
                 options += f' PLACE_PINS_ARGS="-min_distance {value}"'
             elif key == "_SYNTH_FLATTEN":
-                print('[WARNING TUN-0013] Non-flatten the designs are not '
+                print('[WARNING 0013] Non-flatten the designs are not '
```
vvbandeira: @luarss 
I thought we had agreed to keep the `TUN-` here.

vvbandeira (changes requested):

tools/AutoTuner/src/autotuner/distributed.py
luarss: Added back.

tools/AutoTuner/src/autotuner/distributed.py
```diff
@@ -328,6 +331,45 @@ def read_tune_ax(name, this):
...
+    pattern = r"(?:::)?env\((.*?)\)"
+    flow_variables = set()
+    for fv_path in paths:
+        with open(fv_path) as f:
+            matches = re.findall(pattern, f.read())
+            flow_variables.update(s.strip().upper()
+                                for match in matches
+                                for s in match.split('\n') 
+                                if len(s))        
+
```
vvbandeira: @luarss is this correct syntax?

vvbandeira (approved):

PR #2033, opened 2024-05-29
makefile: fix memory target dependency

oharboe:
synth builds mem.json as a side-effect, there is no rule with mem.json as a target, so remove the dependency.


oharboe:
@maliberty unrelated centos7 failures in CI

maliberty (approved):

PR #2034, opened 2024-05-29
Makefile restore synth args intent

oharboe:


oharboe:
@maliberty unrelated centos7 failures in CI

maliberty (approved):

PR #2038, opened 2024-05-29
test: deltadebug fix after formatting change in output

oharboe:
@maliberty I take it this is not hooked up to CI yet? Could you check on how that is going?

@NairiBaveyann updated.


maliberty:
> I take it this is not hooked up to CI yet? Could you check on how that is going?

@vvbandeira please update

maliberty (approved):

oharboe:
@maliberty Unrelated error gcd ihp-sg13g2.

vvbandeira:
Not yet. We have a major rewrite of the CI going in very soon. We opted to wait to avoid the double work of adding on the previous and new CI codebase. The new CI will be deployed over the weekend, or we will add the delta debug to the old CI if we find any roadblocks for the new deployment.

PR #2041, opened 2024-05-31
synth: use log_cmd like elsewhere to log synth command args

oharboe:


oharboe:
@maliberty unrelated CI error

maliberty (approved):

PR #2053, opened 2024-06-10
synthesis: allow blackboxing macros

oharboe:
If a macro is specified in ADDITIONAL_LEFS/LIBS and it is also present in the Verilog files, use MACROS to list those macros so as to have the module be marked as blackbox.

A Verilog file can contain more than one module, so it isn't always convenient to exclude a .v/sv file to have a module marked as black box.

### DIFF
--- a/docs/user/FlowVariables.md
+++ b/docs/user/FlowVariables.md
@@ -253,6 +253,7 @@ configuration file.
 | `CDL_FILES`              | Insert additional Circuit Description Language (`.cdl`) netlist files.                             | 
 | `DFF_LIB_FILES`          | Technology mapping liberty files for flip-flops.                                                   |
 | `DONT_USE_LIBS`          | Set liberty files as `dont_use`.                                                                   |
+| `MACROS`                 | Marked as black box during synthesis, defaults to `BLOCKS` |
 | `PRESERVE_CELLS`         | Mark modules to keep from getting removed in flattening.                                           |
 | `SYNTH_ARGS`             | Optional synthesis variables for yosys.                                                            |
 | `VERILOG_TOP_PARAMS`     | Apply toplevel params (if exist).                                                                  |
--- a/flow/scripts/synth_preamble.tcl
+++ b/flow/scripts/synth_preamble.tcl
@@ -57,9 +57,9 @@ if {[info exist ::env(PRESERVE_CELLS)]} {
 
 
 
-if {[info exist ::env(BLOCKS)]} {
+if {[info exist ::env(MACROS)]} {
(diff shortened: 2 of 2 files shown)


oharboe:
@maliberty unrelated errors

maliberty:
If a Liberty and a Verilog for a module are both present, which wins?  It seems like the Liberty should which would render this unneeded.

oharboe:
> If a Liberty and a Verilog for a module are both present, which wins? It seems like the Liberty should which would render this unneeded.

Makes sense.

However, that doesn't cover the case where SYNTH_ARGS=-flatten (default).

maliberty:
> However, that doesn't cover the case where SYNTH_ARGS=-flatten (default).

If the Liberty wins then it should appear in the flat netlist not the contents of the .v (certainly both shouldn't appear).

oharboe:
> > However, that doesn't cover the case where SYNTH_ARGS=-flatten (default).
> 
> If the Liberty wins then it should appear in the flat netlist not the contents of the .v (certainly both shouldn't appear).

Synthesis doesn't load ADDITIONAL_LIBS, nor LEFS, so how can it "win" without having loaded the .lib file?

maliberty:
> Synthesis doesn't load ADDITIONAL_LIBS, nor LEFS, so how can it "win" without having loaded the .lib file?

Why do you think ADDITIONAL_LIBS isn't used?  LEFs are not relevant to synthesis.

asap7 config.mk:
```
export LIB_FILES             += $(ADDITIONAL_LIBS)
```
Makefile:
```
export DONT_USE_LIBS   = $(patsubst %.lib.gz, %.lib, $(addprefix $(OBJECTS_DIR)/lib/, $(notdir $(LIB_FILES))))
```
synth_preamble.tcl:
```
read_liberty -lib {*}$::env(DONT_USE_LIBS)
```

oharboe:
@maliberty I tried introducing a syntax error to track down where ADDITIONAL_LIBS were loaded and I couldn't find them.

Perhaps that is just a false negative too?

https://github.com/The-OpenROAD-Project/OpenROAD/issues/5217

maliberty:
https://github.com/The-OpenROAD-Project/OpenROAD/issues/5217 relates to LEF not Liberty

oharboe:
> [The-OpenROAD-Project/OpenROAD#5217](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5217) relates to LEF not Liberty

Yes, but similarly, I created a syntax error in .lib and had no error in synthesis. I will see about creating a standalone testcase.

oharboe:

PR #2054, opened 2024-06-10
synthesis: remove superfluous blackbox command

oharboe:
blackbox is not necessary for BLOCKS since .lib files are loaded during synthesis and this will mark modules as blackbox

### DIFF
--- a/flow/scripts/synth_preamble.tcl
+++ b/flow/scripts/synth_preamble.tcl
@@ -55,16 +55,6 @@ if {[info exist ::env(PRESERVE_CELLS)]} {
   }
 }
 
-
-
-if {[info exist ::env(BLOCKS)]} {
-  hierarchy -check -top $::env(DESIGN_NAME)
-  foreach block $::env(BLOCKS) {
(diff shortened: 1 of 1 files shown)


flow/scripts/synth_preamble.tcl
```diff
@@ -55,16 +55,6 @@ if {[info exist ::env(PRESERVE_CELLS)]} {
   }
 }
 
-
-
-if {[info exist ::env(BLOCKS)]} {
-  hierarchy -check -top $::env(DESIGN_NAME)
```
maliberty, line 61: Why remove this check?

oharboe: This is not really a check, it is just making sure that the hierarchy is evaluated so that blackbox can be run.

synth_preamble.tcl is trying to leave as much processing for synth.tcl as possible, near as I can tell.

This check is still being run, only later, as is the intent of synth_preamble/synth.tcl, near as I can tell.

maliberty (approved):

PR #2057, opened 2024-06-10
refactor deltadebug

habibayassin:
Merged together with this [PR](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5223) in OR

### DIFF
--- a/flow/test/test_delta_debug.sh
+++ b/flow/test/test_delta_debug.sh
@@ -18,4 +18,4 @@ mkdir -p results/delta-debug/
 cd results/delta-debug/
 tar --strip-components=1 -xzf ../../$latest_file
 sed -i 's/openroad -no_init/openroad -exit -no_init/g' run-me-$testname-asap7-base.sh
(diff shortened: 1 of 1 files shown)


oharboe (changes requested):
why are the paths added? if these are required now, then deltaDebug is less userfriendly than before

habibayassin:
> why are the paths added? if these are required now, then deltaDebug is less userfriendly than before

@oharboe what do you suggest? should it have the paths set in the script itself as a default?

oharboe:
@habibayassin @maliberty There are now paths that I have never heard about that I have to pass in to deltaDebug.py. 

I just don't know what to make of it, so to my mind, the user interface to deltaDebug.py got worse...

oharboe:
@habibayassin @vvbandeira @maliberty Are the deltaDebug.py tests hooked up to CI https://github.com/The-OpenROAD-Project/OpenROAD/issues/5377 ?

PR #2059, opened 2024-06-11
asap7/mock-array: switch manual macro placement for mpl2 flow

### LINKED ISSUE #1928: Switch manual macro placement by mpl2 flow on mock-array
### Description

Asap7mock-array is using manual macro placement, because our macro placer was not able to properly handle it. With mpl2 new feature for arrays of interconnected macros, manual macro placement shouldn't be necessary.

### Suggested Solution

Turn on mpl2 flow and set boundary penalty to  0.

### Additional Context

_No response_

AcKoucher:
Resolve #1928,

@maliberty 
**I'm updating OR submodule as there's an [mpl2 bug fix](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5478) on master which is necessary for the changes here to work.**

Using #2020 to test mpl2 flow for mock-array showed some metrics degradation in both timing and WL caused by a [group of IOs placed in a different location than we see on master](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2020#issuecomment-2121167498):

```
[ERROR] detailedroute__route__wirelength fail test: 76149.0 <= 73578.0
[ERROR] finish__timing__setup__ws fail test: -78.1376 >= -6.93
[ERROR] finish__timing__wns_percent_delay fail test: -49.411146 >= -10.0
```

When using manual macro placement, when we get to non-random pin placement stage in the flow, PPL has no information about the std cells position - because non of them are placed - so we luckily end up placing the problematic IO group in the center of the right boundary, because PPL uses as reference the center of the design when it finds no placed insts.

However, MPL2 does generate a temporary location for std cells in order to improve macros' orientations. These locations end up been seen by PPL which chooses a different position for the IO group changing upstream results leading to degradation.

A possible solution would be to not skip global placement without IOs when we have constraints for the pins as it is for mock-array. This way, PPL would have a better std cell location data to work with.
Testing this showed that std cells get split based on the arrays' rows and so does the IO group (in the images zooming into the fourth row from bottom to top):

<p float="left">
  
  
  
</p>

This improves WR and ws and we get no failures.
```
[INFO] detailedroute__route__wirelength pass test: 63606.0 <= 73578.0
[INFO] finish__timing__setup__ws pass test: 34.7588 >= -6.93
[INFO] finish__timing__wns_percent_delay pass test: 6.875599 >= -10.0
```

Master:
```
[INFO] detailedroute__route__wirelength pass test: 69582.0 <= 73578.0
[INFO] finish__timing__setup__ws pass test: 10.0482 >= -6.93
[INFO] finish__timing__wns_percent_delay pass test: 14.364462 >= -10.0
```


AcKoucher:
@oharboe FYI

One additional thing:
In this PR the halos generate the exact same macro placement as we'd get using manual macro placement (In your test PR there were some subtle differences). It was just a matter of using a more precise value. I made that to exclude the possibility of the macro placement to be the actual cause of the degradation we see in your test PR.

AcKoucher:
@maliberty Is running a separate secure-ci for the global placement without IOs.

oharboe (changes requested):
Keep manual placement around for testing purposes for now, just in case.

Use export RTLMP_FLOW ?= True (on by default) and then conditionally include manual macro placement.

Once automated macro placement has matured for a while, the manual placement testcase can be deleted.

oharboe:
I looked at the significant WNS regression:

```
[ERROR] finish__timing__setup__ws fail test: -91.4696 >= -6.93
```

io_lsbs_61 is a pipeliened signal from ces_7_5, i.e. top row that is driven all the way to the lower right corner with only a single flip flop for that entire path.

image



image

oharboe:
On master, the io_lsbs_* pins are placed in the middle on the right, which improves WNS by ca. 100ps on a clock period of 250ps.

image

AcKoucher:
@oharboe The changes here depend on _not skipping global placement without IOs in the flow_ which is skipped on master for mock-array (we skip this stage when we have pin constraints). We'll need another PR for this skipping part. We're running a secure-ci for it separately, but there are some other issues regarding different designs which are being studied.

oharboe:
@AcKoucher Any news on this one?

maliberty:
He's waiting on me - I'm nearly done with the problems within problems.

AcKoucher:
@oharboe Finally this should be good to merge :) I left the manual mpl file as you requested.

maliberty:
Please merge master and resolve the OR conflict.

flow/designs/asap7/mock-array/config.mk
```diff
@@ -23,11 +23,17 @@ export DIE_AREA  = $(shell \
...
+export MACRO_PLACE_HALO = 0 2.16
+export RTLMP_BOUNDARY_WT = 0
+export RTLMP_FLOW ?= True
+
 export BLOCKS                ?= Element
 
 ifneq ($(BLOCKS),)
   export GDS_ALLOW_EMPTY       = Element
-  export MACRO_PLACEMENT_TCL   = ./designs/asap7/mock-array/macro-placement.tcl
+  ifeq ($(RTLMP_FLOW), False)
```
oharboe: An niggling little thing about ORFS variables is that there are no types... RTLMP_FLOW actually checks for empty and non-empty....

Also, I would like to keep manual placement under CI.

See here on how to do that:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/designs/asap7/riscv32i-mock-sram/config.mk

1. create designs/asap7/mock-array-manual/config.mk
2. set RTLMP_FLOW to empty
3. include designs/asap7/mock-array/config.mk
4. add the baseline files(make metadata/update_ok/rules, etc.)

oharboe: Fewer surprises: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2208

oharboe (changes requested):
Keep manual placement mock-array under CI. For now anyway...

maliberty:
> Keep manual placement mock-array under CI. For now anyway...

What is the criteria for ending it?  What is the benefit to keeping it?

oharboe:
> > Keep manual placement mock-array under CI. For now anyway...
> 
> What is the criteria for ending it? What is the benefit to keeping it?

Is there any other manual placement with abutment test-case?

maliberty:
> Is there any other manual placement with abutment test-case?

No but is that substantially different from auto placement with abutment?

oharboe:
> > Is there any other manual placement with abutment test-case?
> 
> No but is that substantially different from auto placement with abutment?

Perhaps not worth it then...

Cost effective testing is always a challenge and important for OpenROAD, since running times are very long.

oharboe (approved):
Needs an update after #2208 got merged, but otherwise I say merge it. Very nice! :-)

maliberty (approved):

PR #2061, opened 2024-06-12
makefile: delete preview macro stuff

oharboe:
I was excited about this feature, then I ended up never using it and forgot all about it.

Perhaps someone is using it, but the only way to know is to delete it and see if anyone protests in pull request or after it is merged, in which case it is easily restored or reimplemented.


oharboe:
@maliberty unrelated CI failure

maliberty (approved):

PR #2069, opened 2024-06-17
macro placement: less surprising logging

oharboe:
The `[INFO][FLOW-xxxx]` prefix is surprising, leaves questions... What is xxxx? Is something wrong? Should the code be updated?

Generally the .tcl code just puts logging information, so do the same here. There are no other occurrences in the .tcl scripts of `[INFO][FOO-]`.

The openroad binary has error/warning/info numbers that are allocated and guaranteed by build time checks to be unique.


maliberty (approved):

rovinski:
xxxx should have been a 4 digit number just like OR messages. But there is no enforcement mechanism to do so, so things like this slip through the cracks.

PR #2071, opened 2024-06-18
flow: removed vestiges of prefixing convention in puts/logging

oharboe:
this convention was never followed through in .tcl, so remove it.


maliberty (approved):

oharboe:

maliberty (approved):

rovinski:
I think there are some in Makefiles and also Python scripts too

oharboe:
> I think there are some in Makefiles and also Python scripts too

I did some grepping, could not find anything more after the first few pull requests.

PR #2073, opened 2024-06-18
asap7: delete some inconsistent logging

oharboe:
This logging is not done by other PDKs, so more consistent now.

I'm highly suspicious of that version number that is being logged.

Vestige? Where is the definition of that version and how do I check if it is the most recent?


maliberty (approved):

rovinski:
I think the version number is just internal naming from @louiic. Not related to the PDK. Safe to remove.

PR #2078, opened 2024-06-19
makefile: fix eqy failure, use ORFS built eqy

oharboe:
@maliberty A quick stab at boarding up the broken window with room to refine in the future...

```
$ make DESIGN_CONFIG=designs/nangate45/aes/config.mk cts
[deleted]
[INFO RSZ-0032] Inserted 3 hold buffers.
Error: cts.tcl, 102 couldn't execute eqy: no such file or directory
```

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -303,6 +303,8 @@ OPENROAD_GUI_CMD         = $(OPENROAD_EXE) -gui $(OR_ARGS)
 
 YOSYS_CMD               ?= $(abspath $(FLOW_HOME)/../tools/install/yosys/bin/yosys)
 
+export EQY_EXE ?= $(abspath $(FLOW_HOME)/../dependencies/bin/eqy)
+
 # Use locally installed and built klayout if it exists, otherwise use klayout in path
 KLAYOUT_DIR = $(abspath $(FLOW_HOME)/../tools/install/klayout/)
 KLAYOUT_BIN_FROM_DIR = $(KLAYOUT_DIR)/klayout
--- a/flow/scripts/load.tcl
+++ b/flow/scripts/load.tcl
@@ -111,7 +111,7 @@ proc run_equivalence_test {} {
     write_eqy_verilog 4_after_rsz.v
     write_eqy_script
 
-    eval exec eqy -d $::env(LOG_DIR)/4_eqy_output \
+    eval exec $::env(EQY_EXE) -d $::env(LOG_DIR)/4_eqy_output \
         --force \
         --jobs $::env(NUM_CORES) \
         $::env(OBJECTS_DIR)/4_eqy_test.eqy \


oharboe:
@maliberty A quick stab at fixing this...

oharboe:
@maliberty failed... Something else is needed here. Thoughts?


```
$ make DESIGN_CONFIG=designs/nangate45/aes/config.mk do-cts
[deleted]
[INFO RSZ-0032] Inserted 3 hold buffers.
Error: cts.tcl, 102 ERROR: Failed to combine designs. For details see './logs/nangate45/aes/base/4_eqy_output/combine.log'.
Elapsed time: 3:06.23[h:]min:sec. CPU time: user 196.70 sys 0.90 (106%). Peak memory: 418444KB.
make[1]: *** [Makefile:759: do-4_1_cts] Error 1
make: *** [Makefile:767: do-cts] Error 2
```


```
-- Executing script file `./logs/nangate45/aes/base/4_eqy_output/combine.ys' --
ERROR: Can't load module `/home/oyvind/OpenROAD-flow-scripts/dependencies/bin/../share/yosys/plugins/eqy_combine.so': /home/oyvind/OpenROAD-flow-scripts/dependencies/bin/../share/yosys/plugins/eqy_combine.so: undefined symbol: _ZN5Yosys5RTLIL8IdString17destruct_guard_okE
```

flow/Makefile
```diff
@@ -303,6 +303,8 @@ OPENROAD_GUI_CMD         = $(OPENROAD_EXE) -gui $(OR_ARGS)
 
 YOSYS_CMD               ?= $(abspath $(FLOW_HOME)/../tools/install/yosys/bin/yosys)
 
+export EQY_EXE ?= $(abspath $(FLOW_HOME)/../dependencies/bin/eqy)
```
maliberty, line 306: This will fail if eqy is in the path but not at this location.

oharboe: So how can this be made to work w ORFS? Should eqy be in tools/install?

maliberty: We don't currently demand it be anywhere besides in your path.  Why do we need to narrow that?  For example I often take it from /usr/local/bin/eqy

oharboe: I thought it would be consistent w openroad. Some particular version of eqy is required for a version of ORFS, no?

maliberty:
You might look at KLAYOUT_CMD and how we have a preferred path and a fall back.  Currently there is no version requirement though I suppose that could arise in the future.

oharboe:
Where is eqy locally and on the build server in ORFS?

maliberty:
> Where is eqy locally and on the build server in ORFS?

I think you should use the location that the dependency installer uses or the users path rather than list other possible locations.

PR #2083, opened 2024-06-22
makefile: improve memory report of used source files

oharboe:
This method of finding source files actually in use is useful and timesaving to maintain a list of dependencies, but it is not 100% robust, some files could be left out. Plugged one more hole.


maliberty (approved):

maliberty:
Would you explain what gets left out in the previous method?

oharboe:
> Would you explain what gets left out in the previous method?

The top level when it contains no logic, when it only connects modules.

PR #2100, opened 2024-07-01
gpl: include default overflow threshold values in gpl script

gudeh:
This should provide easier access for the user to modify values. For example, removing some values can reduce runtime. As mentioned in the text added to gpl README in OR PR: https://github.com/The-OpenROAD-Project/OpenROAD/pull/5318

### DIFF
--- a/flow/scripts/global_place.tcl
+++ b/flow/scripts/global_place.tcl
@@ -41,10 +41,6 @@ if {$::env(GPL_ROUTABILITY_DRIVEN)} {
 # Parameters for timing driven mode in global placement
 if {$::env(GPL_TIMING_DRIVEN)} {
   lappend global_placement_args {-timing_driven}
-  
-  # Timing-driven iterations will trigger with these overflow values
-  set reweight_overflow {79 64 49 29 21 15}
-  lappend global_placement_args {-timing_driven_net_reweight_overflow} $reweight_overflow
 }
 
 proc do_placement {place_density global_placement_args} {
--- a/flow/scripts/global_place.tcl
+++ b/flow/scripts/global_place.tcl
@@ -29,6 +29,8 @@ if {[info exist ::env(PLACE_DENSITY_LB_ADDON)]} {
 }
 
 set global_placement_args {}
+
+# Parameters for routability mode in global placement
 if {$::env(GPL_ROUTABILITY_DRIVEN)} {
   lappend global_placement_args {-routability_driven}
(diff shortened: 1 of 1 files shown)


gudeh:
This should help address the OR issue: https://github.com/The-OpenROAD-Project/OpenROAD/issues/5203

flow/scripts/global_place.tcl
```diff
@@ -29,15 +29,22 @@ if {[info exist ::env(PLACE_DENSITY_LB_ADDON)]} {
...
   }
 }
 
+# Parameters for timing driven mode in global placement
 if {$::env(GPL_TIMING_DRIVEN)} {
   lappend global_placement_args {-timing_driven}
+  
+  # Timing-driven iterations will trigger with these overflow values
+  set reweight_overflow {79 64 49 29 21 15}
+  lappend global_placement_args {-timing_driven_net_reweight_overflow} $reweight_overflow
```
maliberty: If OR has defaults why include them here?

gudeh: The only reason would be to make it more clear for the user to what are the default values and give him an easy path to change them. This has a good impact on gpl runtime, and the the default values are kind of hidden in the replace.tcl file.

Also, I believe @oharboe did not know about these values from his issue asking the log messages to be more actionable.

oharboe: I had not heard about these values and I don't know what they mean. global placement is just a magic black box to me... 

Note that I am telling anyone that cares to listen that the the scripts/.tcl files for each stage is effectively part of the ORFS documentation and that they should expect to have to read and understand each script that affects some part of the flow that they need to understand.

So to my mind, improving readability of scripts/stage.tcl is more than an ORFS developer priority, it is an end user concern...

gudeh: I chose to keep the default values in OR because of its testings, also the ORFS global_placement.tcl script should always overwrite the values in OR.

maliberty: These are the sort of knobs that are useful to a developer and mostly meaningless to a user.  If we want to discuss the default values then we can put that in the docs but I don't want to duplicate the code.  Changing the values in gpl will have no effect in ORFS but will have an effect in any other flow which isn't good.

maliberty: If you want to keep this PR to add the comments that fine but I'm rejecting this part of the change.

maliberty (approved):

PR #2101, opened 2024-07-02
synth: fix renaming of registers and wires

oharboe:
Do the renaming earlier so that names of non-private registers are preserved.

Doing the renaming later after more optimizations and transformations in synthesis will cause some "collateral renaming".

Example of improved names. 

```
make DESIGN_CONFIG=designs/asap7/mock-alu/config.mk synth
```

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/f8d87d5bf1b2fa9a7e8724d1586a674180b31ae9/flow/designs/src/mock-alu/MockAlu.v#L66

I can now find my register:

```
$ grep io_out_REG results/asap7/mock-alu/base/1_synth.v  | head -3
  DFFHQNx1_ASAP7_75t_R \io_out_REG[0]$_DFF_P_  (
  DFFHQNx1_ASAP7_75t_R \io_out_REG[10]$_DFF_P_  (
  DFFHQNx1_ASAP7_75t_R \io_out_REG[11]$_DFF_P_  (
```

PR #2102, opened 2024-07-02
[DO NOT REVIEW YET] mock-alu: add a check in .sdc that output register can be found

oharboe:
Will work after https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2101 is merged.


oharboe:
@tspyrou Something like this?

oharboe:
out of sight, out of mind for now.

PR #2112, opened 2024-07-08
mock-array: update .gitignore

oharboe:


flow/designs/src/mock-array/.gitignore
```diff
@@ -5,3 +5,5 @@ target/
 .metals/
 .vscode/
 .bloop/
+.bsp/
+test_run_dir/
```
maliberty, line 9: When would this be created?

maliberty (approved):

PR #2115, opened 2024-07-08
GHA: add push pr job for update rules

habibayassin:


.github/workflows/github-actions-cron-sync-fork-from-upstream.yml
```diff
@@ -10,8 +10,6 @@ on:
         description: Use GitHub --force push.
         default:
 
-  repository_dispatch:
-
```
vvbandeira, line 14: @maliberty @tspyrou 

We are removing this trigger as it conflicts with the `update-rules.yml` action. From what I can tell, this is redundant since the action is triggered by the `cron` above. Let me know if you see any weird behavior in the actions today, and we will revert this PR.

vvbandeira (approved):

PR #2116, opened 2024-07-09
run prima dcalc on aes

tspyrou:


flow/designs/asap7/aes/constraint.sdc
```diff
@@ -1,3 +1,5 @@
+sta::set_delay_calculator prima
```
maliberty, line 1: We have gcd setup for CCS though scripts/load.tcl

```
  if { [info exists ::env(LIB_MODEL)] && $::env(LIB_MODEL) == "CCS" } {
    puts "Using CCS delay calculation"
    set_delay_calculator ccs_sim
  }
```

Should we change ccs_sim to prima?  We don't need to put it in the SDC.

maliberty: designs/asap7/gcd-ccs/config.mk:export LIB_MODEL = CCS

tspyrou: I didn't know that, I will close this PR. Yes changing it to Prima would be good.

PR #2134, opened 2024-07-12
makefile: SUB_MAKE tweak

oharboe:
Be explicit about which makefiles to invoke

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -1,3 +1,4 @@
+THIS_FILE = $(MAKEFILE_LIST)
 # settings.mk is not under source control. Put variables into this
 # file to avoid having to adding the to the make command line.
 -include settings.mk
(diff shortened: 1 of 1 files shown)


maliberty (approved):
What problem prompted this PR?

oharboe:
> What problem prompted this PR?

bazel use cases where the project is not in the same rep as orfs....

maliberty:
This breaks the CI at UCSD and will have to be reverted.  It produces
```
make: the '-f' option requires a non-empty string argument
```
on every build.

oharboe:
> This breaks the CI at UCSD and will have to be reverted. It produces
> 
> ```
> make: the '-f' option requires a non-empty string argument
> ```
> 
> on every build.

Where can I see this error? What version of make are they running?

This should fix it:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2134

Could you:

- run `make print-MAKEFILE_LIST`
- Test the updated https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2138

PR #2135, opened 2024-07-12
ORFS Dockerfile incremental optimization

ivysochyn:
# ORFS Dockerfile incremental optimization

This PR splits the final layers in the Docker image to reduce the download size for consecutive pulls, as well as enhances the Docker image build process by optimizing the Dockerfile and leveraging build-cache more effectively.

By separating the image into additional stages, we ensure faster pull speeds when minor changes are introduced, and significantly improve the build process by utilizing previously produced build cache.

You can find example built images under:

* [ghcr.io/antmicro/openroad-flow-scripts-test-cache/ubuntu22.04:latest](https://github.com/antmicro/OpenROAD-flow-scripts/pkgs/container/openroad-flow-scripts-test-cache%2Fubuntu22.04)
* [ghcr.io/antmicro/openroad-flow-scripts-test-cache/ubuntu20.04:latest](https://github.com/antmicro/OpenROAD-flow-scripts/pkgs/container/openroad-flow-scripts-test-cache%2Fubuntu20.04)

CI runs demonstrating the improvements:

1. **Image Built from Scratch**: [View Run](https://github.com/antmicro/OpenROAD-flow-scripts/actions/runs/9890334708)
2. **Minor Modification to README**: This change does not affect the ORFS tools and does not impact most of the ORFS Docker image layers, showcasing the efficiency of the build cache. [View Run](https://github.com/antmicro/OpenROAD-flow-scripts/actions/runs/9891162104) 

This PR introduces the following modifications:

* `docker/Dockerfile.builder`:
  * Multi-Stage Build: The image is divided into multiple stages of copying sources, to leverage caching effectively.
  * Tool Installation: The installation of tools is modified to prevent invalidation of the build cache by avoiding unnecessary copying of git-submodule files.

### DIFF
--- a/docker/Dockerfile.builder
+++ b/docker/Dockerfile.builder
@@ -1,27 +1,42 @@
+# syntax=docker/dockerfile:1.7-labs
+
 # Target with dependencies to build all flow tools from their sources.
 # i.e., "./build_openroad.sh --local" from inside a docker container
 # NOTE: don't use this file directly unless you know what you are doing,
 # instead use etc/DockerHelper.sh
+
 ARG fromImage=openroad/flow-ubuntu22.04-dev:latest
 
-FROM $fromImage AS openroad-builder-base
+FROM $fromImage AS orfs-base
 
+WORKDIR /OpenROAD-flow-scripts
+COPY --link dev_env.sh dev_env.sh
+COPY --link build_openroad.sh build_openroad.sh
+
+FROM orfs-base AS orfs-builder-base
+
+COPY --link tools tools
 ARG numThreads=$(nproc)
 
-COPY . /OpenROAD-flow-scripts
-WORKDIR /OpenROAD-flow-scripts
+RUN echo "" > tools/yosys/abc/.gitcommit && \
+  ./build_openroad.sh --no_init --local --threads ${numThreads}
 
-RUN ./build_openroad.sh --no_init --local --threads ${numThreads}
+FROM orfs-base
 
-FROM $fromImage AS openroad-flow-scripts-base
+# The order for copying the directories is based on the frequency of changes (ascending order),
+# and the layer size (descending order)
+COPY --link docker docker
+COPY --link flow/tutorials flow/tutorials
+COPY --link docs docs
+COPY --link flow/test flow/test
+COPY --link flow/platforms flow/platforms
+COPY --link flow/util flow/util
+COPY --link flow/scripts flow/scripts
+COPY --link flow/designs flow/designs
 
-COPY . /OpenROAD-flow-scripts
-
-RUN rm -rf /OpenROAD-flow-scripts/tools /OpenROAD-flow-scripts/.git
-
-COPY --from=openroad-builder-base /OpenROAD-flow-scripts/tools/install /OpenROAD-flow-scripts/tools/install
-
-FROM $fromImage
-
-COPY --from=openroad-flow-scripts-base /OpenROAD-flow-scripts /OpenROAD-flow-scripts
-WORKDIR /OpenROAD-flow-scripts
+COPY --link --from=orfs-builder-base /OpenROAD-flow-scripts/tools/install tools/install
+COPY --link \
+     --exclude=.git* --exclude=tools/ --exclude=docs/ --exclude=docker/ \
+     --exclude=flow/designs --exclude=flow/platforms --exclude=flow/scripts \
+     --exclude=flow/test --exclude=flow/tutorials --exclude=flow/util \
+  . ./


docker/Dockerfile.builder
```diff
@@ -1,27 +1,39 @@
...
 
-COPY --from=openroad-flow-scripts-base /OpenROAD-flow-scripts /OpenROAD-flow-scripts
-WORKDIR /OpenROAD-flow-scripts
+RUN echo "" > tools/yosys/abc/.gitcommit && \
+  ./build_openroad.sh --no_init --local --threads ${numThreads}
+
+FROM orfs-base
+
+COPY --link flow/designs flow/designs
+COPY --link flow/platforms flow/platforms
```
oharboe: sort these by descending update size/frequency in git history. Especially platforms is important to have at the top, it is big and rarely changes.

oharboe (changes requested):
suggestion to reduce further. flow/scripts changes freqeuntly and is small and platform changes rarely and is big, so platforms should be before flow/scripts

docker/Dockerfile.builder
```diff
@@ -1,27 +1,39 @@
...
+ARG numThreads=$(nproc)
 
-COPY --from=openroad-flow-scripts-base /OpenROAD-flow-scripts /OpenROAD-flow-scripts
-WORKDIR /OpenROAD-flow-scripts
+RUN echo "" > tools/yosys/abc/.gitcommit && \
+  ./build_openroad.sh --no_init --local --threads ${numThreads}
+
+FROM orfs-base
+
+COPY --link flow/platforms flow/platforms
```
oharboe: add comment that sorting these matter...

docker/Dockerfile.builder
```diff
@@ -1,27 +1,39 @@
...
+COPY --link flow/designs flow/designs
+COPY --link flow/tutorials flow/tutorials
```
oharboe, line 29: These are still not sorted... Can chatgpt write a small script? :-)

oharboe (changes requested):
document that sorting matters.

docker/Dockerfile.builder
ivysochyn: As you suggested in a previous comment, those layers are sorted by `update size` (lines)  in descending order.
Let me know if there are any other modifications to the order you want to see.

Here's a look into the statistics for those layers (calculated on this branch starting from the initial commit):
| Name            | Commits | Updated lines    | Docker layer size |
|-----------------|---------|------------|------------|
| flow/platforms  | 519     | 53224454   |  729 MB |
| flow/designs    | 1763    | 11026554   | 284 MB |
| flow/tutorials  | 10      | 36408      | 4.4 MB |
| docs            | 108     | 7146       |  4.1 MB |
| flow/util       | 702     | 18579      |  240 KB |
| flow/scripts    | 839     | 16455      | 112 KB |
| flow/test       | 74      | 291        | 8 KB
| docker          | 9       | 75         | 2 KB

docker/Dockerfile.builder
ivysochyn: Added comment stating that layer sorting is performed based on their size.

docker/Dockerfile.builder
oharboe: You also need to take update frequency into account...

So flow/designs change MUCH more often than tutorials for instance and show be lower down.

vvbandeira: @ivysochyn 
Could you share a GH Gitst of the script/command line you used to generate the data? This looks interesting and useful for gathering data from other repos.

maliberty: I suspect the rebuilding the <10mb layers is trivial in runtime and designs + platforms is most of the time.  If true, the table order is reasonable.

I expect frequency is most-to-least: scripts, designs, platforms, util, ...

ivysochyn: Taking into account both the frequency of commits and the size of the layers, I would rather prioritize frequency. This results in the following order:

| Name            | Commits | Updated lines | Docker layer size |
|-----------------|---------|---------------|-------------------|
| docker          | 9       | 75            | 2 KB              |
| flow/tutorials  | 10      | 36408         | 4.4 MB            |
| docs            | 108     | 7146          | 4.1 MB            |
| flow/test       | 74      | 291           | 8 KB              |
| flow/platforms  | 519     | 53224454      | 729 MB            |
| flow/util       | 702     | 18579         | 240 KB            |
| flow/scripts    | 839     | 16455         | 112 KB            |
| flow/designs    | 1763    | 11026554      | 284 MB            |

This order aims to optimize the cache usage, as the first layers are less likely to be rebuilt frequently.

Additionally, in the changes introduced in this PR, I rely on the [COPY --link](https://docs.docker.com/reference/dockerfile/#copy---link) behavior, which should treat all of those layers independently, preserving the cache even if previous layers are invalidated.

For your reference, here are the tools and commands I used to collect this data.

To calculate the number of commits:
```
git log --oneline <FILE/DIRECTORY> | wc -l
```

Lines updated calculation:
```
git log --shortstat -- <FILE/DIRECTORY> | grep "files changed" | awk '{inserts += $4; deletes += $6} END {print "Insertions: " inserts "\nDeletions: " deletes "\nTotal Changes: " inserts + deletes}
```

To see details of each specific layer of the image, I usually go with [dive](https://github.com/wagoodman/dive):
```
docker run -ti --rm  -v /var/run/docker.sock:/var/run/docker.sock wagoodman/dive <DOCKER_IMAGE_TAG>
```

oharboe: Makes sense to.me!

oharboe (approved):

vvbandeira (approved):

PR #1813, opened 2024-02-20
syn: move dont_use processing to yosys

rovinski:
* Use new `dont_use` flag for `dfflibmap`
* Remove `dont_use` preprocessing from `markDontUse.py`
* Rename `markDontUse.py` to `preprocessLib.py`

This change requires a sync from upstream yosys. Support was added in https://github.com/YosysHQ/yosys/pull/4219; it should be upstream next month in yosys 0.39. Alternatively, we could update yosys immediately.

PR #2139, opened 2024-07-13
Pass multiple lib files to ABC

rovinski:
Yosys allows for passing though multiple .lib files to ABC since version 0.41. Therefore, external liberty merging is no longer required in any part of the flow. This PR removes liberty merging and updates variable names

* Change `DONT_USE_*` variables to `PROCESSED_*` variables to better reflect their meaning (we stopped modifying `dont_use` properties in lib files since #1813)
* Remove the mergeLib rule in Makefile since there are no more users of it
* Pass multiple libs into the abc options
* Add a default `DFF_LIB_FILE` for `asap7` because it appears that it was always supposed to have one, but it instead relied on the FF files being merged into the standard cell files

If/when `dfflibmap` supports reading multiple lib files, we can completely remove `DFF_LIB_FILE` and `PROCESSED_SC_LIB` variables. Restructure may lose some support but the underlying abc version would just have to be updated.

### DIFF
--- a/docs/user/BuildWithPrebuilt.md
+++ b/docs/user/BuildWithPrebuilt.md
@@ -5,7 +5,7 @@ Please ensure the Klayout version (denoted with `klayoutVersion` variable) is co
 
 Instructions for installing:
 - [Klayout>=0.28.8](https://www.klayout.de/build.html)
-- [Yosys>=0.39](https://github.com/YosysHQ/oss-cad-suite-build/blob/master/README.md#installation)
+- [Yosys>=0.41](https://github.com/YosysHQ/oss-cad-suite-build/blob/master/README.md#installation)
 
 ```{tip} Unfortunately KLayout maintainers do not provide Debian 10/11 compatible packages. You can follow the build-from-sources instruction (Version >=0.25) and Ubuntu 22 instructions [here](https://www.klayout.de/build.html#:~:text=Building%20KLayout%20on%20Linux%20(Version%20%3E%3D%200.25)). 
 ```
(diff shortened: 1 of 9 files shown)


oharboe:
To see what is going on with the failures in projects with BLOCKS, run:

Run 'nuke' here so that submacros are erased too just to run a full test:

```
make nuke
```

Now build without the YOSYS_FLAGS that hide the error message we saw in CI:

```
make DESIGN_CONFIG=designs/asap7/aes-block/config.mk YOSYS_FLAGS= synth
```

Output:

```
BC: execution of command ""/home/oyvind/OpenROAD-flow-scripts/tools/install/yosys/bin/yosys-abc" -s -f /tmp/yosys-abc-ymYdUm/abc.script 2>&1" failed: return code 134.
```

Re-run that last command:

```
"/home/oyvind/OpenROAD-flow-scripts/tools/install/yosys/bin/yosys-abc" -s -f /tmp/yosys-abc-ymYdUm/abc.script
```

Output:

```
[deleted]
yosys-abc: /home/oyvind/OpenROAD-flow-scripts/tools/yosys/abc/src/misc/vec/vecStr.h:356: char abc::Vec_StrEntry(Vec_Str_t *, int): Assertion `i >= 0 && i < p->nSize' failed.
Aborted (core dumped)
```

```
cd flow
$ ../tools/install/yosys/bin/yosys -version
Yosys 0.43 (git sha1 ead4718e5, clang++ 16.0.6 -fPIC -Os)
$ ../tools/install/yosys/bin/yosys-abc
UC Berkeley, ABC 1.01 (compiled Jul 14 2024 13:19:17)
```

oharboe:
@rovinski Filed https://github.com/berkeley-abc/abc/issues/306

maliberty:
@rovinski do you plan to continue with this or close?

rovinski:
I forgot this was open. I think it would be best to close because several things have changed with Yosys since then (dont use lists for several commands, new commands, etc.). If someone wants to investigate, they are welcome.

PR #2141, opened 2024-07-15
Makefile.py: remove, not updated since 2022, so probably not in use

oharboe:


maliberty:
@gadfort this is part of the SiC integration work that hasn't had much attention.  Do you plan to come back to it?  If not, scripts/sc should be part of this PR.

oharboe:
In bazel-orfs there's stuff that *could* have gone into ORFS that is kept in bazel-orfs because it would be too hard to communicate the concerns for the bazel-orfs use-case to the ORFS developers considering it is not part of their daily grind.

For instance, some additional .mk files are kept in bazel-orfs, it is more of an implementation detail in bazel-orfs than part of ORFS as such.

gadfort:
@maliberty I don't think we have time to maintain this, so it would be better to remove all of it, and those who want to use SC can use the SC asicflow which is close to this flow.

maliberty (approved):

PR #2148, opened 2024-07-17
Makefile: add OPENSTA_EXE for read_power_activites in scripts

oharboe:


oharboe:
@maliberty Unrelated problems that must be on master, I think.

I wonder if the recent TNS_END_PERCENT has fuzzed the test-cases... Fixing all those bugs it has unveiled is good of course...

maliberty (approved):

PR #2150, opened 2024-07-17
mock-alu: lock TNS_END_PERCENT to 5

oharboe:


maliberty:
Why?

oharboe:
> Why?

Increasing it to 100 breaks things and investigating that is a separate concern from making master work again. Or we can revert the default to TNS_END_PERCENT and play whackamole on the PR that introduces TNS_END_PERCENT 100 default. It's 6 of one or half a dozen of the other to my mind...

eder-matheus:
This design has no issues in PR 2142. Closing this PR.

PR #2157, opened 2024-07-18
asap7: add path groups to constraints.sdc

oharboe:
platforms/asap7/constraints.sdc, read the comments in that file for the full story, is a shared constraints file for designs, such as mock-array, that is building a hierarchy of macros.

Add path groups to this design so that the slack histogram can filter on path group.

Constraints must be met for each macro in the reg2reg path group, but violations at the macro level for paths outside of the reg2reg path group may or may not be a problem higher up when the macro is used.


oharboe:
@maliberty Unrelated CI error in mock-alu. mock-alu doesn't use the constraints.sdc file I changed.

maliberty (approved):

PR #2161, opened 2024-07-19
mock-array: make Element more interesting in simulation

oharboe:
Previously everything would converge to 1 because everything was OR'ed together, so XOR together to wiggle things more.


maliberty (approved):
Should this imply a new VCD?

oharboe:
> Should this imply a new VCD?

Yes, tested locally w that other PR you just merged.


Note that the .vcd should not be in git as it is from post synthesis

PR #2164, opened 2024-07-19
mock-alu: add a check in .sdc that output register can be found

oharboe:
@tspyrou asked for a test that names are nice...


maliberty (approved):
when does this happen?

oharboe:
> when does this happen?

the .sdc file is read in in floorplan.tcl and if the signal does not exist, then floorplan.tcl will fail.

This is just a smoketest that will serve as a reminder if someone accidentally breaks synthesis naming...

oharboe:
@maliberty unrelated CI failure

PR #2169, opened 2024-07-22
makefile: fix circular dependency warning

oharboe:
make dropped a superfluous circular dependency that was recently introduced by accident in the makefile.

Code is now fixed.


maliberty (approved):

oharboe:

PR #2173, opened 2024-07-23
asap7: speed up report_checks when using path groups for large designs

oharboe:
Takes running time of first report_checks on a large test design from 25 minutes to 2 minutes or so, which is the same roughly as on master.

Note that there are no large projects in ORFS that exhibits this behavior as the goal in ORFS is to maximize testing coverage while reining in running times for tests.

### DIFF
--- a/flow/platforms/asap7/constraints.sdc
+++ b/flow/platforms/asap7/constraints.sdc
@@ -86,6 +86,6 @@ set_max_delay $max_delay -from $non_clk_inputs -to [all_outputs]
 # in the histogram in the GUI and also includes these
 # groups in the report
 group_path -name in2reg -from $non_clk_inputs -to [all_registers]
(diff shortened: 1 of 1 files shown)


gadfort:
What is the cause of this speedup? it seems like it might be worth putting together an issue for opensta. if the two are equivalent it should not take 10x to evaluate.

oharboe:
> What is the cause of this speedup? it seems like it might be worth putting together an issue for opensta. if the two are equivalent it should not take 10x to evaluate.

@tspyrou ?

maliberty (approved):

tspyrou:
@gadfort The cause of the speedup is that the number of objects in the path group list goes down from the number of registers to the number of clocks. An enhancement could be made to detect if all registers clocked by a clock are in the from or to list that the registers all be replaced by the clock object.

PR #2182, opened 2024-07-25
makefile: add GUI_ARGS and GUI_SOURCE env vars

oharboe:
These are used to execute a script upon startup and possibly exit immediately.

Previously `echo report_checks | make open_synth` was used, but the problem here is that if the command fails, then the error isn't and shouldn't be propagated to the user as open_synth is a REPL loop that should not propagate the exit code for a failed command, the app should continue to run.

Example uses:

```
make GUI_ARGS=-exit GUI_SOURCE=blah.tcl open_final
make GUI_ARGS=-exit GUI_SOURCE=blah.tcl open_synth
```

Failure to propagate error example:

```
$ echo abc | make open_synth
/home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad -no_init   /home/oyvind/OpenROAD-flow-scripts/flow/scripts/sta-synth.tcl
[deleted]
openroad> abc
invalid command name "abc"
while evaluating abc
openroad> 
$ echo $?
0
```

### DIFF
--- a/docs/user/FlowVariables.md
+++ b/docs/user/FlowVariables.md
@@ -144,6 +144,8 @@ Note:
 | `IO_PLACER_H`         | The metal layer on which to place the I/O pins horizontally (top and bottom of the die).                                                                                         |
 | `IO_PLACER_V`         | The metal layer on which to place the I/O pins vertically (sides of the die).                                                                                                    |
 | `GUI_NO_TIMING`       | Skip loading timing for a faster GUI load.                                                                                                                                       |
+| `GUI_SOURCE` | Source the script. |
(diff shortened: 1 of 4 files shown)


maliberty (approved):

maliberty:
The "Failure to propagate error" seems unrelated to this PR.

oharboe:
> The "Failure to propagate error" seems unrelated to this PR.

It is not a bug, it is how REPL behaves when it is pressed into service for executing commands via pipes.

It demonstrates an explicit execute script capability is needed.

PR #2205, opened 2024-07-30
DONT MERGE restore sta mt

eder-matheus:


maliberty:
Why do you think sta MT is ready to be turned back on?

eder-matheus:
> Why do you think sta MT is ready to be turned back on?

I don't think it's ready. I made some tests locally with sky130hd/microwatt enabling sta MT, and it didn't produce any differences between the runs. I created the PR to test with all public designs in the different pipelines. I will add the DON'T MERGE in the name of the pr.

PR #2207, opened 2024-07-30
DO NOT MERGE test that tests are hooked up

oharboe:


oharboe:
@maliberty @vvbandeira Should have failed

oharboe:
@maliberty @vvbandeira Aha! When the build is running, this is always green, but after the build is complete it is red. Confusing...

I oftentimes look at builds before they are finished because there are some outlier tests that take (it seems) a couple of times longer than the rest of the build...

image

maliberty:
I've seen that too.  It seems to be a Jenkins thing; I'm not sure we have much control over it.

PR #2214, opened 2024-08-02
update mkdir call since we're writing into RESULTS_DIR

jeffng-or:
The change is a no-op, since the RESULTS_DIR is created a couple of times before the 1_synth.sdc file is copied. But, we are writing into the RESULTS_DIR and not the REPORTS_DIR here.


maliberty (approved):
Good eye

PR #2224, opened 2024-08-07
asap7: report_checks -from [all_registers] -to [all_registers] is now fast

oharboe:
`-from [all_clocks]` is not exactly the same as `-from [all_registers]`, but before OpenSTA was optimized for the `-from [all_registers]` case, `-from [all_clocks]` was the best we code do.

I don't think the DRC errors are relevant. mpl2 should have left [a bit of space between the rows](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5522) of the elements.

image


oharboe:
@tspyrou Updated constraints.sdc now that STA is fast for this case.

oharboe:
Needs https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2225 fixed

maliberty (approved):

PR #2230, opened 2024-08-09
Save hours in floorplan with SKIP_REPORT_METRICS=1 enabled

oharboe:


flow/scripts/floorplan.tcl
```diff
@@ -30,12 +30,6 @@ proc report_unused_masters {} {
 
 report_unused_masters
 
-#Run check_setup
-puts "\n=========================================================================="
-puts "Floorplan check_setup"
-puts "--------------------------------------------------------------------------"
-check_setup
```
maliberty, line 37: This appears to be unconditionally removed not just when skipping.

oharboe: Yes, but why is it done twice in the first place?

Removing the, I hope, superfluous and unconditional check_setup and leaving one conditional in report_metrics.tcl is the idea here.

oharboe:
@maliberty I can't see why CI is failing, everything that I can see is green after the two builds complete... Some sort of CI issue?

flow/scripts/floorplan.tcl
tspyrou: @oharboe normally users want to know early in the flow if there is a major sdc problem vs waiting to the end. If you don't check the logs until the end its not needed. If you monitor your log file and want to kill it on a major error then its needed. I think its good to make this an option.

oharboe:
@tspyrou This PR is obsolete with these two fixes https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2231 and https://github.com/The-OpenROAD-Project/OpenROAD/issues/5539

PR #2232, opened 2024-08-09
Makefile: avoid repetition and fix some bugs

oharboe:
In some cases folders needed where not made, like do-5_1_grt


oharboe:
doh! This breaks dependencies.

PR #2236, opened 2024-08-10
global route: now consistent with detailed routing w.r.t. congestion failures

oharboe:
Just like detailed routing, a global route that fails with congestion, is not a build failure(as in exit code non-zero), it is a successful(as in zero exit code) global route that produce reports detailing the problem.

Detailed route will not proceed, if there is global routing congestion

This allows build systems, such as bazel, to create artifacts for global and detailed route, even if the operation had problems, without having know about the semantics between global and detailed route.

Considering that global and detailed route can run for a long time and use a lot of memory, this allows inspecting results on a laptop for a build that ran on a server.


maliberty:
If grt writes 5_1_grt.odb even when there is congestion then this will not work correctly.  Even if the original run stops, a second invocation of make will see the file and assume that step is complete and start drt.  Is that resolved here?

flow/scripts/detail_route.tcl
```diff
@@ -1,4 +1,10 @@
 utl::set_metrics_stage "detailedroute__{}"
+if {[expr [file exists $::env(REPORTS_DIR)/congestion.rpt] && \
+    [file size $::env(REPORTS_DIR)/congestion.rpt] != 0]} {
+  error "Global routing failed, run `make gui_grt` and load $::env(REPORTS_DIR)/congestion.rpt \
```
oharboe: Detailed route does not proceed if global route had congestion failures.

oharboe:
> If grt writes 5_1_grt.odb even when there is congestion then this will not work correctly. Even if the original run stops, a second invocation of make will see the file and assume that step is complete and start drt. Is that resolved here?

Yes.

global route will *not* stop the build in the case of congestion, but, like detailed route, proceed.

There is a [check](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2236/commits/9e52d1833fd128461690ca4e10c0024a83ecfa7a#r1712645504) in detailed route that will stop detailed routing from proceeding if global route failed.

oharboe:
@maliberty How can I create a global routing failure so I can test this easily?

flow/scripts/detail_route.tcl
maliberty: This moves the error to the wrong step.  We already have
```
set result [catch {do_global_route} errMsg]
if {$result != 0} {
  write_db $::env(RESULTS_DIR)/5_1_grt-failed.odb
  error $errMsg
}
```
Why isn't this sufficient?

maliberty:
> @maliberty How can I create a global routing failure so I can test this easily?

Use huge layer derates so there is very little resource.

oharboe:
> > @maliberty How can I create a global routing failure so I can test this easily?
> 
> Use huge layer derates so there is very little resource.

I don't know how to express that... Example?

maliberty:
See https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/4f79a89dca266324a6f219e5a0d698a1916c2fed/flow/scripts/global_route.tcl#L12

 set_global_routing_layer_adjustment $env(MIN_ROUTING_LAYER)-$env(MAX_ROUTING_LAYER) 0.5

Try 0.9

flow/scripts/detail_route.tcl
oharboe: Separation of concerns and also trying to minimize surprises.

I want the responsiblity for providing .odb files and .rpt in the case of global/detailed routing failures and the detection of fatal(no artifacts should be produced) to reside inside ORFS.

The interface I want between ORFS and bazel is that ORFS provides .odb and .rpt file and a non-zero exit code, in which case artifacts are created. With a non-zero exit code, then no artifacts are created. A non-zero exit code is ORFS's way to tell bazel that there was a fatal error and no artifacts should be created.

This mirrors detailed routing's behavior today, I want the same for global routing.

Yes: this means that there needs to be a check not to continue detailed routing when there is a global routing error. This is somewhat surprising, so not ideal, but the error message in detailed routing is easily visible and not drowned out.

One could imagine the same for detailed routing; that there's no point in continuing with the final stage if there the DRC errors in detailed routing are unacceptable(though I don't know what "unacceptable" is for detailed routing).

oharboe:
```
diff --git a/flow/scripts/global_route.tcl b/flow/scripts/global_route.tcl
index fa195632..dd26c7b7 100644
--- a/flow/scripts/global_route.tcl
+++ b/flow/scripts/global_route.tcl
@@ -21,7 +21,7 @@ proc global_route_helper {} {
   if {[info exist ::env(FASTROUTE_TCL)]} {
     source $::env(FASTROUTE_TCL)
   } else {
-    set_global_routing_layer_adjustment $::env(MIN_ROUTING_LAYER)-$::env(MAX_ROUTING_LAYER) 0.5
+    set_global_routing_layer_adjustment $::env(MIN_ROUTING_LAYER)-$::env(MAX_ROUTING_LAYER) 0.99
     set_routing_layers -signal $::env(MIN_ROUTING_LAYER)-$::env(MAX_ROUTING_LAYER)
     if {[info exist ::env(MACRO_EXTENSION)]} {
       set_macro_extension $::env(MACRO_EXTENSION)
```

```
make  GPL_TIMING_DRIVEN=0 SKIP_REPORT_TIMING=1 DESIGN_CONFIG=designs/asap7/ethmac/config.mk 5_1_grt
```

Output(same as detailed routing, a congestion.rpt file is created and .odb, so nothing to do next time I run 5_1_grt):

```
[INFO GRT-0199] Via filling finished.
[INFO GRT-0111] Final number of vias: 948254
[INFO GRT-0112] Final usage 3D: 3321090
[ERROR GRT-0119] Routing congestion too high. Check the congestion heatmap in the GUI and load ./reports/asap7/ethmac/base/congestion.rpt in the DRC viewer.
GRT-0119
Warning: Global routing failed, detailed routing will not proceed, run `make gui_grt` and load ./reports/asap7/ethmac/base/congestion.rpt in DRC viewer to view congestion
Elapsed time: 3:16.84[h:]min:sec. CPU time: user 196.02 sys 0.45 (99%). Peak memory: 1345628KB.
$ echo $?
0
$ wc -l reports/asap7/ethmac/base/congestion.rpt
305492 reports/asap7/ethmac/base/congestion.rpt
$ ls -lrt results/asap7/ethmac/base/5_1_grt.odb 
-rw-rw-r-- 1 oyvind oyvind 35003425 Aug 10 16:30 results/asap7/ethmac/base/5_1_grt.odb
$ make GPL_TIMING_DRIVEN=0 DESIGN_CONFIG=designs/asap7/ethmac/config.mk 5_1_grt
make: Nothing to be done for '5_1_grt'.
```

Now, when I run:

```
make  GPL_TIMING_DRIVEN=0 SKIP_REPORT_TIMING=1 DESIGN_CONFIG=designs/asap7/ethmac/config.mk route
```

I get:

```
[deleted]
Running detail_route.tcl, stage 5_3_route
[INFO ORD-0030] Using 20 thread(s).
Error: detail_route.tcl, 6 Global routing failed, run `make gui_grt` and load ./reports/asap7/ethmac/base/congestion.rpt  in DRC viewer to view congestion
Elapsed time: 0:00.05[h:]min:sec. CPU time: user 0.04 sys 0.00 (96%). Peak memory: 77432KB.
make[1]: *** [Makefile:813: do-5_3_route] Error 1
make: *** [Makefile:813: results/asap7/ethmac/base/5_3_route.odb] Error 2
```

flow/scripts/detail_route.tcl
maliberty: That might be good for Bazel but it runs against how make works.  An artifact is a sign of success not failure in make.

oharboe: I dont think so: this is how detailed route works.

maliberty: It is how gpl & dpl work.  drt is different as people often manually fix drcs left by drt.

oharboe: does dpl and other stages have similarly useful reports in case of failure, other than the log?

maliberty: You can use -report_file_name with detailed_place though it looks like we don't currently in the flow.  I'm not sure why it matters.

oharboe:
After introducing the `GENERATE_ARTIFACTS_ON_FAILURE` option(set by Bazel, off by default for Makefile) surprises should be minimized for both Makefile and Bazel users. Fewer options is generally better, simplifies code and testing, but the difference in expectations is large enough that it seems worth it to have an option here.

GENERATE_ARTIFACTS_ON_FAILURE=0 (default), Makefile use-case, same as before:

```
$ make GENERATE_ARTIFACTS_ON_FAILURE=0 GPL_TIMING_DRIVEN=0 SKIP_REPORT_METRICS=1 DESIGN_CONFIG=designs/asap7/ethmac/config.mk do-5_1_grt 
Running global_route.tcl, stage 5_1_grt
[deleted]
[ERROR GRT-0119] Routing congestion too high. Check the congestion heatmap in the GUI and load ./reports/asap7/ethmac/base/congestion.rpt in the DRC viewer.
Error: global_route.tcl, 143 GRT-0119
Elapsed time: 3:13.24[h:]min:sec. CPU time: user 192.72 sys 0.46 (99%). Peak memory: 1344560KB.
make: *** [Makefile:830: do-5_1_grt] Error 1
$ echo $?
2
```

GENERATE_ARTIFACTS_ON_FAILURE=1 (not default), Bazel use-case:

```
$ make GENERATE_ARTIFACTS_ON_FAILURE=1 GPL_TIMING_DRIVEN=0 SKIP_REPORT_METRICS=1 DESIGN_CONFIG=designs/asap7/ethmac/config.mk do-5_1_grt
Running global_route.tcl, stage 5_1_grt
[deleted]
[ERROR GRT-0119] Routing congestion too high. Check the congestion heatmap in the GUI and load ./reports/asap7/ethmac/base/congestion.rpt in the DRC viewer.
Elapsed time: 3:09.30[h:]min:sec. CPU time: user 188.63 sys 0.49 (99%). Peak memory: 1344872KB.
$ echo $?
0
$ make GENERATE_ARTIFACTS_ON_FAILURE=1 GPL_TIMING_DRIVEN=0 SKIP_REPORT_METRICS=1 DESIGN_CONFIG=designs/asap7/ethmac/config.mk do-5_2_fillcell
Running fillcell.tcl, stage 5_2_fillcell
[INFO ORD-0030] Using 20 thread(s).
Error: fillcell.tcl, 5 Global routing failed, run `make gui_grt` and load ./reports/asap7/ethmac/base/congestion.rpt  in DRC viewer to view congestion
Elapsed time: 0:00.04[h:]min:sec. CPU time: user 0.04 sys 0.00 (100%). Peak memory: 77228KB.
make: *** [Makefile:834: do-5_2_fillcell] Error 1
```

flow/scripts/detail_route.tcl
oharboe: Just trying to understand where there are frequently useful reports on failure. More reports on place failures could be a useful future improvement.

flow/scripts/fillcell.tcl
```diff
@@ -1,4 +1,8 @@
-utl::set_metrics_stage "globalroute__{}"
+if {[expr [file exists $::env(REPORTS_DIR)/congestion.rpt] && \
```
maliberty, line 1: FYI, It is likely that this step will get moved after drt in the future.

flow/scripts/global_route.tcl
```diff
@@ -1,120 +1,143 @@
...
-if {[info exist env(DONT_USE_CELLS)]} {
-  set_dont_use $::env(DONT_USE_CELLS)
-}
+  set result [catch {do_global_route} errMsg]
 
-if { ![info exists ::env(SKIP_INCREMENTAL_REPAIR)] } {
-  if {[info exist ::env(DETAILED_METRICS)]} {
-    report_metrics 5 "global route pre repair design"
-  }
+  write_guides $::env(RESULTS_DIR)/route.guide
```
maliberty: Depending on the nature of the error this may not make sense.  It would be better to do it only when you don't write a -failed.odb

maliberty (changes requested):

flow/scripts/fillcell.tcl
oharboe: Yes... this support is going to require some maintenance as the flow changes, no way around that. Hopefully not too bad and perhaps we can find some nice way.

flow/scripts/global_route.tcl
oharboe: How is the user-experience diminished from having this information in the case of a failed global route?

maliberty: The same reason we write -failed.odb to indicate it isn't valid data.  Putting invalid data in a same place as valid data is confusing at best.

maliberty (approved):

PR #2248, opened 2024-08-14
Add power simulation support and Verilator to the CI pipeline

ivysochyn:
# Add power simulation support and Verilator to the CI pipeline

This PR enhances the test suite by introducing `simulate` and `power` build targets for designs that include a `power.tcl` file. The `power.tcl` file sets the pass/fail criteria for power simulations, ensuring that tests are run only for designs with these predefined criteria.

This PR modifies the following:

* `flow/test/test_helper.sh` - Updated to execute power simulation for designs that have a `power.tcl` file present
* `docker/Dockerfile.builder` - Adds build and installation instructions for the Verilator
* `etc/DependencyInstaller.sh` - Updates dependencies to cover those of Verilator
* `.gitmodules` - Adds Verilator as a sub-module located under the `tools/verilator` path

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -11,6 +11,7 @@ fi
 
 # package versions
 klayoutVersion=0.28.8
+verilatorVersion=5.026
 
 _versionCompare() {
     local a b IFS=. ; set -f
@@ -33,6 +34,23 @@ _installCommon() {
         pip3 install --no-cache-dir -U $pkgs
     else
         pip3 install --no-cache-dir --user -U $pkgs
+    fi
+
+    baseDir=$(mktemp -d /tmp/DependencyInstaller-orfs-XXXXXX)
+
+    # Install Verilator
+    verilatorPrefix=`realpath ${PREFIX:-"/usr/local"}`
+    if [[ ! -x ${verilatorPrefix}/bin/verilator ]]; then
+        pushd $baseDir
+            git clone --depth=1 -b "v$verilatorVersion" https://github.com/verilator/verilator.git
+            pushd verilator
+                autoconf
+                ./configure --prefix "${verilatorPrefix}"
+                make -j`nproc`
+                make install
+            popd
+            rm -r verilator
+        popd
     fi
 }
 
@@ -71,19 +89,28 @@ _installUbuntuPackages() {
     export DEBIAN_FRONTEND="noninteractive"
     apt-get -y update
     apt-get -y install --no-install-recommends \
+        bison \
         curl \
+        flex \
+        help2man \
+        libfl-dev \
+        libfl2 \
+        libgoogle-perftools-dev \
         libqt5multimediawidgets5 \
+        libqt5opengl5 \
         libqt5svg5-dev \
         libqt5xmlpatterns5-dev \
-        libqt5opengl5 \
         libz-dev \
+        perl \
         python3-pip \
         python3-venv \
         qtmultimedia5-dev \
         qttools5-dev \
         ruby \
         ruby-dev \
-        time
+        time \
+        zlib1g \
+        zlib1g-dev
 
     # install KLayout
     if _versionCompare $1 -ge 23.04; then
@@ -215,6 +242,7 @@ while [ "$#" -gt 0 ]; do
             ;;
         -prefix=*)
             OR_INSTALLER_ARGS="${OR_INSTALLER_ARGS} $1"
+            PREFIX=${1#*=}
             ;;
         *)
             echo "unknown option: ${1}" >&2
--- a/flow/test/test_helper.sh
+++ b/flow/test/test_helper.sh
@@ -2,7 +2,7 @@
 
 set -eoux pipefail
 
-cd "$(dirname $(readlink -f $0))/../"
+cd "$(dirname "$(readlink -f "$0")")/../"
 
 # Setting args (and setting default values for testing)
 DESIGN_NAME=${1:-gcd}
@@ -13,20 +13,20 @@ if [ $# -eq 4 ]; then
 fi
 DESIGN_CONFIG=./designs/$PLATFORM/$DESIGN_NAME/$CONFIG_MK
 LOG_FILE=./logs/$PLATFORM/$DESIGN_NAME.log
-mkdir -p ./logs/$PLATFORM
+mkdir -p "./logs/$PLATFORM"
 
 __make="make DESIGN_CONFIG=$DESIGN_CONFIG"
-if [ ! -z "${FLOW_VARIANT+x}" ]; then
+if [ -n "${FLOW_VARIANT+x}" ]; then
   __make+=" FLOW_VARIANT=$FLOW_VARIANT"
 fi
 
-mkdir -p $(dirname $LOG_FILE)
-$__make clean_all clean_metadata 2>&1 | tee $LOG_FILE
+mkdir -p "$(dirname "$LOG_FILE")"
+$__make clean_all clean_metadata 2>&1 | tee "$LOG_FILE"
 
 # turn off abort on error so we can always capture the result
 set +e
 
-$__make finish metadata 2>&1 | tee -a $LOG_FILE
+$__make finish metadata 2>&1 | tee -a "$LOG_FILE"
 
 # Save the return code to return as the overall status after we package
 # the results
@@ -36,21 +36,21 @@ if [ -z "${PRIVATE_DIR+x}" ]; then
   PRIVATE_DIR="../../private_tool_scripts"
 fi
 
-if [ -f "$PRIVATE_DIR/openRoad/private.mk" ] && [ ! -z ${SAVE_TO_DB+x} ]; then
+if [ -f "$PRIVATE_DIR/openRoad/private.mk" ] && [ -n "${SAVE_TO_DB+x}" ]; then
   $__make save_to_metrics_db
   ret=$(( ret + $? ))
 fi
 
-if [ -f "$PRIVATE_DIR/util/utils.mk" ] && [ ! -z ${RUN_CALIBRE+x} ]; then
+if [ -f "$PRIVATE_DIR/util/utils.mk" ] && [ -n "${RUN_CALIBRE+x}" ]; then
   $__make calibre_drc
   ret=$(( ret + $? ))
   $__make convert_calibre
   ret=$(( ret + $? ))
-  if [ ! -z ${SAVE_TO_DB+x} ]; then
+  if [ -n "${SAVE_TO_DB+x}" ]; then
     $__make save_to_drc_db
     ret=$(( ret + $? ))
   fi
-  if [ ! -z ${CHECK_DRC_DB+x} ]; then
+  if [ -n "${CHECK_DRC_DB+x}" ]; then
     $__make check_drc_db
     ret=$(( ret + $? ))
   fi
@@ -59,8 +59,8 @@ fi
 # Only enabled abort on error at the end to allow script to reach make issue
 set -e
 
-if [ ! -z ${MAKE_ISSUE+x} ]; then
-  $__make final_report_issue 2>&1 | tee -a $LOG_FILE
+if [ -n "${MAKE_ISSUE+x}" ]; then
+  $__make final_report_issue 2>&1 | tee -a "$LOG_FILE"
 fi
 
 exit $ret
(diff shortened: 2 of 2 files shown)


oharboe (changes requested):
Why can't this be part of ./setup.sh and the docker image, just like other tools provided by the OS?

Why should it be a submodule?

There are LOTS of dependencies in ORFS that are not submodules.

flow/test/test_helper.sh
```diff
@@ -2,7 +2,7 @@
 
 set -eoux pipefail
 
-cd "$(dirname $(readlink -f $0))/../"
+cd "$(dirname "$(readlink -f "$0")")/../"
```
oharboe, line 6: split into two commits: one for fixes, one for features, for simpler review.

flow/test/test_helper.sh
```diff
@@ -59,8 +60,13 @@ fi
...
 
-if [ ! -z ${MAKE_ISSUE+x} ]; then
-  $__make final_report_issue 2>&1 | tee -a $LOG_FILE
+if [ -n "${MAKE_ISSUE+x}" ]; then
+  $__make final_report_issue 2>&1 | tee -a "$LOG_FILE"
+fi
+
+if [ $ret -eq 0 ] && [ -f "$POWER_TCL_FILE" ]; then
+  $__make simulate power 2>&1 | tee -a "$LOG_FILE"
+  ret=$?
```
vvbandeira:
```suggestion
  ret=$(( ret + $? ))
```

vvbandeira (changes requested):
Avoid overwriting previous return values.

etc/DependencyInstaller.sh
```diff
@@ -34,6 +35,23 @@ _installCommon() {
     else
         pip3 install --no-cache-dir --user -U $pkgs
     fi
+
+    baseDir=$(mktemp -d /tmp/DependencyInstaller-orfs-XXXXXX)
+
+    # Install Verilator
```
oharboe, line 41: Nice. I think that we're asking for less trouble by building ourselves than to use the version that comes with the OS distribution.

oharboe (approved):
Nit, better to split commit into a fixes and a feature commit to simplify review.

etc/DependencyInstaller.sh
```diff
@@ -34,6 +35,23 @@ _installCommon() {
...
+    verilogPrefix=`realpath ${PREFIX:-"/usr/local"}`
```
maliberty: `verilatorPrefix` makes more sense to me

etc/DependencyInstaller.sh
```diff
@@ -34,6 +35,23 @@ _installCommon() {
     else
         pip3 install --no-cache-dir --user -U $pkgs
     fi
+
+    baseDir=$(mktemp -d /tmp/DependencyInstaller-orfs-XXXXXX)
```
maliberty, line 39: Why is this needed?  It is already set in the script elsewhere.

etc/DependencyInstaller.sh
```diff
@@ -71,19 +89,28 @@ _installUbuntuPackages() {
     export DEBIAN_FRONTEND="noninteractive"
     apt-get -y update
     apt-get -y install --no-install-recommends \
+        bison \
         curl \
+        flex \
+        help2man \
+        libfl-dev \
+        libfl2 \
+        libgoogle-perftools-dev \
```
maliberty, line 98: Does Verilator require this?

etc/DependencyInstaller.sh
```diff
@@ -215,6 +242,7 @@ while [ "$#" -gt 0 ]; do
             ;;
         -prefix=*)
             OR_INSTALLER_ARGS="${OR_INSTALLER_ARGS} $1"
+            PREFIX=${1#*=}
```
maliberty, line 245: What does this do?

flow/test/test_helper.sh
```diff
@@ -59,8 +60,13 @@ fi
...
 set -e
 
-if [ ! -z ${MAKE_ISSUE+x} ]; then
-  $__make final_report_issue 2>&1 | tee -a $LOG_FILE
+if [ -n "${MAKE_ISSUE+x}" ]; then
+  $__make final_report_issue 2>&1 | tee -a "$LOG_FILE"
+fi
+
+if [ $ret -eq 0 ] && [ -f "$POWER_TCL_FILE" ]; then
+  $__make simulate power 2>&1 | tee -a "$LOG_FILE"
```
maliberty: Where is this target defined?

maliberty (changes requested):

flow/test/test_helper.sh
eszpotanski: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/designs/asap7/mock-array/config.mk#L50

etc/DependencyInstaller.sh
eszpotanski: It's optional, but "should be installed for good performance" ([Verilator install prerequisites](https://verilator.org/guide/latest/install.html#install-prerequisites))

maliberty:
Is there a reason to make this narrowly about a single design rather than a framework any design could use?
What is the runtime overhead of using Verilator?  Why not simply capture a vcd file?

oharboe:
> Is there a reason to make this narrowly about a single design rather than a framework any design could use? What is the runtime overhead of using Verilator? Why not simply capture a vcd file?

- Any design can add simulation if they want under this framework.
- A .vcd file has to be regenerated via simulation after every time the ORFS flow has run. The test bench that exercises the design is design specific.
- What do you mean "overhead" of Verilator? Something has to run the simulation, Verilator is doing that. Are you asking how long this simulation stage takes in addition to the normal tests?

flow/test/test_helper.sh
```diff
@@ -59,8 +60,13 @@ fi
 # Only enabled abort on error at the end to allow script to reach make issue
 set -e
 
-if [ ! -z ${MAKE_ISSUE+x} ]; then
-  $__make final_report_issue 2>&1 | tee -a $LOG_FILE
+if [ -n "${MAKE_ISSUE+x}" ]; then
+  $__make final_report_issue 2>&1 | tee -a "$LOG_FILE"
+fi
+
+if [ $ret -eq 0 ] && [ -f "$POWER_TCL_FILE" ]; then
```
oharboe: Here it would be preferable to check if the make file with the current design has a "simulate" and "power" target and let the design implement these targets however it wants.

oharboe (commented):
Suggestion to check for existance of "power" and "simulate" targets per design.

flow/test/test_helper.sh
eszpotanski: Commit has been split

flow/test/test_helper.sh
eszpotanski: Suggestion applied

etc/DependencyInstaller.sh
eszpotanski: changed to `verilatorPrefix`

flow/test/test_helper.sh
```diff
@@ -12,6 +12,7 @@ if [ $# -eq 4 ]; then
   FLOW_VARIANT=$4
 fi
 DESIGN_CONFIG=./designs/$PLATFORM/$DESIGN_NAME/$CONFIG_MK
+POWER_TCL_FILE=./designs/$PLATFORM/$DESIGN_NAME/power.tcl
```
oharboe: Deleted line above, unused.

etc/DependencyInstaller.sh
eszpotanski: I believe that `DependencyInstaller` from OpenROAD has this variable set once, but in ORFS each function has its own definitions of `baseDir`

oharboe (changes requested):
nit

etc/DependencyInstaller.sh
eszpotanski: it assigns everything after `=` to the `PREFIX` variable, which is later used as a Verilator prefix installation

flow/test/test_helper.sh
eszpotanski: changed to separately checking whether make target exist

flow/test/test_helper.sh
eszpotanski: removed

oharboe:
On my laptop:

- 500s `make DESIGN_CONFIG=designs/asap7/mock-array/config.mk `
- 256s `make DESIGN_CONFIG=designs/asap7/mock-array/config.mk simulate`
- 17s `make DESIGN_CONFIG=designs/asap7/mock-array/config.mk power`

Simulate can be sped up to 60s by this change, but I'd want to ask @vvbandeira whether that is a good idea... If there are lots of other processes on that CI server, then better to run in serial? It would improve the local user experience though.

```
$ git diff
diff --git a/flow/designs/asap7/mock-array/simulate.sh b/flow/designs/asap7/mock-array/simulate.sh
index 0796d6ec..8778b89d 100755
--- a/flow/designs/asap7/mock-array/simulate.sh
+++ b/flow/designs/asap7/mock-array/simulate.sh
@@ -27,6 +27,6 @@ verilator -Wall --cc \
   --exe \
   ../../../designs/src/mock-array/simulate.cpp
 
-make -C obj_dir -f VMockArray.mk
+make -j -C obj_dir -f VMockArray.mk
 
 obj_dir/VMockArray
```

vvbandeira:
> On my laptop:
> 
> * 500s `make DESIGN_CONFIG=designs/asap7/mock-array/config.mk `
> * 256s `make DESIGN_CONFIG=designs/asap7/mock-array/config.mk simulate`
> * 17s `make DESIGN_CONFIG=designs/asap7/mock-array/config.mk power`
> 
> Simulate can be sped up to 60s by this change, but I'd want to ask @vvbandeira whether that is a good idea... If there are lots of other processes on that CI server, then better to run in serial? It would improve the local user experience though.
> 
> ```
> $ git diff
> diff --git a/flow/designs/asap7/mock-array/simulate.sh b/flow/designs/asap7/mock-array/simulate.sh
> index 0796d6ec..8778b89d 100755
> --- a/flow/designs/asap7/mock-array/simulate.sh
> +++ b/flow/designs/asap7/mock-array/simulate.sh
> @@ -27,6 +27,6 @@ verilator -Wall --cc \
>    --exe \
>    ../../../designs/src/mock-array/simulate.cpp
>  
> -make -C obj_dir -f VMockArray.mk
> +make -j -C obj_dir -f VMockArray.mk
>  
>  obj_dir/VMockArray
> ```

Each design runs on its own VM, so it should be fine to run in parallel as long as they do not consume resources to the point of starving themselves.

oharboe:
@maliberty mock-array finishes before other slower designs, so I don't see a reason to complicate this PR by speeding up `make simulate`

image

maliberty:
I'm unclear why we are putting this in flow/test/test_helper.sh and not in the flow proper.  Users don't use test_helper.sh, they use make or bazel to run flow steps.

oharboe:
> I'm unclear why we are putting this in flow/test/test_helper.sh and not in the flow proper. Users don't use test_helper.sh, they use make or bazel to run flow steps.

What is "in the flow proper"?

The idea of putting it in test_helper.sh is simply to get CI to pick it up.

Users would run "make simulate power". The reason config.mk of the design implements "simulate" and "power" is that it is a go-no go test where power is optional after simulate and it is highly design specific, so I'm unsure what code can be shared, at least until I see a couple of designs doing the same thing.

I think it makes sense to start with one design under CI for power, then add two three more over time to see what is generic and what is specific to the design. If we try to design something generic without having a few examples, we're sure to have to throw away our overcomplicated general design once we have a few examples as I'm sure they won't fit(because we design a general system with no examples, which is an exercise in futility...).

Can you suggest an alternative way to do this or is this a good 'nuf start?

We have:

- "make power simulate" under CI, if they exist
- we allow users to run the tests locally

maliberty:
> What is "in the flow proper"?

As it stands we will do report_power and analyze_power_grid in final_report.tcl without any vcd.  Then we will do report_power again with a vcd only if the user does some extra manual steps.  We will never do analyze_power_grid with the vcd.  It seems better to handle this in final_report.tcl.

oharboe:
I see. Can you take delivery of this and rework it, we dont have the domain knowledge to integrate....

maliberty:
My comments are to be addressed in a follow on PR

PR #2256, opened 2024-08-16
openroad: bump

oharboe:


maliberty (approved):

oharboe:
@maliberty PDN is broken with latest OpenROAD

@jeffng-or FYI

maliberty:
@gadfort any thoughts on pdn in master?

gadfort:
I would need a proper testcase to evaluate.
Why wouldn't the PDN check_power_grid call have caught this sooner? Is that disabled?

jeffng-or:
Let me see if I can create a test case using MegaBoom. I have a 2_5_floorplan_tapcell.odb which gets passed into pdn.tcl that I think repro's the problem.

Edit: actually, it's the same test case as what's listed in https://github.com/The-OpenROAD-Project/OpenROAD/issues/5585.

Also, the current pdn.tcl has check_power_grid disabled.

gadfort:
@jeffng-or I looked at that and the issue there is the blocks cannot be connected to power.
I don't know when the grid was changed for those blocks, but as far as I can tell if the blocks have their power grid go to M6 and the max level on MegaBoom is M6 there is no way for PDN to connect to those.

jeffng-or:
Yeah, I looked back through those comments and just added an update to https://github.com/The-OpenROAD-Project/OpenROAD/issues/5585.

gadfort:
@jeffng-or @oharboe the testcase referenced in https://github.com/The-OpenROAD-Project/OpenROAD/issues/5585 appears to be an impossible problem, so there is nothing for PDN to do (not sure how that was passing before, unless multiple things were changed in parallel). The issue in https://github.com/The-OpenROAD-Project/OpenROAD/issues/5605 has a file that cannot be accessed by me so I cannot view it to see what could be done (the ORFS pipeline also doesn't seem to have any useful artifacts for this).
It's entirely possible that the change I made to reduce the file of the odbs/def created an issue ([this](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2258) could be related, but also no test case), but without a testcase I can access and of reasonable size I cannot help.

oharboe:
@maliberty Latest OpenROAD master as of writing

oharboe:
@maliberty It passed... OK to merge?

PR #2257, opened 2024-08-16
mock-array: speed up simulate

oharboe:


vvbandeira (approved):

flow/designs/asap7/mock-array/simulate.sh
```diff
@@ -27,6 +27,6 @@ verilator -Wall --cc \
   --exe \
   ../../../designs/src/mock-array/simulate.cpp
 
-make -C obj_dir -f VMockArray.mk
+make -j -C obj_dir -f VMockArray.mk
```
maliberty, line 30: `-j $(NUM_CORES)` would be more consistent with the rest of the flow and gives the control

PR #2261, opened 2024-08-18
Test helper less verbose

oharboe:
@maliberty is this better? Quick stab at reducing verbosity of test_helper...


maliberty:
All tests fail with this change
```
++ grep -E 'simulate:|power:'
+ TARGETS=
script returned exit code 1
```

Issue #2264, opened 2024-08-19
asap7/mock-array: Re-enable simulate and power


jeffng-or:
### Description

Debug and implement fixes required to reenable simulate and power steps for mockarray that were disabled in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/f2f37027d1369c52a7c4d35cf251769f071abd87.

### Suggested Solution

Review CI failure and determine next steps

oharboe:
Added some post-merge review comments to be addressed.

PR #2267, opened 2024-08-19
Re-worked mock-array simulate and power steps to work with CI

jeffng-or:
Addresses https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2264

Updated the mock-array simulate and power steps so that they read/write VCD and associated files from the flow/results directory vs. the flow/designs directory. The new directory paths are:

- "post" directory - results/asap7/mock-array/base/verilator/post
- "obj_dir" directory - results/asap7/mock-array/base/verilator/obj

The simulate.sh still allows you to call it from an arbitrary directory, but it will use FLOW_HOME to figure out the location of the files that it needs to access. The simulate.cpp file has been updated to write the VCD file to the results/asap7/mock-array/base directory, which is also where the power.tcl file now looks for the VCD file.

Cleaned up some .gitignores as a result of the changes.

### DIFF
--- a/flow/designs/asap7/mock-array/power.tcl
+++ b/flow/designs/asap7/mock-array/power.tcl
@@ -20,7 +20,7 @@ for {set x 0} {$x < 8} {incr x} {
 
 report_parasitic_annotation
 report_power
-read_power_activities -scope TOP/MockArray -vcd designs/src/mock-array/MockArrayTestbench.vcd
+read_power_activities -scope TOP/MockArray -vcd $::env(RESULTS_DIR)/MockArrayTestbench.vcd
 report_power
 
 # FIXME add an automated test to check that the top-level power is
--- a/flow/designs/asap7/mock-array/simulate.sh
+++ b/flow/designs/asap7/mock-array/simulate.sh
@@ -1,20 +1,34 @@
 #!/usr/bin/env bash
+#
+# Executes Verilator to generate a VCD file based on simulation
+# simulate.cpp is set up to write the VCD file into the results directory
+#
 set -ex
 
-# allow this script to be invoked from any folder
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-BASE=$DIR/../..
 
-cd $DIR
+RESULTS_DIR="$FLOW_HOME/results/asap7/mock-array/base"
+OBJ_DIR="$RESULTS_DIR/verilator/obj"
+POST_DIR="$RESULTS_DIR/verilator/post"
 
-cd ../../src/mock-array
-cp ../../../results/asap7/mock-array/base/6_final.v post/MockArrayFinal.v
-cp ../../../results/asap7/mock-array_Element/base/6_final.v post/MockArrayElementFinal.v
+# If you run outside of the Makefile system, then PLATFORM_DIR won't be set
+if [[ -z "$PLATFORM_DIR" ]]; then
+    PLATFORM_DIR="$FLOW_HOME/platforms/asap7"
+fi
 
+# Make sure the output directories are created
+mkdir -p $OBJ_DIR
+mkdir -p $POST_DIR
+
+# Copy Verilog files used for simulation to post dir in the objects area
+cp $FLOW_HOME/results/asap7/mock-array/base/6_final.v $POST_DIR/MockArrayFinal.v
+cp $FLOW_HOME/results/asap7/mock-array_Element/base/6_final.v $POST_DIR/MockArrayElement.v
+
+# Run simulation and have Verilator write the output files to the objects area
 verilator -Wall --cc \
   -Wno-DECLFILENAME \
   -Wno-UNUSEDSIGNAL \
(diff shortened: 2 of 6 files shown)


jeffng-or:
@oharboe , @maliberty FYI.

maliberty (approved):

flow/designs/asap7/mock-array/power.tcl
```diff
@@ -20,7 +20,7 @@ for {set x 0} {$x < 8} {incr x} {
 
 report_parasitic_annotation
 report_power
-read_power_activities -scope TOP/MockArray -vcd designs/src/mock-array/MockArrayTestbench.vcd
+read_power_activities -scope TOP/MockArray -vcd $::env(RESULTS_DIR)/MockArrayTestbench.vcd
```
oharboe, line 23: @maliberty @jeffng-or What is the difference between the RESULTS_DIR and REPORTS_DIR? What goes where?

.lib, .lef, .odb, .sdc seems obviously "results". Less sure about .vcd.

Arbitrary? I think we should minimize surprises. I'm somewhat surprised to see a .vcd file as a "result" and not a "report". It is somewhat like a "log", but perhaps more like a "report" since it is structured very specifically.

flow/designs/asap7/mock-array/simulate.sh
```diff
@@ -1,32 +1,48 @@
...
+# simulate.cpp is set up to write the VCD file into the results directory
+#
 set -ex
 
-# allow this script to be invoked from any folder
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-BASE=$DIR/../..
 
-cd $DIR
+RESULTS_DIR="$FLOW_HOME/results/asap7/mock-array/base"
```
oharboe, line 9: If we have FLOW_HOME, we should have the other variables, why are we overriding them here?

flow/designs/src/mock-array/simulate.cpp
```diff
@@ -2,6 +2,20 @@
 #include "verilated.h"
 #include <verilated_vcd_c.h>
 
+/**
+ * Returns the VCD output path under the results directory
+ **/
+static std::string getVCDFilePath() {
+
+    std::string vcd_file_name = "results/asap7/mock-array/base/MockArrayTestbench.vcd";
```
oharboe, line 10: use RESULTS_DIR variable

oharboe (commented):
Make it a bit more ORFSesque.

flow/designs/asap7/mock-array/power.tcl
maliberty: Its a bit arbitrary but I think of a report as something a human would inspect in some way.  I doubt anyone wants to read a vcd file.  A log is more of a record of the tools output while running.

oharboe: a .rpt file of many megabytes is best viewed in OpenROAD GUI

flow/designs/asap7/mock-array/simulate.sh
jeffng-or: It looks like I wasn't consistent in handling variables that are typically set by the Makefile. I set RESULTS_DIR regardless, but did a check of PLATFORM_DIR.

The question is: Do we ever envision for this script where we've sourced env.sh, but are running outside of the Makefile system?

If the answer is Yes, then RESULTS_DIR and PLATFORM_DIR need to be checked and set in the bash since they wouldn't have been set by directly running the script. If the answer is No, then I can remove setting RESULTS_DIR and PLATFORM_DIR within the script.

flow/designs/src/mock-array/simulate.cpp
jeffng-or: Will update in my next PR

flow/designs/asap7/mock-array/simulate.sh
oharboe: No, it will always be called from the Makefile context, so no need to redefine variables already set in the Makefile

flow/designs/asap7/mock-array/power.tcl
maliberty: That is still human inspection.  There is no GUI for a vcd file and I don't see one that would make sense.  It is just inter-tool data.

oharboe: Sure there is a GUI: gtkwave

maliberty: Fair point - I don't really care that much.  If you want to PR a change to report that's fine.

oharboe: I'm just collecting information for now, no strong opinion, just curious how we ended up with two and if we had a clear objective selection criterea. It is a bit fuzzy, that's fine.

maliberty: It predates me so I'm just giving my opinion on how it might make sense.

PR #2269, opened 2024-08-20
Updated env var usage in Verilator simulation flow per PR feedback

jeffng-or:
Incorporated feedback from https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2264.

Assumption is that simulate.sh will always be called from within Makefile system, so we can assume that the relevant env vars have been set and don't need to check them.

### DIFF
--- a/flow/designs/asap7/mock-array/simulate.sh
+++ b/flow/designs/asap7/mock-array/simulate.sh
@@ -5,15 +5,8 @@
 #
 set -ex
 
-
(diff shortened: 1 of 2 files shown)


maliberty (approved):

maliberty:
@oharboe I'll wait to merge for your approval as it is your feedback being addressed

oharboe (approved):

Issue #2160, opened 2024-07-19
Create codespaces does not work


oharboe:
### Subject

[Stage]: Other. Please describe below.

### Describe the bug

Create codespace on master. The default should be the ubuntu gui image:

![image](https://github.com/user-attachments/assets/88acd756-4ec5-4256-bd9d-db39920153eb)

Fails with:

```
====================================== ERROR ====================================
2024-07-19 05:03:32.771Z: Failed to create container.
=================================================================================
2024-07-19 05:03:32.771Z: Error: Command failed: docker inspect --type image ghcr.io/the-openroad-project/openroad-flow-scripts:lxqt
2024-07-19 05:03:32.774Z: Error code: 1302 (UnifiedContainersErrorFatalCreatingContainer)

====================================== ERROR ====================================
2024-07-19 05:03:32.782Z: Container creation failed.
=================================================================================
```

### Expected Behavior

No failure.

### Environment

```shell
v3.0-941-g3395328e
```

### To Reproduce

See above.

### Relevant log output

```shell
2024-07-19 05:02:53.142Z: Host information
2024-07-19 05:02:53.142Z: ----------------
2024-07-19 05:02:53.142Z: OS: Ubuntu 22.04.4 LTS (stable release)
2024-07-19 05:02:53.142Z: Image details: https://github.com/github/codespaces-host-images/blob/main/README.md
2024-07-19 05:02:53.142Z: ----------------

=================================================================================
2024-07-19 05:02:53.142Z: Configuration starting...
2024-07-19 05:02:53.165Z: Cloning...
2024-07-19 05:02:53.200Z: $ git -C "/var/lib/docker/codespacemount/workspace" clone --branch "master" --depth 1 https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts "/var/lib/docker/codespacemount/workspace/OpenROAD-flow-scripts"
2024-07-19 05:02:53.206Z: Cloning into '/var/lib/docker/codespacemount/workspace/OpenROAD-flow-scripts'...
2024-07-19 05:03:22.745Z: Updating files:  23% (533/2222)
2024-07-19 05:03:22.762Z: Updating files:  24% (534/2222)
2024-07-19 05:03:22.772Z: Updating files:  25% (556/2222)
2024-07-19 05:03:22.776Z: Updating files:  26% (578/2222)
2024-07-19 05:03:22.806Z: Updating files:  27% (600/2222)
[...]

PR #2272, opened 2024-08-21
Fix codespace image

vvbandeira:


vvbandeira:
@maliberty @oharboe 
This should close #2160 as I am changing the image. However, I could not create a codespace in my account due to a lack of disk space. I believe that the free version of the codespace machine is very limited. 
```
docker: failed to register layer: write /OpenROAD-flow-scripts/tools/OpenROAD/test/sky130hd/sky130_fd_sc_hd__tt_025C_1v80.lib: no space left on device.
```
I think that to use this feature properly a user must either pay GitHub for a bigger VM or host their own.

oharboe (approved):
The ask in the github issue here was to hook up it to CI, which is done.

Improving the experience and the docker image is followon work not required by the currently open issue.

I tried it and I can create VMs and connect to them.

True: the smallest github codespace configuration is quite limited, so for that to work well, would certainly take some work.

PR #2282, opened 2024-08-23
Less iffy makefile for flattening policy

oharboe:
No new or removed features, just moving conditional code into .tcl.

Introducing a do-synth-flatten-policy stage that figures out which modules to keep in flattening, with SYNTH_HIERARCHICAL=0, all Verilog modules are flattened.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -205,11 +205,10 @@ include $(PLATFORM_DIR)/config.mk
 
 # Enables hierarchical yosys
 export SYNTH_HIERARCHICAL ?= 0
-export SYNTH_STOP_MODULE_SCRIPT = $(OBJECTS_DIR)/mark_hier_stop_modules.tcl
-ifeq ($(SYNTH_HIERARCHICAL), 1)
+export SYNTH_STOP_MODULE_SCRIPT = $(RESULTS_DIR)/flatten_policy.tcl
 export HIER_REPORT_SCRIPT = $(SCRIPTS_DIR)/synth_hier_report.tcl
 export MAX_UNGROUP_SIZE ?= 0
-endif
+
 # Enables Re-synthesis for area reclaim
 export RESYNTH_AREA_RECOVER ?= 0
 export RESYNTH_TIMING_RECOVER ?= 0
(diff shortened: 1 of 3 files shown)


flow/designs/asap7/aes-block/config.mk
```diff
@@ -30,3 +30,6 @@ export PWR_NETS_VOLTAGES      =
 
 # The macros are very small so use a smaller halo
 export MACRO_PLACE_HALO        ?= 5 5
+
+export CACHED_FLATTEN_POLICY = ./designs/$(PLATFORM)/$(DESIGN_NICKNAME)/flatten.tcl
+export ADDITIONAL_FILES = $(CACHED_FLATTEN_POLICY)
```
maliberty: Why is this necessary for this one design (along with flatten.tcl)?

flow/Makefile
```diff
@@ -541,7 +537,7 @@ $(RESULTS_DIR)/1_synth.rtlil:
 	$(UNSET_AND_MAKE) do-yosys-canonicalize
 
 $(RESULTS_DIR)/1_1_yosys.v: $(RESULTS_DIR)/1_synth.rtlil
-	$(UNSET_AND_MAKE) do-yosys
+	$(UNSET_AND_MAKE) do-yosys-flatten-policy do-yosys
```
maliberty: It looks like SYNTH_HIERARCHICAL is no longer supported.  Is that correct and if so, why?

oharboe: Still supported, but the iffy code from Makefile is gone. It is all handled in .tcl now.

flow/designs/asap7/aes-block/config.mk
oharboe: aes-block is more of an ORFS script test case for blocks and other and not really a quality of results/real design, so it seems like a good choice for testing this feature.

flow/scripts/synth_hier_report.tcl
```diff
@@ -69,3 +75,4 @@ if { [info exist ::env(REPORTS_DIR)] && [file isfile $::env(REPORTS_DIR)/synth_h
   close $out_script_ptr
 }
 
+write_flatten_policy
```
maliberty: The naming feels a bit backwards throught.  The file contains the list of things _not_ to flatten.

oharboe: That to me is an implementation detail.

In a compiler we call it inlining policy regardless of how it is implemented. unflattening policy implies to me that the modules dont  exist, but have to be created... but they exist and we have a policy of which to flatten.

maliberty: You have promoted this to a result and not an implementation detail.  It was mark_hier_stop_modules.tcl; perhaps just keep_hierarchy.tcl

oharboe: fixed.

flow/scripts/synth_hier_report.tcl
```diff
@@ -1,22 +1,28 @@
-source $::env(SCRIPTS_DIR)/synth_preamble.tcl
+proc write_flatten_policy {} {
```
maliberty: Please update to match

flow/Makefile
```diff
@@ -508,15 +507,12 @@ memory:
 export SYNTH_SCRIPT ?= $(SCRIPTS_DIR)/synth.tcl
 export SYNTH_MEMORY_MAX_BITS ?= 4096
 
-$(SYNTH_STOP_MODULE_SCRIPT):
+.PHONY: do-yosys-flatten-policy
+do-yosys-flatten-policy:
```
maliberty: update to match

maliberty (changes requested):

flow/scripts/synth_hier_report.tcl
oharboe: fixed

flow/Makefile

maliberty (approved):

Issue #1429, opened 2023-09-01
Create 0_prep.log to view logs related to design flow preparation


vijayank88:
### Description

ORFS flow starts recording logs from synthesis stage with `1_1_yosys.log`. Snippet of logs
```
1. Executing Verilog-2005 frontend: ./designs/src/gcd/gcd.v
2. Executing Liberty frontend.
3. Executing Liberty frontend.
4. Executing Liberty frontend.
```
But previously there is design flow preparation with `dont use cells` and creating `merged.lib`, it was missing part of logs.

It is useful when trying with different corners of ASAP7 and to make sure updated `DONT_USE_CELLS` processed properly or not.

### Suggested Solution

As we start synthesis with `1_1_yosys.log` either append platform preparation log in same file or create another new file `0_prep.log`. 
The following format should be captured in log.
```
[INFO][FLOW] Using platform directory ./platforms/asap7
[INFO-FLOW] ASU ASAP7 - version 2
Default PVT selection: BC
./util/markDontUse.py -p "*x1p*_ASAP7* *xp*_ASAP7* SDF* ICG* DFFH*" -i platforms/asap7/lib/asap7sc7p5t_AO_RVT_FF_nldm_211120.lib.gz -o objects/asap7/gcd/base/lib/asap7sc7p5t_AO_RVT_FF_nldm_211120.lib
Opening file for replace: platforms/asap7/lib/asap7sc7p5t_AO_RVT_FF_nldm_211120.lib.gz
Marked 19 cells as dont_use
Commented 0 lines containing "original_pin"
Replaced malformed functions 0
Writing replaced file: objects/asap7/gcd/base/lib/asap7sc7p5t_AO_RVT_FF_nldm_211120.lib
./util/markDontUse.py -p "*x1p*_ASAP7* *xp*_ASAP7* SDF* ICG* DFFH*" -i platforms/asap7/lib/asap7sc7p5t_INVBUF_RVT_FF_nldm_220122.lib.gz -o objects/asap7/gcd/base/lib/asap7sc7p5t_INVBUF_RVT_FF_nldm_220122.lib
Opening file for replace: platforms/asap7/lib/asap7sc7p5t_INVBUF_RVT_FF_nldm_220122.lib.gz
Marked 6 cells as dont_use
Commented 0 lines containing "original_pin"
Replaced malformed functions 0
[...]

PR #2315, opened 2024-09-02
Printing a list of DONT_USE CELLS.

socclosure:
Fix for the issue
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1429

### DIFF
--- a/flow/scripts/synth_preamble.tcl
+++ b/flow/scripts/synth_preamble.tcl
@@ -79,7 +79,8 @@ set abc_args [list -script $abc_script \
 # LIB_FILES and ADDITIONAL_LIBS that are included in LIB_FILES.
 if {[info exist ::env(DONT_USE_CELLS)] && $::env(DONT_USE_CELLS) != ""} {
   foreach cell $::env(DONT_USE_CELLS) {
-    lappend abc_args -dont_use $cell
+      puts "INFO: Lib Cell $cell is in the list of DONT_USE_CELLS so it will not be used."
+      lappend abc_args -dont_use $cell
   }
 }
 


rovinski:
This seems to just be printing the environment variable? Is that useful?

socclosure:
It is not unix environment which can be viewed using echo while it can be accessed in the tool. In issue # 1429 user has requested to print DONT_USE cell list in the log file which is being implemented int this PR.

rovinski:
I have a few issues with this:

1. This can already be done with the `make print-%` rule pattern:
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/bb190447797192a1bc2c43db81d5cb568513dd59/flow/Makefile#L1125-L1127
So running `make print-DONT_USE_CELLS` will provide this result.
2. This doesn't actually print cells that aren't used, it only prints each item in the variable. The item in the variable may not match any cells in the lib (for example, `INV_x1` won't match `INV_X1`, the former may not actually exist in the library).
3. It will also not resolve any wildcards, so it would print `INV_*` as the dont_use cell rather than all of `INV_X1`, `INV_X2`, `INV_X4`, ...
4. As stated in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1429#issuecomment-2296425056 this information technically already shows in the log

I think this would be more useful if it actually used a `get_cells` call to resolve the cell names, and also formatted it in a list

socclosure:
@rovinski @maliberty @vijayank88 

I need your help for the requirement or need of the ISSUE_1429.
Please find my inline response of last comment.

> I have a few issues with this:
> 
> 1. This can already be done with the `make print-%` rule pattern:
>    https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/bb190447797192a1bc2c43db81d5cb568513dd59/flow/Makefile#L1125-L1127
>    
>    So running `make print-DONT_USE_CELLS` will provide this result.

**### Jignesh: When I execute "make print-DONT_USE_CELLS", it does not display any list. It's possible that I may need to set or change a variable. If the user agrees with this approach, then we can close this issue without checking in any code changes and we can discard this pull request.**

> 2. This doesn't actually print cells that aren't used, it only prints each item in the variable. The item in the variable may not match any cells in the lib (for example, `INV_x1` won't match `INV_X1`, the former may not actually exist in the library).

** Jignesh: Correct. The request is to print setting of "DONT_USE_CELLS". If any value of "DONT_USE_CELLS" is incorrect then it should be informed by Synthesis tool when it will apply that list. **

> 3. It will also not resolve any wildcards, so it would print `INV_*` as the dont_use cell rather than all of `INV_X1`, `INV_X2`, `INV_X4`, ...

** Jignesh: Same feedback as point 2 **

> 4. As stated in [Create 0_prep.log to view logs related to design flow preparation #1429 (comment)](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1429#issuecomment-2296425056) this information technically already shows in the log
> 
> I think this would be more useful if it actually used a `get_cells` call to resolve the cell names, and also formatted it in a list

** Jignesh:  If user can clarify more on requirement, then it will be better my interpretation from the comment is that user can see merge.lib in log file and a list of "DONT_USE_CELL" not actual cells still need to be print in Yosys log file. **

Please guide on what need to be done for ISSUE_1496. If code change is not required, then we can discard the PR and close the ISSUE.

rovinski:
According to https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1429#issuecomment-2296425056 the current state is good enough, so I would recommend closing the issue and this PR.

socclosure:
Thanks @rovinski for recommending.
I will discard the PR.
@vijayank88 / @maliberty 
Please close the ISSUE 1429 or give your input.

socclosure:
I am closing this PR.

PR #2327, opened 2024-09-06
makefile: make STDBUF_CMD optional

oharboe:
originally added for Mac in 4fa7b95bb, solves problems that don't exist on x86 Linux distributions normally.

Removing stdbuf dependency is helpful when running in some containerized environments, like Bazel.


oharboe:
@jeffng-or @hovind FYI

maliberty (approved):

PR #2332, opened 2024-09-09
makefile: make memory mem_dump.py regression

oharboe:
accidentially checked total memory size for all
instances of a type of memory, instead of the maximum size of a memory instance, like before.


oharboe:
@jeffng-or FYI

maliberty (approved):

PR #2333, opened 2024-09-09
scripts: less repetition

oharboe:


flow/scripts/floorplan.tcl
```diff
@@ -104,13 +104,7 @@ if {[info exists ::env(FOOTPRINT_TCL)]} {
...
   remove_buffers
 } else {
-  set additional_args "-verbose"
-  append_env_var additional_args SETUP_SLACK_MARGIN -setup_margin 1
-  append_env_var additional_args TNS_END_PERCENT -repair_tns 1
-  append_env_var additional_args SKIP_PIN_SWAP -skip_pin_swap 0
-  append_env_var additional_args SKIP_GATE_CLONING -skip_gate_cloning 0
-  append_env_var additional_args SKIP_BUFFER_REMOVAL -skip_buffer_removal 0
-  repair_timing {*}$additional_args
+  repair_timing_helper 0
```
oharboe, line 107: @precisionmoon was the omission in ddce1aa48b of HOLD_SLACK_MARGIN for floorplan intentional?

maliberty: Is it even worth doing hold fixing this early?

maliberty:
The code is fine but I'll wait for the discussion to finish

flow/scripts/floorplan.tcl
precisionmoon: Yes, this was intentional.  We are not doing any hold fixing at FP step.   Buffer removal is only part of setup fixing.

maliberty: @precisionmoon if you give neither -setup nor -hold then both are repaired.  I think this is doing hold fixing currently.

oharboe: Makes sense. The updated code makes the behavioral difference in floorplan more clear(otherwise you have to spot the difference).

maliberty: I think we should merge this as a non-functional change and address hold fixing seperately.

maliberty (approved):

PR #2335, opened 2024-09-10
Set constant build path and timestamp for executables in Docker image

glatosinski:
This PR modifies the build flow for Docker images to:

* Set build path for dependencies to a constant path (so that `__FILE__` is not different on every build of the image)
* Set TIME/DATE/TIMESTAMP to constant values so that the binaries are not affected by a different build time

Those changes prevent cache invalidation in Bazel due to different binary SHAs.

This PR depends on https://github.com/The-OpenROAD-Project/OpenROAD/pull/5726, since it introduces `-constant-build-dir`

### DIFF
--- a/docker/Dockerfile.builder
+++ b/docker/Dockerfile.builder
@@ -19,6 +19,8 @@ COPY --link tools tools
 ARG numThreads=$(nproc)
 
 RUN echo "" > tools/yosys/abc/.gitcommit && \
+  env CFLAGS="-D__TIME__=0 -D__DATE__=0 -D__TIMESTAMP__=0 -Wno-builtin-macro-redefined" \
+  CXXFLAGS="-D__TIME__=0 -D__DATE__=0 -D__TIMESTAMP__=0 -Wno-builtin-macro-redefined" \
   ./build_openroad.sh --no_init --local --threads ${numThreads}
 
 FROM orfs-base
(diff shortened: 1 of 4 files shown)


maliberty (approved):

maliberty:
Can't merge until "This branch has conflicts that must be resolved" is fixed.

oharboe:
@maliberty it is only a tools/OpenROAD conflict. pr-head is passing, so you can probably just merge it locally ignoring the tools/OpenROAD update and close the pr.

maliberty:
Replaced by #2353

eder-matheus:
@glatosinski Recently I had to revert this TIME/DATE/TIMESTAMP constant values due to build errors in the latest Yosys version. Do you know if these macros could be redefined only during compilation time? I.e., instead of redefining them in the environment in the dockerfiles, we would just define them in the build_openroad.sh in the docker build.

oharboe:
> @glatosinski Recently I had to revert this TIME/DATE/TIMESTAMP constant values due to build errors in the latest Yosys version. Do you know if these macros could be redefined only during compilation time? I.e., instead of redefining them in the environment in the dockerfiles, we would just define them in the build_openroad.sh in the docker build.

@eder-matheus Will you fix this?

maliberty:
What motived this change to begin with?

@oharboe we already reverted the change

oharboe:
> What motived this change to begin with?
> 
> @oharboe we already reverted the change

Bazel-orfs depends on binaries in the docker image and dates in binaries break content hashes.

oharboe:
@QuantamHD FYI

PR #2337, opened 2024-09-11
Makefile POLA and less repetition tweaks

oharboe:
Fewer surprises and less reptition.

Printing the default value now works:

```
$ make DESIGN_CONFIG=designs/asap7/mock-array/config.mk print-EQUIVALENCE_CHECK
EQUIVALENCE_CHECK = 0
$ make DESIGN_CONFIG=designs/asap7/aes/config.mk print-EQUIVALENCE_CHECK
EQUIVALENCE_CHECK = 1
```

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -144,6 +144,7 @@ export ROUTING_LAYER_ADJUSTMENT ?= 0.5
 export RECOVER_POWER ?= 0
 export SKIP_INCREMENTAL_REPAIR ?= 0
 export DETAILED_METRICS ?= 0
+export EQUIVALENCE_CHECK ?= 0
 
 # If we are running headless use offscreen rendering for save_image
 ifndef DISPLAY
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -2,6 +2,11 @@
 # file to avoid having to adding the to the make command line.
 -include settings.mk
 
+# Include design and platform configuration before setting default options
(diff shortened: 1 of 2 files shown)


oharboe:
@jeffng-or FYI

flow/Makefile
```diff
@@ -2,6 +2,11 @@
 # file to avoid having to adding the to the make command line.
 -include settings.mk
 
+# Include design and platform configuration before setting default options
+# in this file. This allows the DESIGN_CONFIG to set different defaults than
+# this file.
+include $(DESIGN_CONFIG)
```
maliberty, line 8: Wouldn't the design be overriding the other values rather than setting new defaults?  I don't get the idea here.

oharboe: The design can provide a different default value than Makefile+platform, such as:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/8184f7116cf43c3d1f9964e7a2e900946bb7e1a8/flow/designs/asap7/aes-block/config.mk#L32

maliberty: Why would a design provide a default?  Wouldn't you just set the value you want with = ?

oharboe: these are test designs so it is sometimes convenient to quickly try alternatives on the command line.

maliberty: 💡 got it

maliberty (approved):

PR #2338, opened 2024-09-11
Makefile: remove debug noise

oharboe:
Cleans up/less surprising output. No difference in `make print-FOO` behavior with and without BLOCKS.

```
$ make DESIGN_CONFIG=designs/asap7/mock-array/config.mk print-EQUIVALENCE_CHECK
[INFO][FLOW] Invoked hierarchical flow.
Block Element needs to be hardened.
EQUIVALENCE_CHECK =
```

To:

```
$ make DESIGN_CONFIG=designs/asap7/mock-array/config.mk print-EQUIVALENCE_CHECK
EQUIVALENCE_CHECK = 
```


maliberty:
This change would affect the normal flow too?  It seems useful in that case.

oharboe:
> This change would affect the normal flow too? 

Yes.

> It seems useful in that case.

I think it is just noise. There are a 100 other things that are always true that could be printed...

maliberty (approved):

PR #2339, opened 2024-09-11
[DO NOT REVIEW] aes-blocks fix constraints

oharboe:


oharboe:
@maliberty fixes unknown "clock" warning, but fails in PDN. Will look at it later, backing up for now...

oharboe:
out of sight, out of mind for now

PR #2341, opened 2024-09-11
Setup 24 04 fix

oharboe:


maliberty (commented):
What is the issue in 24.04?  As you've been quoting principles, this one goes against [don't repeat yourself](https://en.wikipedia.org/wiki/Don't_repeat_yourself)

oharboe:
> What is the issue in 24.04? As you've been quoting principles, this one goes against [don't repeat yourself](https://en.wikipedia.org/wiki/Don't_repeat_yourself)

Syntax error not triggered in 23.04 because of lazy evaluation.

Yes: I would have liked to, and I tried, to avoid the repetition.

Someone more versed than me and ChatGPT in bash will be able to fix this easily....

maliberty:
Does
```
    if  [[ $1 == "rodete" || $(_versionCompare "$1" -ge 23.04) ]]; then
```
work?

oharboe:
> Does
> ```
>     if  [[ $1 == "rodete" || $(_versionCompare "$1" -ge 23.04) ]]; then
> ```
> work?

No. Tried ten variants...

maliberty (approved):
Ok I don't want to fight bash either.

PR #2344, opened 2024-09-12
floorplan: replace surprises and documentation with actionable error messages

oharboe:
I've been suckerpunched several times by CORE_UTILIZATION overriding DIE/CORE_AREA.

Remove documentation/rules and replace it with checks and actionable error messages.

A floorplan config suckerpunch can manifest itself as late as in macro placement with an [inactionable error message](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5735)....

### DIFF
--- a/docs/user/FlowVariables.md
+++ b/docs/user/FlowVariables.md
@@ -274,6 +274,8 @@ configuration file.
 | `MACRO_HALO_X`         | Set macro halo for x-direction. Only available for ASAP7 PDK. |
 | `MACRO_HALO_Y`         | Set macro halo for y-direction. Only available for ASAP7 PDK. |
 
+The various methods to specify the die and core area(`FLOORPLAN_DEF`, `FOOTPRINT`, `DIE_AREA`, `CORE_AREA` and `CORE_UTILIZATION`) are mutually exclusive. If two methods are specified, floorplan.tcl will exit with an error requiring that a single method is used.
+
 #### Placement
 
 
--- a/docs/user/FlowVariables.md
+++ b/docs/user/FlowVariables.md
@@ -275,7 +275,7 @@ configuration file.
 | `MACRO_HALO_X`         | Set macro halo for x-direction. Only available for ASAP7 PDK. |
 | `MACRO_HALO_Y`         | Set macro halo for y-direction. Only available for ASAP7 PDK. |
 
-The various methods to specify the die and core area(`FLOORPLAN_DEF`, `FOOTPRINT`, `DIE_AREA`, `CORE_AREA` and `CORE_UTILIZATION`) are mutually exclusive. If two methods are specified, floorplan.tcl will exit with an error requiring that a single method is used.
+The various methods to specify the die and core area(`FLOORPLAN_DEF`, `FOOTPRINT`, `DIE_AREA` and `CORE_UTILIZATION`) are mutually exclusive. If two methods are specified, floorplan.tcl will exit with an error requiring that a single method is used.
 
 #### Placement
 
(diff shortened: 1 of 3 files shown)


oharboe:
@jeffng-or FYI

maliberty:
I don't see why we want to un-document this.  It is good to have errors but it shouldn't be necessary to run the flow to discover how it works.

oharboe:
> I don't see why we want to un-document this. It is good to have errors but it shouldn't be necessary to run the flow to discover how it works.

Because it is no longer true: if you have two ways to configure the floorplan enabled, you get an error message that you should pick one instead of one silently winning...

oharboe:
> > I don't see why we want to un-document this. It is good to have errors but it shouldn't be necessary to run the flow to discover how it works.
> 
> Because it is no longer true: if you have two ways to configure the floorplan enabled, you get an error message that you should pick one instead of one silently winning...

I looked at almost all the CORE_UTILIZATION examples in flow, I didn't find a single one that also specified DIE/CORE_AREA.

Hence, I don't think that having an ambigious floorplan definition is a common intentional use-case.

I seriously doubt that anyone would read the documentation to try to find out if one should pick a single way to specify the floorplan dimensions... To me this is just documentation drivel that reduces the documentation quality.

Also, the existing documentation was incomplete, e.g. it did not document FLOORPLAN_DEF vs. CORE_UTILIZATION priorities.

Hence... replace incomplete and obtuse documentation and surprising behaviors with actionable error messages.

oharboe:
@maliberty unrelated errors. Not a single design under test is specifying floorplan ambigously.

maliberty:
> > I don't see why we want to un-document this. It is good to have errors but it shouldn't be necessary to run the flow to discover how it works.
> 
> Because it is no longer true: if you have two ways to configure the floorplan enabled, you get an error message that you should pick one instead of one silently winning...

Then lets say "cannot be used with" or "mutually exclusive with" to capture it.

maliberty:
Please merge master to resolve the failures.

oharboe:
@maliberty @jeffng-or [More evidence](https://github.com/The-OpenROAD-Project/bazel-orfs/pull/146/commits/4ebd2e574d885cfd1d22b94b5c7d0751a928531b#r1759002285) today that errors beat documentation; that is, an error is encountered sooner than documentation is read.

docs/user/FlowVariables.md
```diff
@@ -265,16 +265,18 @@ configuration file.
...
+| `CORE_ASPECT_RATIO`   | The core aspect ratio (height / width). This values is ignored if `CORE_UTILIZATION` undefined. |
+| `CORE_MARGIN`         | The margin between the core area and die area, in multiples of SITE heights. The margin is applied to each side. This variable is ignored if `CORE_UTILIZATION` is undefined. |
+| `DIE_AREA`            | The die area specified as a list of lower-left and upper-right corners in microns (X1 Y1 X2 Y2). |
+| `CORE_AREA`           | The core area specified as a list of lower-left and upper-right corners in microns (X1 Y1 X2 Y2). |
 | `RESYNTH_AREA_RECOVER` | Enable re-synthesis for area reclaim. |
 | `RESYNTH_TIMING_RECOVER` | Enable re-synthesis for timing optimization. | 
 | `MACRO_HALO_X`         | Set macro halo for x-direction. Only available for ASAP7 PDK. |
 | `MACRO_HALO_Y`         | Set macro halo for y-direction. Only available for ASAP7 PDK. |
 
+The various methods to specify the die and core area(`FLOORPLAN_DEF`, `FOOTPRINT`, `DIE_AREA`, `CORE_AREA` and `CORE_UTILIZATION`) are mutually exclusive. If two methods are specified, floorplan.tcl will exit with an error requiring that a single method is used.
```
maliberty: There are four methods
* FLOORPLAN_DEF
* DIE_AREA and CORE_AREA
* FOOTPRINT  (obsolete and should be undocumented)
* CORE_UTILIZATION with optional CORE_ASPECT_RATIO

Its not right to say DIE_AREA and CORE_AREA are mutually exclusive.

oharboe: Well spotted. Fixed.

maliberty (approved):

PR #2345, opened 2024-09-12
Makefile: fix my gaffe where 'make' taking DESIGN_CONFIG from Makefile broke

oharboe:
Mea culpa.

Locally it is useful to edit the Makefile to set a default DESIGN_CONFIG to use, whereas on servers and for non-ORFS designs DESIGN_CONFIG is always specified.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -2,11 +2,6 @@
 # file to avoid having to adding the to the make command line.
 -include settings.mk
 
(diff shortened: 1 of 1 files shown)


maliberty (approved):

oharboe:
@vvbandeira @maliberty Unrelated failure.

PR #2347, opened 2024-09-13
setup: avoid confusion when user is not up to date

oharboe:
A colleague of mine got confused by out of date submodules and root ownership yesterday after having updated to latest main and trying to run `sudo ./setup.sh` and I got confused today because I logged into a laptop I haven't used in a while and didn't think to submodules before running `sudo ./setup.sh`.

It is a pattern...

These changes should make it easier for users not to get caught in these snags...

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -10,8 +10,20 @@ if [ $EUID -ne 0 ]; then
   exit 1
 fi
 
-if git submodule status --recursive | grep -q "^-"; then
-  git submodule update --init --recursive
+tmpfile=$(mktemp)
+# any error messages from this command will stand out
+# more clearly when run as a separate command rather than
+# being piped
+git submodule status --recursive > "$tmpfile"
+
+if grep -q "^-" "$tmpfile"; then
+  sudo -u $SUDO_USER git submodule update --init --recursive
+elif grep -q "^+" "$tmpfile"; then
+  # Make it easy for users who are not hacking ORFS to do the right thing,
+  # run with current submodules, at the cost of having ORFS
+  # hackers disable this test manually when hacking setup.sh
+  echo "Submodules are not up to date, run 'git submodule update --recursive' and try again"
+  exit 1
 fi
 
 "$DIR/etc/DependencyInstaller.sh" -base


setup.sh
```diff
@@ -10,8 +10,20 @@ if [ $EUID -ne 0 ]; then
...
-if git submodule status --recursive | grep -q "^-"; then
-  git submodule update --init --recursive
+tmpfile=$(mktemp)
+# any error messages from this command will stand out
+# more clearly when run as a separate command rather than
+# being piped
+git submodule status --recursive > "$tmpfile"
+
+if grep -q "^-" "$tmpfile"; then
+  sudo -u $SUDO_USER git submodule update --init --recursive
```
oharboe, line 20: @jeffng-or @maliberty Fascinating that `sudo -u $SUDO_USER` prefix was missing up until now...

This causes all sorts of mystical errors that users must have silently suffered through fixed by fixing permissions....

It is not normally visible for experts who have had ORFS installed since forever, but occasional/first time hackers of ORFS must have gotten snagged on this...

maliberty: Why does updating ORFS need sudo?

oharboe: It is the oposite, it needs NOT to use sudo.

Recall that  `sudo ./setup.sh` is run, `sudo -u $SUDO_USER ...` drops out of sudo and runs it as the user.

maliberty (approved):

setup.sh
```diff
@@ -10,8 +10,20 @@ if [ $EUID -ne 0 ]; then
...
+elif grep -q "^+" "$tmpfile"; then
+  # Make it easy for users who are not hacking ORFS to do the right thing,
+  # run with current submodules, at the cost of having ORFS
+  # hackers disable this test manually when hacking setup.sh
+  echo "Submodules are not up to date, run 'git submodule update --recursive' and try again"
+  exit 1
```
maliberty, line 26: I'm not enthusiastic about this being a hard error but we'll see how it goes.

oharboe: I would rather err on the side of doing nothing with an actionable message, than damage...

oharboe: But yes... let's see how it goes...

oharboe: Though: this PR is evidence of silent suffering... I think we'd have to peer over colleagues shoulders to pick up the problems they are seeing when occasionally hacking ORFS....

rovinski:
@oharboe you should remove the tmpfile after it's done being useful. I think it should be possible to avoid a tmpfile at all if you capture the output in a variable (you can print the variable too if you like)

oharboe:
I tried and failed to avoid the tmpfile... It seems like it should be simple but all I got to show for it was a bøack eye...

rovinski:
Does this not work for you? It worked for me.
```
output=$(git submodule status --recursive)
if echo "$output" | grep -q "^-"; then
  echo "Submodules are not up to date, run 'git submodule update --recursive' and try again"
  exit 1
fi
```

oharboe:
> Does this not work for you? It worked for me.
> 
> ```
> output=$(git submodule status --recursive)
> if echo "$test" | grep -q "^-"; then
>   echo "Submodules are not up to date, run 'git submodule update --recursive' and try again"
>   exit 1
> fi
> ```

I messed around with this and what I could not figure out was why on Ubuntu 24.04 it didn't work for `grep -q "^+"`, whereas it worked for `grep -q "^-"`.

Disgusted and frustrated with bash, I decided to focus on the price and fix the problem at hand... :-)

oharboe:
Something funny is going on. I'm loath to make the code more fancy until I can understand what is happening here...

```
$ git submodule status --recursive | grep -q "^+"
fatal: failed to recurse into submodule 'tools/OpenROAD'
$ git submodule status --recursive | grep -q "^-"
$ git submodule status --recursive
+a487b9b68b50381b212beb920a79e645b4f75cc0 tools/OpenROAD (v2.0-15616-ga487b9b68)
 aafee90f8a21bf7867cef2e159929440cf45b2e5 tools/OpenROAD/src/sta (v2.2.0-1106-gaafee90)
 ef5389d31526003c2ebd7e6d6d6fe3848a20f0a2 tools/OpenROAD/third-party/abc (remotes/origin/HEAD)
 80ba43d26264738c93900129dc0aab7fab36c53f tools/yosys (yosys-0.44)
 28d955ca97a1c4be3aed4062aec0241a734fac5d tools/yosys/abc (yosys-0.44)
```

rovinski:
Well if you're looking to fix the `fatal` it's probably printing to `stderr` which by default doesn't get captured when piping. If you do want to capture it, you can use a pipe redirect.

As for why it's failing, I'm not sure.

oharboe:
when I redirect to a file it doesn't happen...

rovinski:
Redirection and piping work the same way. If you wanted to capture stdout+stderr in the file you would need `&>` instead:
```
git submodule status --recursive &> "$tmpfile"
```
So I think that the fatal is unrelated to pipe vs. file.

oharboe:
Leaving this sleeping dog lie, because there is something here I dont understand and it works now.

maliberty:
It looks like 'grep -q' exits when it sees the first match and git doesn't like that (feels like a git bug to me).  I don't see a better solution without a tmp file.  I sent a note to the git mailing list to see what happens.

oharboe:
Nice! `sponge` can be used to "fix" this...

```
$ git submodule status --recursive | sponge | grep -q "^+"
$ echo $?
0
$ git submodule status --recursive | sponge | grep -q "^-"
$ echo $?
1
```

maliberty:
Cool but it would be another dependency to add.

oharboe:
let the dog lie?

maliberty:
ok with me

rovinski:
What is wrong with redirecting to a variable first then?

```
output=$(git submodule status --recursive)
echo "$output" | grep -q "^-"
```

oharboe:
Try it out, but I am concerned about variable length limit and error propagation here

rovinski:
StackOverflow indicates the variable limit is [at least several megabytes](https://stackoverflow.com/questions/1078031/what-is-the-maximum-size-of-a-linux-environment-variable-value/1078125#1078125) and errors can be handled but I'm not going to push this issue further.

oharboe:
> StackOverflow indicates the variable limit is [at least several megabytes](https://stackoverflow.com/questions/1078031/what-is-the-maximum-size-of-a-linux-environment-variable-value/1078125#1078125) and errors can be handled but I'm not going to push this issue further.

It looks like a promising approach, but I'm loath to change anything without bottoming out on error handling and testing it fully... Contributions welcome...

PR #2348, opened 2024-09-13
asap7/constraints.sdc: make set_max_delay configurable

oharboe:
this oftentimes needs to be adjusted, so making it configurable avoids copy and paste in smaller test-rigs and makes the pertinent changes in those test-rig easier to read in the constraints.sdc files using the asap7/constraints.sdc file.


maliberty (approved):

maliberty:
To inject a value do you make a wrapper sdc?

PR #2352, opened 2024-09-15
generate_abstract: ABSTRACT_SOURCE=5_1_grt now works

oharboe:
gui.tcl smarts to find .sdc file belonging to .odb now centralized and used by generate_abstract.

### DIFF
--- a/flow/scripts/util.tcl
+++ b/flow/scripts/util.tcl
@@ -44,3 +44,18 @@ proc recover_power {} {
   report_wns
   report_power
 }
+
+proc find_sdc_file {input_file} {
+    # Determine design stage (1 ... 6)
+    set design_stage [lindex [split [file tail $input_file] "_"] 0]
+
+    # Read SDC, first try to find the most recent SDC file for the stage
+    set sdc_file ""
+    for {set s $design_stage} {$s > 0} {incr s -1} {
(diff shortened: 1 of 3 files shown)


maliberty (approved):

flow/scripts/util.tcl
```diff
@@ -44,3 +44,18 @@ proc recover_power {} {
...
 }
+
+proc find_sdc_file {input_file} {
+    # Determine design stage (1 ... 6)
+    set design_stage [lindex [split [file tail $input_file] "_"] 0]
+
+    # Read SDC, first try to find the most recent SDC file for the stage
+    set sdc_file ""
+    for {set s $design_stage} {$s > 0} {incr s -1} {
+        set sdc_file [glob -nocomplain -directory $::env(RESULTS_DIR) -types f "${s}_\[A-Za-z\]*\.sdc"]
```
oharboe, line 55: @maliberty This will find 5_route.sdc for 5_1_grt.odb, which is wrong... But that problem was there *before* this PR, so a separate concern to this PR. I haven't seen any problems because of this, but thought I should mention it.

maliberty: There is no change in the SDC at this point in the flow but noted.

oharboe: And here is an example of bad things happening because of this bug...

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2354

PR #2368, opened 2024-09-23
tcl: use utility functions instead of inconsistent idioms

oharboe:
Less code and fewer surprises, make FOO= should generally turn off a feature, not run the feature with an empty argument.

### DIFF
--- a/flow/scripts/floorplan.tcl
+++ b/flow/scripts/floorplan.tcl
@@ -40,7 +40,7 @@ set num_instances [llength [get_cells -hier *]]
 puts "number instances in verilog is $num_instances"
 
 set additional_args ""
-append_env_var additinal_args ADDITIONAL_SITES -additional_sites 1
+append_env_var additional_args ADDITIONAL_SITES -additional_sites 1
 
 set use_floorplan_def [env_var_exists_and_non_empty FLOORPLAN_DEF]
 set use_footprint [env_var_exists_and_non_empty FOOTPRINT]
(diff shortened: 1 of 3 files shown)


flow/scripts/util.tcl
```diff
@@ -64,3 +64,21 @@ proc find_sdc_file {input_file} {
     }
     return [list $design_stage $sdc_file]
 }
+
+proc env_var_equals {env_var value} {
+    return [info exists ::env($env_var)] && $::env($env_var) == $value
```
maliberty: Does this need an expr around it?

flow/scripts/floorplan.tcl
```diff
@@ -40,13 +40,7 @@ set num_instances [llength [get_cells -hier *]]
...
 
 set additional_args ""
-if { [info exists ::env(ADDITIONAL_SITES)]} {
-  append additional_args " -additional_sites $::env(ADDITIONAL_SITES)"
-}
-
-proc env_var_exists_and_non_empty {env_var} {
-    return [expr {[info exists ::env($env_var)] && ![string equal $::env($env_var) ""]}]
-}
+append_env_var additinal_args ADDITIONAL_SITES -additional_sites 1
```
maliberty:
```suggestion
append_env_var additional_args ADDITIONAL_SITES -additional_sites 1
```

maliberty: `1` is wrong as it should be appending $::env(ADDITIONAL_SITES)

maliberty (approved):

PR #2375, opened 2024-09-24
synthesis: improve error messages

oharboe:
YOSYS_FLAGS=-v 3 will silence error messages, such as the below.

However, without this option, synthesis will give an enormous amount of output.

Fortunately, the most common error messages, like unknown top module, is shown during canonicalization and not specifying -v 3 doesn't increase the logging noise noticably during the canonicalization stage.

To test:

1. modify gcd.v to rename "gcd" module to "gcdxxx"
2. Run "make synth"

Now it fails with a helpful error message

[deleted]
4. Executing HIERARCHY pass (managing design hierarchy). ERROR: Module `gcd' not found!
ERROR: TCL interpreter returned an error: Yosys command produced an error Command exited with non-zero status 1

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -540,7 +540,8 @@ do-yosys:
 .PHONY: do-yosys-canonicalize
 do-yosys-canonicalize: yosys-dependencies
 	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR) $(OBJECTS_DIR)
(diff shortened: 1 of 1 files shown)


maliberty:
@povik Is this the best solution?  No value for -v seems to recover this message.

povik:
There are two kinds of errors in Yosys: recoverable, which will throw you back to the prompt in an interactive session, and non-recoverable, which will exit the process. The first appear to be silenced under `-v 3`, which I don't think we want. I will write a patch.

maliberty:
@oharboe can you wait for the next yosys update cycle or do you need this as a stop-gap?

oharboe:
It is easily reverted, isnt it? This way I can forget all about it :-)

povik:
The patch is up

maliberty (approved):

PR #2379, opened 2024-09-24
Enable optionally naja_edit between Yosys and OR

xtofalex:
Two env variables are used:

- DO_NAJA: if = 1, naja_edit will be launched after yosys
- NAJA_INSTALL: to point to naja installation 
naja_edit will perform netlist cleaning and optimizations: Dead Logic Elimination and constant propagation.


oharboe (changes requested):
Since all this does is to load 1_synth.v file and overwrite it, I dont think it is necessary or beneficial to have this in the ORFS repository until it is mature enough that a large percentage of ORFS users would want to run with it.

This PR uses a lot of dependencies which would have to be added to the ORFS setup install scripts and docker images to be useful/effective in a end user or CI context.

Near as I can see, the command that needs to be run by the users for such optimization tools is:

make synth
optimize -mode flat/hierarchical results/.../1_synth.v
make final

Though I think some adoptations could be made to ORFS to make injection of such a stage easier.

Possibly the 1_synth.v path to be read in by the next stage should be a config variable to allow injection of such an optimization stage...

Thoughts?

oharboe:
@xtofalex will try https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2380 to see if that covers what is needed and I'll offer som assistance in that effort.

maliberty:
I think we should either adopt naja or leave it out.  Half adopting it into the flow seems odd.  Anyone trying this option will almost certainly fail to have naja_edit installed, as its not part of the dependency installer, and will be confused.

I suggest having a more complete PR where anyone can try this out that's interested (imagine it were on by default).  Also I would post more information about Naja's benefits to get community buy-in.

oharboe:
The other PR that I have adds support for out of ORFS tree naja test cases(or any other similar synthesis enhancement).

Later, if ORFA decides to have naja in depenency installers, that is a straightforward tweak on top by the user.

maliberty:
I looked at that PR but it feels like a complication that nobody will actually use.  You would have to install it yourself at which point you might as well use this PR.

oharboe:
> I looked at that PR but it feels like a complication that nobody will actually use. You would have to install it yourself at which point you might as well use this PR.

I think updating the dependencyinstaller and this PR are different concerns.

There are several use-cases:

- naja needs some ORFS test designs in CI in their own git repository, my PR enables that. naja CI needs to be able to update their repository during dev without creating ORFS pull requests(or it gets awkward).
- end-user uses naja. This really requires that the dependencyinstaller in ORFS is updated + either my PR above + a tweak or something like this PR.

I looked at the numbers and until naja makes a bigger splash (currently low single digit percent on most designs), I don't think there will be a rush of ORFS users wanting to use it. naja does not yet support register duplication and cloning, which is the big missing feature for megaboom(according to BOOM documentation).

So there's time and I think supporting naja CI with minimal and changes to ORFS is the best next step.

xtofalex:
I should be able to get PnR results with/without naja_edit by next week. Also I will run eqy compare on designs where it is feasible.I suggest to reorient the discussion based on those results.Thanks Christophe Le 1 oct. 2024 à 06:25, Øyvind Harboe ***@***.***> a écrit :﻿

I looked at that PR but it feels like a complication that nobody will actually use. You would have to install it yourself at which point you might as well use this PR.

I think updating the dependencyinstaller and this PR are different concerns.
There are several use-cases:

naja needs some ORFS test designs in CI in their own git repository, my PR enables that. naja CI needs to be able to update their repository during dev without creating ORFS pull requests(or it gets awkward).
end-user uses naja. This really requires that the dependencyinstaller in ORFS is updated + either my PR above + a tweak or something like this PR.

I looked at the numbers and until naja makes a bigger splash (currently low single digit percent on most designs), I don't think there will be a rush of ORFS users wanting to use it. naja does not yet support register duplication and cloning, which is the big missing feature for megaboom(according to BOOM documentation).
So there's time and I think supporting naja CI with minimal and changes to ORFS is the best next step.

maliberty:
"naja needs some ORFS test designs in CI in their own git repository" is easily kept in a fork.

oharboe:
many ways to skin a cat.

PR #2386, opened 2024-09-26
makefile: support specifying input to do-6_report

oharboe:
It can be useful to "flush the pipe" with information from earlier stages in the flow to reduce turnaround times. Although the result is not valid, it can still find problems with the flow more quickly.

To generate a report based upon information available at place time, run:

make FINAL_FILE_STEM=3_place do-6_report

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -148,6 +148,7 @@ export EQUIVALENCE_CHECK ?= 0
 export CORE_UTILIZATION ?=
 export DIE_AREA ?=
 export CORE_AREA ?=
+export FINAL_FILE_STEM ?= 6_1_fill
 
 # If we are running headless use offscreen rendering for save_image
 ifeq ($(DISPLAY),)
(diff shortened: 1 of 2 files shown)


maliberty:
What do you get from running final on a db that isn't complete?  rcx won't do anything if the design isn't detailed routed.

oharboe:
> What do you get from running final on a db that isn't complete? rcx won't do anything if the design isn't detailed routed.

I can check for syntax errors and the like in scripts like https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/designs/asap7/mock-array/power.tcl

maliberty:
power.tcl is only used from the design specific power target and has nothing to do with final_report.tcl.  Rather than hard code 6_final.v you can make that configurable.

maliberty:
You could make a target like "# Write a def for the corresponding odb" for .v to generate it at any step.

oharboe:
.spef files are generated in final_report.tcl

maliberty:
> .spef files are generated in final_report.tcl

No data will be generated if you haven't done detailed routing.

oharboe:
I see. Examining a different path forward.

PR #2393, opened 2024-09-28
variables: formalize variables and make metainformation machine readable

oharboe:
This inital version is just to decide on the format and that we want ORFS to provide this information.

Extending number of variables and using them(AutoTuner, bazel-orfs, consistency checks, generating docs, etc.) can come in followup PRs.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -109,46 +109,6 @@ DESIGN_CONFIG ?= ./designs/nangate45/gcd/config.mk
 # this file.
 include $(DESIGN_CONFIG)
 
-# For instance Bazel needs artifacts (.odb and .rpt files) on a failure to
-# allow the user to save hours on re-running the failed step locally, but
-# when working with a Makefile flow, it is more natural to fail the step
-# and leave the user to manually inspect the logs and artifacts directly via
-# the file system.
-#
-# Set to 1 to change the behavior to generate artifacts upon failure to
-# e.g. do a global route. The exit code will still be non-zero on all other
-# failures that aren't covered by the "useful to inspect the artifacts on
-# failure" use-case.
-#
-# Example: just like detailed routing, a global route that fails with congestion, is not
-# a build failure(as in exit code non-zero), it is a successful(as in zero exit code)
-# global route that produce reports detailing the problem.
-#
-# Detailed route will not proceed, if there is global routing congestion
-#
-# This allows build systems, such as bazel, to create artifacts for global
-# and detailed route, even if the operation had problems, without having
-# know about the semantics between global and detailed route.
-#
-# Considering that global and detailed route can run for a long time and
-# use a lot of memory, this allows inspecting results on a laptop for
-# a build that ran on a server.
-export GENERATE_ARTIFACTS_ON_FAILURE ?= 0
-
-# Default TNS_END_PERCENT value for post CTS timing repair
-# Try fixing all violating endpoints by default (reduce to 5% for runtime)
-export TNS_END_PERCENT ?=100
-
-# Default routing layer adjustment
-export ROUTING_LAYER_ADJUSTMENT ?= 0.5
-export RECOVER_POWER ?= 0
-export SKIP_INCREMENTAL_REPAIR ?= 0
-export DETAILED_METRICS ?= 0
-export EQUIVALENCE_CHECK ?= 0
-export CORE_UTILIZATION ?=
-export DIE_AREA ?=
-export CORE_AREA ?=
-
 # If we are running headless use offscreen rendering for save_image
 ifeq ($(DISPLAY),)
 export QT_QPA_PLATFORM ?= offscreen
@@ -199,6 +159,8 @@ export WORK_HOME     ?= .
 export UTILS_DIR     ?= $(FLOW_HOME)/util
 export SCRIPTS_DIR   ?= $(FLOW_HOME)/scripts
 export TEST_DIR      ?= $(FLOW_HOME)/test
+
+$(foreach line,$(shell $(SCRIPTS_DIR)/defaults.py),$(eval export $(line)))
 
 PUBLIC=nangate45 sky130hd sky130hs asap7 ihp-sg13g2 gf180
 
--- a/flow/scripts/defaults.py
+++ b/flow/scripts/defaults.py
@@ -1,0 +1,15 @@
+#!/usr/bin/env python3
+
+import os
+import yaml
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+yaml_path = os.path.join(dir_path, "defaults.yaml")
+with open(yaml_path, "r") as file:
+    data = yaml.safe_load(file)
+
+for key, value in data.items():
+    if value["value"] is None:
+        continue
+    print(f'{key}?={value["value"]}')
--- a/flow/scripts/defaults.yaml
+++ b/flow/scripts/defaults.yaml
@@ -1,0 +1,78 @@
+GENERATE_ARTIFACTS_ON_FAILURE:
+  description: >
+    For instance Bazel needs artifacts (.odb and .rpt files) on a failure to
+    allow the user to save hours on re-running the failed step locally, but
+    when working with a Makefile flow, it is more natural to fail the step
+    and leave the user to manually inspect the logs and artifacts directly via
+    the file system.
+
+    Set to 1 to change the behavior to generate artifacts upon failure to
+    e.g. do a global route. The exit code will still be non-zero on all other
+    failures that aren't covered by the "useful to inspect the artifacts on
+    failure" use-case.
+
+    Example: just like detailed routing, a global route that fails with congestion, is not
+    a build failure(as in exit code non-zero), it is a successful(as in zero exit code)
+    global route that produce reports detailing the problem.
+
+    Detailed route will not proceed, if there is global routing congestion
+
+    This allows build systems, such as bazel, to create artifacts for global
+    and detailed route, even if the operation had problems, without having
+    know about the semantics between global and detailed route.
+
+    Considering that global and detailed route can run for a long time and
+    use a lot of memory, this allows inspecting results on a laptop for
+    a build that ran on a server.
+  value: 0
+  stages:
+    - all
+
+TNS_END_PERCENT:
+  description: >
+    Default TNS_END_PERCENT value for post CTS timing repair.
+    Try fixing all violating endpoints by default (reduce to 5% for runtime).
+  value: 100
+  stages:
+    - crt
+    - floorplan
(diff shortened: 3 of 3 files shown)


maliberty:
It is a nice idea.  I assume you plan to flesh out stages which are currently ignored.  It reminds me of SiC (@gadfort thoughts?).

oharboe:
My thinking would be to add important variables over time.

Still pondering the idea.

bazel-orfs has some of this information, though ideally, ORFS would have complete info here, probably with some enforcement as to visibility of variables at various stages.

flow/Makefile
```diff
@@ -109,45 +109,8 @@ DESIGN_CONFIG ?= ./designs/nangate45/gcd/config.mk
...
-# Default routing layer adjustment
-export ROUTING_LAYER_ADJUSTMENT ?= 0.5
-export RECOVER_POWER ?= 0
-export SKIP_INCREMENTAL_REPAIR ?= 0
-export DETAILED_METRICS ?= 0
-export EQUIVALENCE_CHECK ?= 0
-export CORE_UTILIZATION ?=
-export DIE_AREA ?=
-export CORE_AREA ?=
+DEFAULTS := $(shell python defaults.py)
```
gadfort: `python` and `python3` are not interchangeable on all machines. I would suggest making `default.py` executable and add `#!`

flow/Makefile
```diff
@@ -109,45 +109,8 @@ DESIGN_CONFIG ?= ./designs/nangate45/gcd/config.mk
...
+$(foreach line,$(DEFAULTS),$(eval export $(line)))
```
gadfort: why not just do `$(foreach line,$(shell ./defaults.py),$(eval export $(line)))`? the `$DEFAULTS` doesn't seem needed.

flow/defaults.py
```diff
@@ -0,0 +1,14 @@
+import os
```
gadfort: add `#!/usr/bin/env python3`

flow/defaults.yaml
```diff
@@ -0,0 +1,78 @@
...
+    Detailed route will not proceed, if there is global routing congestion
+
+    This allows build systems, such as bazel, to create artifacts for global
+    and detailed route, even if the operation had problems, without having
+    know about the semantics between global and detailed route.
+
+    Considering that global and detailed route can run for a long time and
+    use a lot of memory, this allows inspecting results on a laptop for
+    a build that ran on a server.
+  value: 0
```
gadfort: `true/false`, feels a lot more natural than `0/?`

flow/defaults.yaml
```diff
@@ -0,0 +1,78 @@
...
+  stages:
+    - grt
+DETAILED_METRICS:
+  value: 0
+  stages:
+    - cts
+    - grt
+EQUIVALENCE_CHECK:
+  value: 0
```
gadfort: true/false

gadfort (commented):
@maliberty in SC we use the variables to be able to self document them, so that is a good idea. I really don't see a reason to not go ahead and move all variables into this, if this is the route to go towards (doing it piecemeal seems like a tortured approach).
One issue I do have is the use of yaml (python does not natively support yaml), so this would require users to install yet another dependency (or it is installed via a python distribution).
```
pip install pyyaml
```
This is part of the reason why we use JSON since that is supported by python3 without any dependencies installed.

flow/defaults.yaml
oharboe: ORFS uses 0/1 today in .tcl

flow/defaults.yaml
oharboe: 0/1 in tcl. of course true/false works too, I'm just inclined to let the dog lie.

oharboe:
> @maliberty in SC we use the variables to be able to self document them, so that is a good idea. I really don't see a reason to not go ahead and move all variables into this, if this is the route to go towards (doing it piecemeal seems like a tortured approach). One issue I do have is the use of yaml (python does not natively support yaml), so this would require users to install yet another dependency (or it is installed via a python distribution).
> 
> ```
> pip install pyyaml
> ```
> 
> This is part of the reason why we use JSON since that is supported by python3 without any dependencies installed.

Trying to do it all at once solves one problem and introduces another: who's going to do the job and it will be somewhat longlived task and the PR will have to be merged with master several times.

I don't see the exact reasons why doing it in several chunks is "tortured", could you clarify?

oharboe:
I like yaml because it is easier to read for humans.

yaml is very nearly documentation, .json is an eyesore.

ORFS depends on Python and probably some non standard packages(?) so adding one more package is no big deal. Needs verification.

I'll work through the good suggestions from @gadfort and update the PR and then we can see where we are.

oharboe:
Check out human readability of .json vs. yaml....

maliberty:
> I like yaml because it is easier to read for humans.

It also has a bunch of ambiguities (see https://noyaml.com/ https://github.com/cblp/yaml-sucks).  If we are adding a dependency then https://toml.io/en/ is a another choice.

oharboe:
Nice!

oharboe:
On second thought on toml: familiarity counts too. I'm not excited about being an early adopter, haven't seen it in other projects.

I'm happy with any choice that the group makes here though. We're talking dictionaries, lists and simple strings only, so either json or yaml would be fine by me.

maliberty:
I don't care that strongly.  We already have python package dependencies.

oharboe:
FYI pyyaml is already a dependency https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/53873f6b5a049a999386861e7a9ee80aa61080c5/etc/DependencyInstaller.sh#L32

oharboe:
@maliberty @gadfort I used .toml instead. It is a nice format with less ambiguity and it supports comments... One more dependency doesn't seem like it changes anything...

Thoughts?

oharboe:
@luarss FYI, this would enable you to create a simple and robust test here

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/53873f6b5a049a999386861e7a9ee80aa61080c5/tools/AutoTuner/src/autotuner/distributed.py#L402

maliberty:
I looked a bit further and learned that toml has been abandoned by its author (https://github.com/avakar/pytoml/issues/15#issuecomment-217739462) so I guess it is not a good choice.  Yaml is more common so I guess we'll just live with its warts.

oharboe:
> I looked a bit further and learned that toml has been abandoned by its author ([avakar/pytoml#15 (comment)](https://github.com/avakar/pytoml/issues/15#issuecomment-217739462)) so I guess it is not a good choice. Yaml is more common so I guess we'll just live with its warts.

Force pushed an update where I got rid of json and toml experiments.

oharboe:
@maliberty Rather than going down the path of automatically generated documentation from default.yaml(which I think is documentation drivel), I think that the documentation should be stripped of the variables explanations and that it should all be moved into defaults.yaml and a reference to defaults.yaml should be left in the documentation.

.yaml files are readable enough I really don't think a variable reference in .md format is that valuable. For anyone diving into these details, they really are going to have to read the .tcl files too for all the stages to really understand what is going on.

I think we're just creating problems for our future selves and for users by creating false expectations that reading .md files is enough. The user's expectations should be adjusted: There's nothing for it. You should, sooner rather than later, read the .tcl scripts to really understand everything that is going on in all the stages that you care about and read up on the variables need to use in the requisite .tcl scripts and  defaults.yaml.

maliberty:
I think it is ok so long as the docs clearly direct users to the yaml file and no information is lost.  That said I think there are plenty of black box users for all flows.

oharboe:
so there is something to be said for docs.py in this PR?

maliberty:
> so there is something to be said for docs.py in this PR?

You get nicer formatting with markdown so if it isn't hard to do it is nicer than reading yaml.

oharboe:
see example in PR.

CI would need to be updated to enforce defaults.md to be up to date (run scripts/docs.py and check for now diff against head)

oharboe:
@maliberty I think that if we're going to support updating .md files automatically, then docs.py should be updated to do surgery(modify) on the existing .md files so that the variables are documented in a meaningful context rather than being a separate file. Though some CI work is needed as I mentioned above to enforce updates of .md files in git. 

.md files can't be generated dynamically, they are served up from git, so anything generated would have to be generated and then put into git commits.

maliberty:
> CI would need to be updated to enforce defaults.md to be up to date (run scripts/docs.py and check for now diff against head)

I guess an action like https://github.com/The-OpenROAD-Project/OpenROAD/blob/26edfa870bdf2356eec343679c7ff90c212ca363/.github/workflows/github-actions-are-odb-files-generated.yml could be used for that.

oharboe:
> > CI would need to be updated to enforce defaults.md to be up to date (run scripts/docs.py and check for now diff against head)
> 
> I guess an action like https://github.com/The-OpenROAD-Project/OpenROAD/blob/26edfa870bdf2356eec343679c7ff90c212ca363/.github/workflows/github-actions-are-odb-files-generated.yml could be used for that.

Yes.

Though that is for followup PRs that can linger longer and we have more time to test.

I want to do suregery(modify) to the current .md files and check that they have been updated after updates to defaults.yaml in a CI along the lines that you show above.

maliberty (approved):

maliberty:
This is a partial step in this direction.  I just want to make sure you are committed to finishing the job as I don't want to end up in "no man's land".

oharboe:
I will add some tests & .md surgery and add a dozen common variables that I am familiar with and those that I run into and use/become familiar with.

I will leave the rest of the variables to be added by the community as well as updating AutoTuner and other users.

maliberty:
> I will add some tests & .md surgery and add a dozen common variables that I am familiar with and those that I run into and use/become familiar with.
> 
> I will leave the rest of the variables to be added by the community as well as updating AutoTuner and other users.

This sounds like a recipe for it to be half done eternally.  If you want to proceed then can we plan some of @jeffng-or  time to finish this off?

oharboe:
You mean you want a first pass to convert all variables known to man?

maliberty:
> You mean you want a first pass to convert all variables known to man?

No I want what I said "I just want to make sure you are committed to finishing the job".  I doesn't all have to be in this PR but I do want to have a plan to get there.

oharboe:
> > You mean you want a first pass to convert all variables known to man?
> 
> No I want what I said "I just want to make sure you are committed to finishing the job". I doesn't all have to be in this PR but I do want to have a plan to get there.

Understood.

Things are never really finished though...

My question/request is that we define "finished" and then we talk about how to get there.

maliberty:
Discussed and agreed to a plan.

PR #2394, opened 2024-09-29
[Autotuner] Plotting utility + test

luarss:
- Plotting utility
- README

### DIFF
--- a/docs/user/InstructionsForAutoTuner.md
+++ b/docs/user/InstructionsForAutoTuner.md
@@ -148,7 +148,7 @@ python3 -m autotuner.distributed --design gcd --platform sky130hd \
 #### Plot images
 
 After running an AutoTuner experiment, you can generate a graph to understand the results better.
-The graph will show one of the metrics listed below progression over the execution of the experiment.
+The graph will show the progression of one metric (see list below) over the execution of the experiment.
 
 - QoR
 - Runtime per trial
--- a/docs/user/InstructionsForAutoTuner.md
+++ b/docs/user/InstructionsForAutoTuner.md
@@ -145,6 +145,18 @@ python3 -m autotuner.distributed --design gcd --platform sky130hd \
                        sweep
 ```
 
+#### Plot images
+
+After running the autotuner experiments, you can visualize the results as follows.
+Currently, we support the following metrics:
+- QoR
+- Runtime per trial
+- Clock Period
+- Worst slack
+
+```shell
+python3 utils/plot.py --results_dir <your-autotuner-result-path>
+```
 
 ### Google Cloud Platform (GCP) distribution with Ray
 
--- a/docs/user/InstructionsForAutoTuner.md
+++ b/docs/user/InstructionsForAutoTuner.md
@@ -147,8 +147,9 @@ python3 -m autotuner.distributed --design gcd --platform sky130hd \
 
 #### Plot images
 
-After running the autotuner experiments, you can visualize the results as follows.
-Currently, we support the following metrics:
+After running an AutoTuner experiment, you can generate a graph to understand the results better.
+The graph will show one of the metrics listed below progression over the execution of the experiment.
+
 - QoR
 - Runtime per trial
 - Clock Period
--- a/flow/test/test_autotuner.sh
+++ b/flow/test/test_autotuner.sh
@@ -9,9 +9,9 @@ cd ../
 . ./tools/AutoTuner/setup.sh
 
 # remove dashes and capitalize platform name
-PLATFORM=${PLATFORM//-/}
+LOWERCASE_PLATFORM=${PLATFORM//-/}
 # convert to uppercase
-PLATFORM=${PLATFORM^^}
+PLATFORM=${LOWERCASE_PLATFORM^^}
 
 echo "Running Autotuner smoke tune test"
 python3 -m unittest tools.AutoTuner.test.smoke_test_tune.${PLATFORM}TuneSmokeTest.test_tune
@@ -31,7 +31,7 @@ if [ "$PLATFORM" == "asap7" ] && [ "$DESIGN_NAME" == "gcd" ]; then
 fi
 
 echo "Running Autotuner plotting smoke test"
-all_experiments=$(ls -d ./flow/logs/${PLATFORM}/${DESIGN_NAME}/*/)
+all_experiments=$(ls -d ./flow/logs/${LOWERCASE_PLATFORM}/${DESIGN_NAME}/*/)
 if [ -z "$all_experiments" ]; then
   echo "No experiments found for plotting"
   exit 0
--- a/flow/test/test_autotuner.sh
+++ b/flow/test/test_autotuner.sh
@@ -26,11 +26,7 @@ if [ "$PLATFORM" == "asap7" ] && [ "$DESIGN_NAME" == "gcd" ]; then
 fi
 
 echo "Running Autotuner plotting smoke test"
-all_experiments=$(ls -d ./flow/logs/${PLATFORM}/${DESIGN_NAME}/*/)
-if [ -z "$all_experiments" ]; then
-  echo "No experiments found for plotting"
-  exit 0
-fi
+all_experiments=$(ls -d ./flow/logs/${PLATFORM}/${DESIGN_NAME}/smoke-test-tune*)
 all_experiments=$(basename -a $all_experiments)
 for expt in $all_experiments; do
   python3 tools/AutoTuner/src/autotuner/utils/plot.py \
--- a/flow/test/test_autotuner.sh
+++ b/flow/test/test_autotuner.sh
@@ -26,7 +26,7 @@ if [ "$PLATFORM" == "asap7" ] && [ "$DESIGN_NAME" == "gcd" ]; then
   echo "Running Autotuner ref file test (only once)"
   python3 -m unittest tools.AutoTuner.test.ref_file_check.RefFileCheck.test_files
 
-  echo "Running Autotuner plotting smoke test (only once)"`
+  echo "Running Autotuner plotting smoke test (only once)"
   latest_folder=$(ls -dt ./flow/logs/asap7/gcd/test-tune*/ | tail -n 1)
   python3 tools/AutoTuner/src/autotuner/utils/plot.py --results_dir $latest_folder
 
--- a/flow/test/test_autotuner.sh
+++ b/flow/test/test_autotuner.sh
@@ -26,6 +26,10 @@ if [ "$PLATFORM" == "asap7" ] && [ "$DESIGN_NAME" == "gcd" ]; then
   echo "Running Autotuner ref file test (only once)"
   python3 -m unittest tools.AutoTuner.test.ref_file_check.RefFileCheck.test_files
 
+  echo "Running Autotuner plotting smoke test (only once)"`
+  latest_folder=$(ls -dt ./flow/logs/asap7/gcd/test-tune*/ | tail -n 1)
+  python3 tools/AutoTuner/src/autotuner/utils/plot.py --results_dir $latest_folder
+
   echo "Running AutoTuner resume test (only once)"
   python3 -m unittest tools.AutoTuner.test.resume_check.ResumeCheck.test_tune_resume
 fi
--- a/flow/test/test_autotuner.sh
+++ b/flow/test/test_autotuner.sh
@@ -8,11 +8,6 @@ cd ../
 ./tools/AutoTuner/installer.sh
 . ./tools/AutoTuner/setup.sh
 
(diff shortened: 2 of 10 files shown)


docs/user/InstructionsForAutoTuner.md
```diff
@@ -142,6 +142,18 @@ python3 distributed.py --design gcd --platform sky130hd \
                        sweep
 ```
 
+#### Plot images
+
+After running the autotuner experiments, you can visualize the results as follows.
+Currently, we support the following metrics:
```
vvbandeira:
```suggestion
After running an AutoTuner experiment, you can generate a graph to understand the results better.
The graph will show one of the metrics listed below progression over the execution of the experiment.
```

tools/AutoTuner/src/autotuner/utils/plot.py
```diff
@@ -0,0 +1,102 @@
...
+        plot(df, key)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Plot AutoTuner results.")
+    parser.add_argument(
+        "results_dir",
+        type=str,
+        help="Directory containing the results.",
+        default="../../../../../flow/logs/asap7/gcd/test-tune-2024-09-17-12-00-44",
```
vvbandeira:
```suggestion
```
Remove hardcoded.
This argument should be required.

vvbandeira (changes requested):

docs/user/InstructionsForAutoTuner.md
```diff
@@ -146,6 +146,19 @@ python3 distributed.py --design gcd --platform sky130hd \
                        sweep
 ```
 
+#### Plot images
+
+After running an AutoTuner experiment, you can generate a graph to understand the results better.
+The graph will show one of the metrics listed below progression over the execution of the experiment.
```
vvbandeira:
```suggestion
The graph will show the progression of one metric (see list below) over the execution of the experiment.
```

vvbandeira (changes requested):
Please add a call to this script in the `test_helper.py` so we test this script.

flow/test/test_autotuner.sh
```diff
@@ -20,28 +22,32 @@ python3 -m unittest tools.AutoTuner.test.smoke_test_sweep.${PLATFORM}SweepSmokeT
...
+#   cd "$script_dir"/../../
+#   latest_image=$(./etc/DockerTag.sh -dev)
+#   echo "ORFS_VERSION=$latest_image" > ./tools/AutoTuner/.env
+#   cd ./tools/AutoTuner
+#   docker compose up --wait
+#   docker compose exec ray-worker bash -c "cd /OpenROAD-flow-scripts/tools/AutoTuner/src/autotuner && \
+#     python3 distributed.py --design gcd --platform asap7 --server 127.0.0.1 --port 10001 \
+#     --config ../../../../flow/designs/asap7/gcd/autotuner.json tune --samples 1"
+#         docker compose down -v --remove-orphans
+# fi
```
vvbandeira: Remove?

tools/AutoTuner/src/autotuner/utils/plot.py
```diff
@@ -0,0 +1,106 @@
...
+                _dict["trial_id"] = re.search(AT_REGEX, fname).group(1)
+            with open(fname.replace("params.json", "metrics.json"), "r") as f:
+                metrics = json.load(f)
+                ws = metrics["finish"]["timing__setup__ws"]
+                metrics["worst_slack"] = ws
+                _dict.update(metrics)
+            params.append(_dict)
+        except Exception as e:
+            print(f"Error in {fname}: {e}")
+            continue
```
vvbandeira: Should we keep a list of the files that were not processed so we can print them at the very end in order for the error not to be buried in the logs? And should the script fail silently in this case?

tools/AutoTuner/src/autotuner/utils/plot.py
```diff
@@ -0,0 +1,106 @@
...
+
+def plot(df: pd.DataFrame, key: str):
+    # Plot box plot and time series plot for key
+    fig, ax = plt.subplots(1, figsize=(15, 10))
+    ax.scatter(df["timestamp"], df[key])
+    ax.set_xlabel("Time (s)")
+    ax.set_ylabel(key)
+    ax.set_title(f"{key} vs Time")
+    z = np.polyfit(df["timestamp"], df[key], 1)
+    p = np.poly1d(z)
```
vvbandeira: `z` I assume is the z-axis; what is `p`?

vvbandeira (changes requested):

tools/AutoTuner/src/autotuner/utils/plot.py
luarss: > And should the script fail silently in this case?

Just to clarify, this exception happens when any Autotuner run does not proceed to the "finish" stage, it does not fail silently

> Should we keep a list of the files that were not processed so we can print them at the very end in order for the error not to be buried in the logs? 

Sure I can add that.

tools/AutoTuner/src/autotuner/utils/plot.py
luarss: `z` is the np array of polynomial (ax+b) coefficients, and `p` is the actual polynomial function.

vvbandeira: Then let's use more descriptive names, maybe:
`z` --> `coeff`
`p` --> `poly_func`

tools/AutoTuner/src/autotuner/utils/plot.py
```diff
@@ -0,0 +1,109 @@
...
+        df["timestamp"], p(df["timestamp"]), "r--", label=f"y={z[0]:.2f}x+{z[1]:.2f}"
+    )
+    ax.legend()
+    fig.savefig(f"images/{key}.png")
+
+    plt.figure(figsize=(15, 10))
+    plt.boxplot(df[key])
+    plt.ylabel(key)
+    plt.title(f"{key} Boxplot")
+    plt.savefig(f"images/{key}-boxplot.png")
```
vvbandeira: Let's also parametrize the output folder to save images from CI into the reports folder for the design. This way it will be automatically saved to the build artifacts.

flow/test/test_autotuner.sh
```diff
@@ -20,28 +22,23 @@ python3 -m unittest tools.AutoTuner.test.smoke_test_sweep.${PLATFORM}SweepSmokeT
...
+all_experiments=$(ls -d ./flow/logs/${PLATFORM}/${DESIGN_NAME}/*/)
+if [ -z "$all_experiments" ]; then
+  echo "No experiments found for plotting"
+  exit 0
 fi
+all_experiments=$(basename -a $all_experiments)
+for expt in $all_experiments; do
+  python3 tools/AutoTuner/src/autotuner/utils/plot.py \
+    --platform ${PLATFORM} \
+    --design ${DESIGN_NAME} \
```
vvbandeira, line 34: At the beginning of this script, you are casting these values to be uppercase, which does not match the path since Linux is case-sensitive. From the CI:
```
02:32:36  Running Autotuner plotting smoke test
02:32:36  ls: cannot access './flow/logs/SKY130HD/gcd/*/': No such file or directory
02:32:36  No experiments found for plotting
02:32:36  + exit 0
```

vvbandeira (changes requested):

flow/test/test_autotuner.sh
vvbandeira: If we call the plot script and do not find data to plot, this should be an error, right?

luarss: Yup, these are 2 separate errors, will fix them.

flow/test/test_autotuner.sh
```diff
@@ -20,28 +22,23 @@ python3 -m unittest tools.AutoTuner.test.smoke_test_sweep.${PLATFORM}SweepSmokeT
 echo "Running Autotuner smoke tests for --sample and --iteration."
 python3 -m unittest tools.AutoTuner.test.smoke_test_sample_iteration.${PLATFORM}SampleIterationSmokeTest.test_sample_iteration
 
-if [ "$PLATFORM" == "asap7" ] && [ "$DESIGN" == "gcd" ]; then
+if [ "$PLATFORM" == "asap7" ] && [ "$DESIGN_NAME" == "gcd" ]; then
```
vvbandeira: The first check will never be true because you are casting to uppercase on line 14.
Would it be simpler/easier if we just use the platform name everywhere and just remove the dashes if that is not supported by the `unittest` Python lib?

flow/test/test_autotuner.sh
```diff
@@ -20,28 +22,23 @@ python3 -m unittest tools.AutoTuner.test.smoke_test_sweep.${PLATFORM}SweepSmokeT
...
-  docker compose up --wait
-  docker compose exec ray-worker bash -c "cd /OpenROAD-flow-scripts/tools/AutoTuner/src/autotuner && \
-    python3 distributed.py --design gcd --platform asap7 --server 127.0.0.1 --port 10001 \
-    --config ../../../../flow/designs/asap7/gcd/autotuner.json tune --samples 1"
-        docker compose down -v --remove-orphans
+echo "Running Autotuner plotting smoke test"
+all_experiments=$(ls -d ./flow/logs/${LOWERCASE_PLATFORM}/${DESIGN_NAME}/*/)
+if [ -z "$all_experiments" ]; then
+  echo "No experiments found for plotting"
+  exit 0
```
vvbandeira: This is still exiting with a success code.

tools/AutoTuner/src/autotuner/utils/plot.py
```diff
@@ -0,0 +1,195 @@
...
+        dir (str): The directory path containing the subdirectories with `progress.csv`, `params.json`, and `metrics.json` files.
+    Returns:
+        pd.DataFrame: A DataFrame containing the merged data from the progress, parameters, and metrics files.
+    """
+
+    # Concatenate progress DFs
+    progress_csvs = glob.glob(f"{dir}/*/progress.csv")
+    if len(progress_csvs) == 0:
+        print("No progress.csv files found.")
+        sys.exit(0)
```
vvbandeira: This should also be an error and exit with non-zero value.

tools/AutoTuner/src/autotuner/utils/plot.py
```diff
@@ -0,0 +1,195 @@
...
+
+    # Merge all dataframe
+    params_df = pd.DataFrame(params)
+    try:
+        progress_df = progress_df.merge(params_df, on="trial_id")
+    except KeyError:
+        print(
+            "Unable to merge DFs due to missing trial_id in params.json (possibly due to failed trials.)"
+        )
+        sys.exit(0)
```
vvbandeira: Same as above.

tools/AutoTuner/src/autotuner/utils/plot.py
```diff
@@ -0,0 +1,195 @@
...
+        df = df.rename(columns=rename_dict)
+        df = df.drop(columns=cols_to_remove)
+        df = df[df[METRIC] != 9e99]
+        df["timestamp"] -= df["timestamp"].min()
+        return df
+    except KeyError as e:
+        print(
+            f"KeyError: {e} in the DataFrame. Dataframe does not contain necessary columns."
+        )
+        sys.exit(0)
```
vvbandeira: Same as above.

vvbandeira (changes requested):

flow/test/test_autotuner.sh
luarss: The main concern is that I prefer name of smoke tests to be capitalized. We could also do something like `asap7SampleIterationSmokeTest`, but might be less readable

Perhaps a `UPPERCASE_PLATFORM` would make the distinction clearer?

flow/test/test_autotuner.sh
luarss: This is returning 0 because I didn't want to error out for smoke tests that are not suitable for plotting. I will modify it to only run `plot.py` for valid smoke tests then.

vvbandeira: We could then change line 31 to list either manually the tests we want or use a better regex in the experiment names to differentiate whether the plot is available or not.

flow/test/test_autotuner.sh
vvbandeira: Modifying PLATFORM or another variable already in use somewhere else in the code can easily lead to confusion. Other options could be `PLATFORM_UT` where UT is 'under test' or `TEST_PLATFORM`. That said, I don't see any issue with `asap7SampleIterationSmokeTest`.

vvbandeira (approved):

PR #2397, opened 2024-09-29
scripts: clean up cts

oharboe:


tools/AutoTuner/src/autotuner/distributed.py
```diff
@@ -397,10 +397,13 @@ def parse_config(config, path=os.getcwd()):
...
+            # FIXME there is no robust way to get this metainformation from
+            # ORFS about the variables, so disable this code for now.
+
             # Sanity check: ignore all flow variables that are not tunable
-            if key not in flow_variables:
-                print(f"[ERROR TUN-0017] Variable {key} is not tunable.")
-                sys.exit(1)
+            # if key not in flow_variables:
+            #     print(f"[ERROR TUN-0017] Variable {key} is not tunable.")
+            #     sys.exit(1)
```
maliberty, line 406: How does this problem manifest?

oharboe: When we modify the scripts/*.tcl so that the parsing of the scripts/*.tcl variables no longer find all the variables such as CTS_CLUSTER_SIZE, then AutoTuner mistakenly things it isn't a tunable parameter.

What is needed is metainformation about the variables so that AutoTuner can check if a string matches a variable name, ideally some extra features can be added like legal values/range(boolean, float, integer, ranges, enumerations, etc.).

maliberty (approved):

oharboe:
@maliberty merge despite the pr-merge CI failures due to network outage or retrigger build?

PR #2415, opened 2024-10-03
[Autotuner] Documentation fix for old file

luarss:
- removes missing file
- update paths in docs.


flow/util/distributed.py
maliberty, line 1: Is it intended that this file is removed?

luarss: Yes. That was the original autotuner code before we shifted it to `tools`

maliberty (approved):

PR #2416, opened 2024-10-03
docs: autogenerate flow variables table

oharboe:


oharboe:
@maliberty Something like this will be a first pass, but I need to flesh out the metadata a bit to get back to more or less where FlowVariables.md used to be at.

flow/scripts/variables.yaml
```diff
@@ -0,0 +1,577 @@
...
+POST_CTS_TCL:
+  description: >
+    Specifies a Tcl script with commands to run after CTS is completed.
+  stages:
+    - cts
+FASTROUTE_TCL:
+  description: >
+    Specifies a Tcl script with commands to run before FastRoute.
+  stages:
+    - grt
```
maliberty: I thought you would leave the stages as all by default.  This is needed in global_place.tcl too (and gui.tcl).

oharboe: yes, but I was looking at the documentation sections and pondering how to best represent things...

oharboe: KISS: refine stages later and leave variables in "all" section when stages section is not carefully curated.

oharboe:
@maliberty Thoughts? 

ci pr-merge seems unrelated.

oharboe:
@maliberty Flaky autotuner test in pr-merge? It went away last time with a retriggered build?

flow/scripts/variables.py
maliberty, line 1: Needs a general comment about its purpose.  The file name is not very descriptive (it generates documentation which you couldn't guess).

flow/scripts/stage_variables.py
maliberty, line 1: This is the set of variables *not* in the given stage.  The filename gives the opposite impression.

maliberty (changes requested):

maliberty:
image

is not very readable and also misleading (eg it suggests synthesis depends on the KLAYOUT_TECH_FILE which it clearly doesn't).

oharboe:
> image
> 
> is not very readable and also misleading (eg it suggests synthesis depends on the KLAYOUT_TECH_FILE which it clearly doesn't).

Fixed misleading bit: only list variables if they have the stage in "stages:".

As to more readable...

Suggestions?

This is a summary of variables for that stage.

maliberty:
> Fixed misleading bit: only list variables if they have the stage in "stages:".

Did you push this?

flow/scripts/variables.py
oharboe: fixed

flow/scripts/stage_variables.py
oharboe: fixed

oharboe:
> > Fixed misleading bit: only list variables if they have the stage in "stages:".
> 
> Did you push this?

No, I'm writing up comments as I fix, then I will push. There's no great workflow for where to keep such comments in "draft" until I have pushed...

oharboe:
@maliberty As to when re-review: I retrigger a review after updating the PR, so I guess that is the best that I know of w.r.t. the review being ready for another review. Generally ignore noise from github until then...

flow/scripts/gui-preload.tcl
maliberty, line 1: Not part of this PR?

oharboe: accident. Delete and force push or revert?

maliberty:
I think generating a bullet list of variables for each stage would be more readable.  If they could link back to their definition that would be helpful.  

I would be nice to have a list of variables that are "all" and those that are uncategorized.  It will be easier to push those forward if there is a list.

maliberty:
Once this is all settled I'll run a full CI as the erased envars need testing in all flows.

oharboe:
> I think generating a bullet list of variables for each stage would be more readable. 

That would give you doom scrolling, no?

flow/scripts/gui-preload.tcl
oharboe: deleted by force push.

oharboe:
> I think generating a bullet list of variables for each stage would be more readable. If they could link back to their definition that would be helpful.
> 
> I would be nice to have a list of variables that are "all" and those that are uncategorized. It will be easier to push those forward if there is a list.

Fixed

maliberty:
Secure CI started.

maliberty:
The CI is fine but the lint error needs fixing.

oharboe:
> The CI is fine but the lint error needs fixing.

Fixed.

oharboe:
I prefer to have this merged before I update docs in other PRs....

PR #2420, opened 2024-10-03
Pdn simplified

oharboe:


maliberty (approved):

oharboe:
@maliberty One more cleanup commit in same vein now that I saw mock-array passed.

PR #2426, opened 2024-10-07
[DO NOT REVIEW] make: add GUI_PRELOAD, slow stuff headless, fast in GUI

oharboe:


oharboe:
@maliberty What commands can I run to preload stuff into the GUI process other than report_checks?

maliberty:
The various heatmaps can be enabled from tcl.

oharboe:
> The various heatmaps can be enabled from tcl.

Great!

Example?


Also, clock tree? slack histogram?

maliberty:
https://openroad.readthedocs.io/en/latest/main/src/gui/README.html#set-heatmap

The other two are mostly about sta being updated.

oharboe:
> https://openroad.readthedocs.io/en/latest/main/src/gui/README.html#set-heatmap

How do I wait for rendering/preparation of heatmap to be complete?

This just enables them, no?

> 
> The other two are mostly about sta being updated.

So report_checks?

still a long wait in the GUI.

maliberty:
There will always be some wait as we have to build the boost rtrees when the GUI starts.   Where is this heading?  I don't get the purpose of this from the title.

oharboe:
> There will always be some wait as we have to build the boost rtrees when the GUI starts. Where is this heading? I don't get the purpose of this from the title.

Initially, I think the experience is better if the GUI isn't launched until it is ready. It is then a headless process running in the background on my computer.

Today I have to launch the GUI, click a button, wait 5 minutes, click another button, wait 5 minutes.

If it was better to wait from within the GUI, I would prefer to run builds from within the GUI, but clearly builds are better to run in a shell in the background.

I'm also interested in investigating if criu can be added to the bazel-orfs flow on top of this. This would make the GUI preparation become part of the artifacts generated for a project, further eliminating click+wait+click+wait times in the GUI.

oharboe:
Is there a tcl command to create boost rtrees before starting the GUI?

oharboe:
> There will always be some wait as we have to build the boost rtrees when the GUI starts. Where is this heading? I don't get the purpose of this from the title.

@jeffng-or Will be able to shed some light on the fun of the for (;;) {click+wait} experience with megaboom... :-)

maliberty:
> Is there a tcl command to create boost rtrees before starting the GUI?

No.  Trying to make the GUI fully active while not actually active is a weird use model.  If you are using criu  then this is all irrelevant as you can just start the GUI and then freeze it.   If not it is better to start the GUI and just iconify it while it starts up.

oharboe:
> > Is there a tcl command to create boost rtrees before starting the GUI?
> 
> No. Trying to make the GUI fully active while not actually active is a weird use model. If you are using criu then this is all irrelevant as you can just start the GUI and then freeze it. If not it is better to start the GUI and just iconify it while it starts up.

Weird as it is, I prefer to avoid click+wait+click+wait...

Is it easy to add a "gui::rtrees" command that builds rtress?

Starting the GUI is not great, because it solves one problem and introduces another w.r.t. criu. GUIs are more complicated(as in I get crashes) to save/restore than console apps. Better not to start GUI in the first place.

maliberty:
Isn't the whole point to be able to start the GUI?  If the GUI is unreliable I don't see how all this pre-work will solve that.

I agree that avoiding click-wait makes sense.  I just don't see why it should be done in text mode first.  Start the GUI, do a bunch of setup, then start using it seems reasonable.

oharboe:
> Isn't the whole point to be able to start the GUI? If the GUI is unreliable I don't see how all this pre-work will solve that.

suspending/dumping the GUI makes it unstable, starting it after a resumed console works fine.

criu is magic to me, save a process and resume it. I'm not at all surprised that restoring a GUI app is harder and more fraught with imponderables.

> I agree that avoiding click-wait makes sense. I just don't see why it should be done in text mode first. Start the GUI, do a bunch of setup, then start using it seems reasonable.

It avoids having to understand and fix the problems with dumping/restoring the OpenROAD GUI and also, it is a better match for a bazel scenario with artifacts where the warmup+dump is done remotely, stored in an artifact and restoring is done on a local interactive machine.

oharboe:
Hmm... how do I populate the heatmap?

```
$ make open_final
ODB_FILE=./results/nangate45/gcd/base/6_final.odb /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad -no_init -threads 16  /home/oyvind/OpenROAD-flow-scripts/flow/scripts/gui.tcl
OpenROAD v2.0-16259-g2e1ee9c70 
read_spef ./results/nangate45/gcd/base/6_final.spef
openroad> gui::set_heatmap Placement rebuild
openroad> gui::dump_heatmap Placement abc.csv
[ERROR GUI-0072] "Placement Density" is not populated with data.
GUI-0072
while evaluating gui::dump_heatmap Placement abc.csv
openroad> 
```

maliberty:
I don't want to try to turn every gui feature into a force load just to enable this use model.  If crui doesn't work then you should file an issue there.

oharboe:
> I don't want to try to turn every gui feature into a force load just to enable this use model. If crui doesn't work then you should file an issue there.

It is hardly every feature, I just need a way to prepare those things which can take a very long time for very large projects.

I think the exhaustive list is pretty much: fix what is there but doesn't work(heatmaps), endpoint slack histogram, clock tree viewer, timing update.

What else has long click+wait times for large designs?

Also, when studying criu, it is not everything that can be expected to work and some of the fixes are going to have to be in openroad anyway(segfaults when OpenROAD is tortured in some unusual, but strictly speaking legal way).

This quickly becomes apparent when I google/chatgpt.

The recommendation is to do what I'm doing; avoid launching the GUI until after restore.

oharboe:
@maliberty Filed https://github.com/The-OpenROAD-Project/OpenROAD/issues/5897 w.r.t. heatmaps.

oharboe:
preload of timing update would be useful:

image

oharboe:
Fails:

```
openroad> save_clocktree_image -clock clock blah.png
[INFO GUI-0001] Command save_clocktree_image is not usable in non-GUI mode
```

Works:

```
openroad> gui::show "save_clocktree_image -clock core_clock blah.png" false
[WARNING GUI-0076] QSocketNotifier: Can only be used with threads started with QThread
openroad> 
```

However, that fixes one problem and creates another:

```
$ criu dump --unprivileged -t 3286617 -D xxx --shell-job --ext-unix-sk
Warn  (compel/arch/x86/src/lib/infect.c:367): Will restore 3286703 with interrupted system call
Warn  (compel/arch/x86/src/lib/infect.c:367): Will restore 3286706 with interrupted system call
Warn  (compel/arch/x86/src/lib/infect.c:367): Will restore 3286708 with interrupted system call
Warn  (compel/arch/x86/src/lib/infect.c:367): Will restore 3286709 with interrupted system call
Error (criu/sk-unix.c:881): unix: Can't dump half of stream unix connection. name: (null); peer name: /home/oyvind/.cache/ibus/dbus-MQkShiqp
Error (criu/cr-dump.c:2098): Dumping FAILED.
```


image

image

oharboe:
Out of sight out of mind for now. Two other much less ambitious PRs  filed.

PR #2435, opened 2024-10-09
makefile: gui_, bring up gui when it is responsive

oharboe:
Do not bring up GUI while it is unresponsive and the OS is complaining about it and there's nothing for the user to do, except to comfort the OS that the application has not crashed by clicking on the "don't close the application" button.

image

After this change, even for a large design, when the GUI comes up, "Design loading..." is only shown for a very brief time and the GUI is immediately responsive:

image


flow/Makefile
```diff
@@ -571,9 +571,11 @@ open_$(1): open_$(2)
 endef
 
 define OPEN_GUI
-.PHONY: $(1)_$(2)
-$(1)_$(2):
-	$(3)=$(RESULTS_DIR)/$(2) $(4) $(SCRIPTS_DIR)/gui.tcl
+.PHONY: open_$(1) gui_$(1)
+open_$(1):
+	$(2)=$(RESULTS_DIR)/$(1) $(OPENROAD_NO_EXIT_CMD) $(SCRIPTS_DIR)/gui.tcl
```
maliberty: It is confusing the open_* now uses gui.tcl but doesn't show the gui.  Perhaps gui.tcl should have a more generic name like reload.tcl?

oharboe: open.tcl?

oharboe: Change default value of GUI_SHOW to 1 and keep name?

maliberty: open.tcl is fine.

oharboe:
@maliberty good to merge.

maliberty (approved):

PR #2436, opened 2024-10-09
makefile: market GUI_TIMING=0 better, useful for large designs

oharboe:
This feature is very useful for large designs and when studying the floorplan, but a bit buried.

Add a bit of 'marketing' in progress output.


flow/scripts/gui.tcl
```diff
@@ -49,6 +49,7 @@ proc read_timing {input_file} {
   fast_route
 }
 
-if {![env_var_equals GUI_NO_TIMING 1]} {
+if {[env_var_equals GUI_TIMING 1]} {
```
maliberty: Why change the use model?  The docs would also need updating.   I would just mention the opposite "Reading timing which takes a while for large designs.  If you don't need timing use GUI_NO_TIMING =1 for a faster load"

oharboe: Default on and no negative is better. 

Having user variables and options default to "on" rather than "off" with a negation in the name is generally better for clarity, usability, and reducing cognitive load. Here are a few reasons, along with references to well-known antipatterns:

### 1. **Clarity and Readability**
   - Defaulting to "on" with positive logic is often clearer to users because it avoids the mental burden of processing double negatives or reverse logic. For instance, an option like `ENABLE_DEBUG=true` is easier to understand than `DISABLE_DEBUG=false`. Double negatives can confuse users, especially in complex configurations.
   - **Antipattern: Inconsistent Naming or Double Negatives**
     - Having variables like `NOT_DISABLE_LOGGING=false` forces users to invert their thinking and adds unnecessary complexity. This is akin to the **Ambiguous Naming Antipattern**, where unclear variable names lead to misinterpretation.

### 2. **Reduced Cognitive Load**
   - Positive variable names with default "on" settings align with most users’ natural expectations. Users are accustomed to enabling or turning features on, so defaulting to "on" with easy-to-read names lowers the cognitive effort required to configure a system.
   - **Antipattern: Magic Numbers and Flags**
     - Using negated logic or flags that don't immediately communicate their purpose can confuse users, similar to the **Magic Numbers Antipattern**, where variables are not intuitively understood and need extra context to decipher.

### 3. **Simpler Defaults and Overrides**
   - When the default behavior is “on” or positive, it becomes easier to override specific configurations or customize behaviors by adding a clear negation. For example, `ENABLE_CACHE=true` is intuitive, and if a user wants to disable it, `ENABLE_CACHE=false` is straightforward. On the other hand, negated defaults (e.g., `DISABLE_CACHE=false`) require more mental gymnastics to reverse.
   - **Antipattern: Negative Booleans**
     - This resembles the **Negative Boolean Trap** antipattern, where using negation in boolean logic leads to confusion. Developers and users may accidentally misconfigure systems when they have to mentally invert logic.

### 4. **Error Prevention**
   - Having options default to "on" reduces the risk of unintended behavior because users are more likely to miss turning off a feature than forgetting to turn it on. Positive naming helps prevent common mistakes, whereas using negative defaults can result in overlooked issues during configuration.
   - **Antipattern: Hidden Failures**
     - This is similar to the **Hidden Failure Antipattern**, where configurations fail silently or the user’s intent isn’t properly captured, often due to confusing logic or defaults. Defaults that are "on" with positive names tend to make failures more apparent.

In short, using default "on" variables with positive naming improves clarity, reduces mistakes, and enhances overall user experience. It helps avoid antipatterns like **Ambiguous Naming**, **Negative Boolean Trap**, and **Hidden Failures** that can arise from using negations in configuration options.

maliberty: Should I post a chatgpt reply?

maliberty: You will have to update the docs then.

oharboe: Will do.

Re ChatGPT blurb; I did try to search for the specific antipattern w.r.t. negative names in variables, but what I can do on the phone is limited.

oharboe: Fixed

oharboe:
@maliberty Merged with master to fix conflict. Good to merge.

oharboe:
@maliberty merged with master. Good to go.

oharboe:
@maliberty I had to remerge this small change once, can you merge to avoid conflicts with coming PRs?

maliberty (approved):

PR #2440, opened 2024-10-10
docs: include a defaults column for the variables

oharboe:


oharboe:
@maliberty unrelated autotuner failure in pr-merge

maliberty:
@luarss @vvbandeira  do you see why the autotuner is failing here?  It is clearly unrelated to this doc change.

PR #2447, opened 2024-10-12
variables: ABSTRACT_SOURCE is only used when generating abstract

oharboe:


oharboe:
@maliberty more unrelated autotuner pr-merge failures. Good to merge I think.

maliberty (approved):

PR #2448, opened 2024-10-12
variables: flesh out SKIP_REPORT_METRICS and SKIP_INCREMENTAL_REPAIR

oharboe:


flow/scripts/variables.yaml
```diff
@@ -80,6 +84,13 @@ CORE_AREA:
 SKIP_REPORT_METRICS:
   description: >
     If set to 1, then metrics, report_metrics does nothing. Useful to speed up builds.
+  stages:
```
maliberty: Wouldn't this be all stages?

oharboe: No, not generate_abstract

maliberty (approved):

PR #2449, opened 2024-10-12
makefile: add do-synth-sdc for consistency

oharboe:


oharboe:
@hovind FYI

flow/Makefile
```diff
@@ -506,10 +506,14 @@ $(RESULTS_DIR)/1_synth.rtlil: $(YOSYS_DEPENDENCIES)
 $(RESULTS_DIR)/1_1_yosys.v: $(RESULTS_DIR)/1_synth.rtlil
 	$(UNSET_AND_MAKE) do-yosys-keep-hierarchy do-yosys
 
-$(RESULTS_DIR)/1_synth.sdc: $(SDC_FILE)
+.PHONY: do-synth-sdc
+do-synth-sdc:
```
maliberty, line 510: How is this different from `make synth`?

oharboe: "make synth" is the counterpart to do-synth, not related to .sdc.

synthesis does not, unlike openroad steps, write out an .sdc file.

I dont know if this difference is intentional.

It would be possible to cp .sdc file from within synthesis .tcl files, making it more consistent with openroad stages.

do-synth-sdc allows copying the .sdc file independent of the synthesis and this PR addresses this as a single concern

Do you envisage more changes w.r.t. synthesis and .tcl and .sdc to make it more consistent with openroad stages?

maliberty: It would be fine to copy it from within synth.  I don't see any use to having a target that only copies the file without running synthesis. 

Yosys is a bit simple in this regard and most proprietary tools do write a modified SDC.  I suppose the intention was to leave room for this possibility in the future but just do a simple cp now.  In fact almost nothing in the flow modifies the SDC.

oharboe: I will fix synth.sdc and simplify the makefile

PR #2454, opened 2024-10-13
makefile: prune transitive dependencies

oharboe:
Remove superfluous transititive dependencies.

Fewer questions(why are there some transitive dependencies here and there instead of everywhere or nowhere?) when maintaining the Makefile and also correct and unchanged behavior.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -651,11 +651,11 @@ $(eval $(call do-step,2_2_floorplan_io,$(RESULTS_DIR)/2_1_floorplan.odb $(IO_CON
 
 # STEP 3: Timing Driven Mixed Sized Placement
 #-------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


maliberty:
The .v is unneeded but the sdc and lib files seem needed (they are not saved in the odb).

oharboe:
> The .v is unneeded but the sdc and lib files seem needed (they are not saved in the odb).

This is about dependencies: the .sdc is always generated when the .odb dependency is generated. So superfluous transitive dependency.

maliberty:
Someone can manually modify the sdc and expect things to re-run.  You are eliding a real dependency not a transitive one.

oharboe:
> Someone can manually modify the sdc and expect things to re-run. You are eliding a real dependency not a transitive one.

They still do: if they modify the .sdc file output from the synthesis stage, then make continues with 2_1_floorplan.odb, which will cause the subsequent steps to re-run.

So these dependencies that are pruned here, have no effect on the dependency tree as seen by make.

oharboe:
Letting the sleeping dog lie.

PR #2445, opened 2024-10-12
gui: less click+wait, populate timing paths as part of loading timing

oharboe:
This makes the GUI more responsive for large designs when interacting with timing.

report_checks returns instantaneously.

Use GUI_TIMING=0 when timing is not needed.

PR #2456, opened 2024-10-13
gui/open: minimize surprises, handle errors in reading timing

oharboe:
read_timing can fail estimating parasitics for a failed global route .odb file and examining global route DRC .rpt files in the GUI, for instance.

The "surprise" for the user is that he has to type `gui::show`. This PR eliminates that surprise.

This is surprising because there is nothing in particular the user can do from the console, they probably have to open the GUI to examine the problem, such as loading DRC report from global routing.

Example output after fix(from MegaBoom):

```
$ tmp/make gui_grt
ODB_FILE=.//results/asap7/BoomTile/base/5_1_grt.odb /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad -no_init -threads 16  /home/oyvind/OpenROAD-flow-scripts/flow/scripts/open.tcl
OpenROAD v2.0-16443-g372a00c48 
estimate_parasitics -global_routing
[WARNING GRT-0097] No global routing found for nets.
[ERROR RSZ-0005] Run global_route before estimating parasitics for global routing.
Error reading timing: RSZ-0005
[WARNING GUI-0076] QSocketNotifier: Can only be used with threads started with QThread
```

### DIFF
--- a/flow/scripts/open.tcl
+++ b/flow/scripts/open.tcl
@@ -51,7 +51,9 @@ proc read_timing {input_file} {
 
 if {[env_var_equals GUI_TIMING 1]} {
   puts "GUI_TIMING=1 reading timing, takes a little while for large designs..."
-  read_timing $input_file
+  if {[catch {read_timing $input_file} errMsg]} {
+      puts "Error reading timing: $errMsg"
+  }
 }
 
 if {[env_var_equals GUI_SHOW 1]} {


maliberty:
I disagree with this.  It is an error and the timing will not be correct.  If you don't want the error then use GUI_TIMING=0

oharboe:
> I disagree with this.  It is an error and the timing will not be correct.  If you don't want the error then use GUI_TIMING=0

That is not the effect of this change, the error still stops the rest of the timing reading: the effect is that the script continues on to open the gui.

flow/scripts/open.tcl
```diff
@@ -51,7 +51,9 @@ proc read_timing {input_file} {
 
 if {[env_var_equals GUI_TIMING 1]} {
   puts "GUI_TIMING=1 reading timing, takes a little while for large designs..."
-  read_timing $input_file
+  if {[catch {read_timing $input_file} errMsg]} {
+      puts "Error reading timing: $errMsg"
+  }
 }
```
oharboe, line 58: Above the catch will report the error and continue execution, below we open the GUI, which we want to do even in the event of an error in read_timing, which is why the catch and continue above is what is needed.

Note that openroad -gui returns non-zero exit code, even when there are errors in the script.

If the user wants to run a script and have a non-zero exit code, then use `make run RUN_SCRIPT=foo.tcl` instead of `make open/gui_`

oharboe:
@maliberty pr-head passed, just an autotuner problem. Merge? 

Restores behavior on master: open GUI after displaying error.

```
ERROR: test_sample_iteration (tools.AutoTuner.test.smoke_test_sample_iteration.SKY130HDSampleIterationSmokeTest)
```

oharboe:
The new behavior is the desired behavior. Closing.

oharboe:
@maliberty I think that this PR should go in, it minimizes surprises for the user.  @jeffng-or Ran into it.

maliberty:
> @maliberty I think that this PR should go in, it minimizes surprises for the user.  @jeffng-or Ran into it.

What problem is that?  I think it creates surprises.

oharboe:
> > @maliberty I think that this PR should go in, it minimizes surprises for the user.  @jeffng-or Ran into it.
> 
> What problem is that? I think it creates surprises.

I tried to elaborate a bit on it in the PR comment: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2445

Not to put words into the mout of @jeffng-or, but I think that he might have a situation where:

1. there is an old problem that results in an error message
2. the GUI doesn't open

This is interpreted by @jeffng-or as that the error cause the GUI not to open, which is wrong. open.tcl *chose* not to open the GUI because there was an error, the error has nothing to do with whether the GUI can be opened or not.

I can't come up with a good reason not to open the GUI when there is an error here, it used to.

I just find this altogether confusing :-)

maliberty:
It seems like a better argument for reverting this whole path of delaying the GUI open.  Now you can't issue errors when they happen and convert them to ignorable puts later on.

oharboe:
Choise of poison is easy: I prefer a GUI that is responsive and ready when it opens.

oharboe:
I agree that it is a problem with this PR that it opens the GUI and "hides" the console error messages.

Although this gives the user the "prize" of opening the GUI, it hides the errors, which could cause other problems for the user.

PR #2459, opened 2024-10-14
ADDITIONAL_* variables.yaml stages

oharboe:
- Fix `make generate_abstract_issue`
- update ADDITTIONAL_* stages.

### DIFF
--- a/flow/scripts/variables.yaml
+++ b/flow/scripts/variables.yaml
@@ -438,13 +438,31 @@ ADDITIONAL_LEFS:
   description: >
     Hardened macro LEF view files listed here. The LEF information of the macros
     is immutable and used throughout all stages. Stored in the .odb file.
+  stages:
+  - synth
+  - floorplan
+  - place
+  - cts
+  - grt
+  - route
+  - final
 ADDITIONAL_LIBS:
   description: >
     Hardened macro library files listed here. The library information is immutable
     and used throughout all stages. Not stored in the .odb file.
+  stages:
+  - synth
+  - floorplan
+  - place
+  - cts
+  - grt
+  - route
+  - final
 ADDITIONAL_GDS:
   description: >
     Hardened macro GDS files listed here.
+  stages:
+  - final
 VERILOG_INCLUDE_DIRS:
   description: >
     Specifies the include directories for the Verilog input files.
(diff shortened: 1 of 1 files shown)


oharboe:
@vvbandeira @maliberty FYI, I have to mark PRs as non-draft for them to be built. Using a prominent header to mark as not ready for review/merge.

maliberty:
> @vvbandeira @maliberty FYI, I have to mark PRs as non-draft for them to be built. Using a prominent header to mark as not ready for review/merge.

That is part of our effort to reduce load on the jenkins server.  There are a lot of old PRs are they would wake up nightly to run pr-merge (due to master changing).

oharboe:
> > @vvbandeira @maliberty FYI, I have to mark PRs as non-draft for them to be built. Using a prominent header to mark as not ready for review/merge.
> 
> That is part of our effort to reduce load on the jenkins server. There are a lot of old PRs are they would wake up nightly to run pr-merge (due to master changing).

I see. merge-train scales much better. You can run full testing on merge-train and much lighter on PR level. Automatic bisection of PRs. If you want the code, you're welcome to it...

oharboe:
@maliberty merge?

oharboe:
@maliberty Merge? it just keeps rebuilding, even after it is green.

flow/scripts/variables.yaml
```diff
@@ -442,17 +446,23 @@ ADDITIONAL_FILES:
...
     Hardened macro library files listed here. The library information is immutable
     and used throughout all stages. Not stored in the .odb file.
+  stages:
+  - synth
+  - floorplan
+  - place
+  - cts
+  - grt
+  - route
+  - final
```
maliberty: Is there a more succinct way to say 'all'?  ADDITIONAL_LEFS lists no stages which I assume is equivalent though less obvious (is it an oversight or all).

oharboe: We haven't added that and I have been dithering on whether it is a good idea, it adds a tiny bit complication for a marginal benefit. Now we just leave it blank if we don't care enough to distinguish exactly when it is used and that or it must always be included.

The above is not "all", it is specifically all minus generate_abstract.

maliberty: Ok then it is wrong as write_abstract_lef depends on timing.

oharboe: *additional_libs* are not needed to write the abstract .lib file today, is it?

Could it be in the future?

(I'll add them, just checking for clarification).

oharboe: I see now that additional_libs is read used in generate_abstract. Removing specific stages for additional_libs, they are used pervasively...

maliberty (approved):

oharboe:
@maliberty autotuner failure

PR #2466, opened 2024-10-15
open/gui: when loading a failed global route, advertise congestion report

oharboe:
Instead of giving an error.

Single source of truth as to the name of the global routing congestion report.

### DIFF
--- a/flow/scripts/open.tcl
+++ b/flow/scripts/open.tcl
@@ -44,7 +44,8 @@ proc read_timing {input_file} {
     if { [grt::have_routes] } {
       log_cmd estimate_parasitics -global_routing
     } else {
-      puts "No global routing results: load $::global_route_congestion_report for details"
+      puts "No global routing results available, skipping estimate_parasitics"
+      puts "Load $::global_route_congestion_report for details"
     }
   } elseif {$design_stage >= 3} {
     log_cmd estimate_parasitics -placement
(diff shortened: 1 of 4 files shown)


flow/scripts/open.tcl
```diff
@@ -41,7 +41,11 @@ proc read_timing {input_file} {
   if {$design_stage >= 6 && [file exist $::env(RESULTS_DIR)/6_final.spef]} {
     log_cmd read_spef $::env(RESULTS_DIR)/6_final.spef
   } elseif {$design_stage >= 5} {
-    log_cmd estimate_parasitics -global_routing
+    if { [grt::have_routes] } {
+      log_cmd estimate_parasitics -global_routing
+    } else {
+      puts "No global routing results: load $::global_route_congestion_report for details"
```
maliberty: If GUI=1 then give a warning about missing parasitics for timing.

oharboe: Added warning, but why should GUI=1/0 be different in open/gui?

Shouldn't the only difference between open/gui_ be console vs. GUI?

maliberty: Sorry I meant GUI_TIMING=1

oharboe: Yes, GUI_TIMING=1/0 should have the same behavior. GUI vs. console; same behavior here.

maliberty (approved):

PR #2472, opened 2024-10-17
Tunable clock period

oharboe:
@jeffng-or @maliberty @luarss This makes ABC_CLOCK_PERIOD_IN_PS tunable, which will come in handy in MegaBoom.


maliberty:
ABC_CLOCK_PERIOD_IN_PS is a hack for abc limitations not something we should try to promote to a first class variable.  I have separately explain why I don't think this is a good approach.

oharboe:
@jeffng-or revisit when/if we explore parameters later.

PR #2474, opened 2024-10-17
Clockgate demo

widlarizer:
This PR shows possible usage of the yosys clockgate command to reduce area


maliberty:
I started a secure CI to see how well this works.

povik:
@maliberty It's configured for sky130hd only, secure CI won't tell us anything interesting yet

povik:
Someone with access to the private PDKs needs to fill in the names of the clock gating cells and their ports

maliberty:
I cancelled the private part but we'll still see for sky.  Can you configure the rest of the public PDKs?

widlarizer:
Here you go

PR #2492, opened 2024-10-22
variables: FLOORPLAN_DEF is also used in place stage

smunaut:
See flow/scripts/global_place_skip_io.tcl which references this variable too.
Without the change, pin definitions in the DEF are simply ignored.


oharboe (approved):
Nice catch!

maliberty (approved):

PR #2493, opened 2024-10-22
flow: don't use cell sg13g2_sdfbbp_1

htfab:
Yosys can only infer flip-flops where the `next_state` property in the liberty file is a single pin or its negation. For `sg13g2_sdfbbp_1` it is `(SCE*SCD)+(SCE'*D)` so it prints a warning and ignores the cell. By adding the cell to the `DONT_USE_CELLS` list we can get rid of the warning.

### DIFF
--- a/flow/platforms/ihp-sg13g2/config.mk
+++ b/flow/platforms/ihp-sg13g2/config.mk
@@ -31,7 +31,8 @@ export DONT_USE_CELLS += \
 sg13g2_lgcp_1 \
 sg13g2_sighold \
 sg13g2_slgcp_1 \
-sg13g2_dfrbp_2
+sg13g2_dfrbp_2 \
+sg13g2_sdfbbp_1
 
 
 # Define fill cells


maliberty:
@povik a scan flop seems like an ordinary thing - why can't yosys handle this cell?

povik:
@maliberty With YosysHQ/yosys#4652 we are extending the `next_state` handling to deal with arbitrary expressions, by which we should get rid of the warning in question.

Is there anything you would expect a synthesis tool to do with a scan flop other than ignore it?

maliberty:
I would hope it would quietly ignore it.  Putting it in the DONT_USE lists suggests that it should never be used (should DFT ignore it too?).

Is there an ETA to land that PR?

povik:
> Is there an ETA to land that PR?

My estimate is the start of December release.

maliberty:
@htfab are you ok to wait ~1 month to have this resolved by yosys?  If not we can merge but will need to track reverting this then.

htfab:
I'm ok with waiting. We are using it in Tiny Tapeout, but we already have a script that patches it on our side, so it just means we keep that script for somewhat longer.

maliberty:
This will resolve automatically when the right yosys update happens.

PR #2501, opened 2024-10-23
floorplan: HOLD_SLACK_MARGIN is now applied in floorplan.tcl

oharboe:
My suspicion is that the difference in floorplan.tcl vs other places repair_timing_helper was invoked was unintentional copy and paste error

### DIFF
--- a/flow/scripts/floorplan.tcl
+++ b/flow/scripts/floorplan.tcl
@@ -108,7 +108,7 @@ if { [env_var_equals REMOVE_ABC_BUFFERS 1] } {
   # remove buffers inserted by yosys/abc
   remove_buffers
 } else {
-  repair_timing_helper 0
+  repair_timing_helper
 }
 
 ##### Restructure for timing #########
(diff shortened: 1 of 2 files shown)


maliberty:
I think the intention is to prevent overfixing this early in the flow.  Perhaps you can just allow it for underfixing.

oharboe:
HOLD_SLACK_MARGIN is 0 by default, it seems.

Are you proposing having min of 0 and HOLD_SLACK_MARGIN for floorplan?

maliberty:
> HOLD_SLACK_MARGIN is 0 by default, it seems.
> 
> Are you proposing having min of 0 and HOLD_SLACK_MARGIN for floorplan?

Yes as we need to cover non-default positive values too.

oharboe:
A shared SETUP/HOLD_SLACK_MARGIN is looking increasingly dubious to me.

MegaBoom has WNS -1000 after floorplan and -4500 after cts.

@maliberty I think this needs some further pondering. This PR was an interesting datapoint: no regressions in all the land of ORFS CI tests, yet for larger designs there is much more going on.

PR #2511, opened 2024-10-27
global route: fix GENERATE_ARTIFACTS_ON_FAILURE use-case

oharboe:
the .sdc file must be written out alongside the .odb file and .rpt file in the failing test case whenGENERATE_ARTIFACTS_ON_FAILURE=1.

I suppose copying would have worked too, but write_sdc should be fine.

### DIFF
--- a/flow/scripts/global_route.tcl
+++ b/flow/scripts/global_route.tcl
@@ -31,6 +31,7 @@ proc global_route_helper {} {
     if {[expr !$::env(GENERATE_ARTIFACTS_ON_FAILURE) || \
         ![file exists $::global_route_congestion_report] || \
         [file size $::global_route_congestion_report] == 0]} {
+      write_sdc -no_timestamp $::env(RESULTS_DIR)/5_1_grt.sdc
       write_db $::env(RESULTS_DIR)/5_1_grt-failed.odb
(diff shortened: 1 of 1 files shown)


flow/scripts/global_route.tcl
```diff
@@ -31,6 +31,7 @@ proc global_route_helper {} {
     if {[expr !$::env(GENERATE_ARTIFACTS_ON_FAILURE) || \
         ![file exists $::global_route_congestion_report] || \
         [file size $::global_route_congestion_report] == 0]} {
+      write_sdc -no_timestamp $::env(RESULTS_DIR)/5_1_grt.sdc
```
maliberty: Why is this in the !$::env(GENERATE_ARTIFACTS_ON_FAILURE) branch?  You said you want it for GENERATE_ARTIFACTS_ON_FAILURE=1

oharboe: Failing eyesight, will fix

oharboe: Fixed.

I wish there was an easy and robust way to induce a global route failure in one of the ORFS designs...

maliberty (approved):

PR #2514, opened 2024-10-28
Variables global route args

oharboe:


flow/Makefile
```diff
@@ -160,7 +160,7 @@ export UTILS_DIR     ?= $(FLOW_HOME)/util
 export SCRIPTS_DIR   ?= $(FLOW_HOME)/scripts
 export TEST_DIR      ?= $(FLOW_HOME)/test
 
-$(foreach line,$(shell $(SCRIPTS_DIR)/defaults.py),$(eval export $(line)))
+$(foreach line,$(shell $(SCRIPTS_DIR)/defaults.py),$(eval export $(subst __SPACE__, ,$(line))))
```
maliberty, line 165: What is __SPACE__?

oharboe: makefile foreach whitespace escaping hell: no support for newlines, so handle this with a search & replace w/python.

maliberty: Does this produce a valid result in `make issue`?  It seems like removing the whitespace would be a problem.

oharboe: I'm not removing the whitespace, I'm adding it. We had no previous defaults that inclued whitespace, now we can have defaults in variables.py with space. Also, we have a CI regression test of make issue... So we should be good, or the CI make issue regression test is broken...

Also, I tried "make nuke floorplan floorplan_issue". GLOBAL_ROUTE_ARGS comes through correctly.

After adding the comment I get the following from github copilot when I asked it to explain:

This snippet from a Makefile is using a combination of `foreach`, `shell`, `eval`, and `subst` functions to process and export environment variables. Here's a step-by-step explanation:

1. **`$(shell $(SCRIPTS_DIR)/defaults.py)`**:
   - This part runs the `defaults.py` script located in the `$(SCRIPTS_DIR)` directory.
   - The output of this script is expected to be a list of environment variable assignments, with spaces replaced by `__SPACE__`.

2. **`$(foreach line, ..., $(eval export $(subst __SPACE__, ,$(line))))`**:
   - `foreach` iterates over each line of the output from the `defaults.py` script.
   - For each `line`, it performs the following:
     - `$(subst __SPACE__, ,$(line))`: Replaces `__SPACE__` with an actual space in the `line`.
     - `$(eval export ...)`: Evaluates the modified line as a Makefile command, effectively exporting the environment variable.

### Summary
This Makefile snippet runs a Python script to generate environment variable assignments, replaces `__SPACE__` with actual spaces, and exports these variables. This approach is used to handle whitespace issues in the `foreach` function.

maliberty (approved):

PR #2516, opened 2024-10-28
[DO NOT REVIEW] Partial automatic macro placement

oharboe:


oharboe:
revisit when #6033 is fixed

PR #2519, opened 2024-10-28
makefile: fix make do-floorplan and add test/test-do-stage.sh

oharboe:
@maliberty Can you add test/test-do-stage.sh to CI?


flow/test/test-do-stage.sh
maliberty, line 1: What is the purpose?

maliberty (approved):

PR #2520, opened 2024-10-28
makefile: fix one more do-floorplan gaffe

oharboe:
Missed one. I was sure that I ran `test/test-do-stage.sh` locally before previous PR to fix do-floorplan; apparently not.


maliberty:
Why do you need to do 2_floorplan.sdc separately?

maliberty (approved):

oharboe:
> Why do you need to do 2_floorplan.sdc separately?  

We need it because it copies 2_1_flooorplan.sdc to 2_floorplan.sdc

PR #2523, opened 2024-10-29
[DO NOT REVIEW]  Makefile: facilitate writing standalone metrics.tcl like scripts

oharboe:


oharboe:
@maliberty Thoughts?

maliberty:
It is less efficient to quit OR and then reload in order to compute metrics, particularly as we don't save the sta state.  I'm not sure if that's where you are headed.

oharboe:
> It is less efficient to quit OR and then reload in order to compute metrics, particularly as we don't save the sta state. I'm not sure if that's where you are headed.

I care about total running time. CTS completes in 600 seconds, whereas report_metrics for CTS takes an additional 2400 seconds. I prefer to run the metrics in parallel with the global route stage rather than in serial.

oharboe:
revisit later

PR #2524, opened 2024-10-29
[DO NOT REVIEW] metrics: make report_design_area consistent with other reporting

oharboe:
Requires https://github.com/The-OpenROAD-Project/OpenROAD/pull/6046

### DIFF
--- a/flow/scripts/report_metrics.tcl
+++ b/flow/scripts/report_metrics.tcl
@@ -226,9 +226,9 @@ proc report_metrics { stage when {include_erc true} {include_clock_skew true} } 
   }
 
   # TODO these only work to stdout, whereas we want to append to the $filename
-  puts "\n=========================================================================="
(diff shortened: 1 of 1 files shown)


flow/scripts/report_metrics.tcl
```diff
@@ -226,9 +226,9 @@ proc report_metrics { stage when {include_erc true} {include_clock_skew true} }
   }
 
   # TODO these only work to stdout, whereas we want to append to the $filename
```
maliberty, line 228: the comment would be obsolete with this change

oharboe: Will fix. I take it from the lack of other comments that there was no particular reason and that fixing it is just good gardening.

oharboe:
@maliberty let me know if you want to move forward with this

PR #2526, opened 2024-10-30
synth: nit, ephemral files belong in OBJECTS_DIR

oharboe:


maliberty:
I saw you opened & closed a PR to keep these files.  Are you wanting them to not be ephemeral?

oharboe:
> I saw you opened & closed a PR to keep these files.  Are you wanting them to not be ephemeral?

They should be emphemral, I briefly misread the code and mistook these for  the policy file, which is kept. These files are created as a workaround for this info not being available any more reasonable way.

maliberty (approved):

PR #2533, opened 2024-10-31
open: fix non-aborting incorrect operation when there is no routing

oharboe:
When using open.tcl from automation, if there is no routing available as of stage 5 and on, fail rather than log and continue.

The user experience when running 'make gui_grt' is that an error is shown and a console prompt. The user can type 'gui::show' to show the GUI.

Ideally, 'make gui_grt' would run with the GUI hidden and show the GUI upon first error or warning, but that requires an update to OpenROAD.

Example use-case:

```
$ make gui_grt
ODB_FILE=.//results/asap7/BoomTile/base/5_1_grt.odb /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad -no_init -threads 16  /home/oyvind/OpenROAD-flow-scripts/flow/scripts/open.tcl
OpenROAD v2.0-16827-g02ce90d85 
[WARNING GRT-0097] No global routing found for nets.
Error: open.tcl, 75 No global routing results available, skipping estimate_parasitics, load .//reports/asap7/BoomTile/base/congestion.rpt for details
[The GUI does not open, the console is shown, I can launch the GUI by running gui::show]
openroad> gui::show
```

### DIFF
--- a/flow/scripts/open.tcl
+++ b/flow/scripts/open.tcl
@@ -1,3 +1,13 @@
+# This script is used to launch the OpenROAD user ineterface
+# console as well as GUI mode. The user interface will remain
+# go from batch to interactive mode after this script has
+# been executed, whether it failed or not. For instance,
+# a failed global route run can be loaded to examine the
+# DRC errors.
+#
+# However, this script is also used to load the design and timing
+# in automation scripts, hence the script should fail upon
+# errors, not continue.
 source $::env(SCRIPTS_DIR)/util.tcl
 # Read liberty files
 source $::env(SCRIPTS_DIR)/read_liberty.tcl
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty Comments?

maliberty:
ORFS doesn't generate such a db so this is strictly a bazel issue.  If you are opening the -failed db then this would be odd as you intended to open a db without routing.

oharboe:
> ORFS doesn't generate such a db so this is strictly a bazel issue.  If you are opening the -failed db then this would be odd as you intended to open a db without routing.

Yes, I am opening a failed global route to look at .drc errors

maliberty:
We now write the routes to the failed db so this seems moot.  https://github.com/The-OpenROAD-Project/OpenROAD/pull/6076

PR #2535, opened 2024-10-31
Update IHP PDK

Syndace:
This syncs the IHP PDK to the most recent main branch commit [035037f](https://github.com/IHP-GmbH/IHP-Open-PDK/commit/035037f6312851414e4bb5daa5a9d18975c768b3).

Next to a few minor changes and new SRAMs, the most interesting updates are:

- Actual, non-dummy liberty for I/O cells
- DRC scripts made compatible with ORFS' `make drc` by respecting `in_gds` and `report_file`


flow/platforms/ihp-sg13g2/lib/sg13g2_io_dummy.lib
maliberty, line 1: If we have real IOs do we still need the dummy ones?

Syndace: The real libs aren't complete yet, a few cells are only available in the dummy still.

maliberty (approved):

PR #2542, opened 2024-11-02
mock-array: works fine with default number of iterations

oharboe:
default is 30, no need to specify more


oharboe:
@maliberty ping?

maliberty (approved):

PR #2547, opened 2024-11-04
Update config files to use DESIGN_HOME variable

VidyaChhabria:
Update config files to use DESIGN_HOME variable needed to setup the RobustDesignFlow

### DIFF
--- a/flow/designs/asap7/aes-block/config.mk
+++ b/flow/designs/asap7/aes-block/config.mk
@@ -3,8 +3,8 @@ export PLATFORM               = asap7
 export DESIGN_NAME            = aes_cipher_top
 export DESIGN_NICKNAME        = aes-block
 
-export VERILOG_FILES = $(sort $(wildcard ./designs/src/aes/*.v))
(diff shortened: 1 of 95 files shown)


maliberty:
What is RobustDesignFlow and why does it require these changes?

VidyaChhabria:
> What is RobustDesignFlow and why does it require these changes?

This is a part of the IEEE CEDA DATC activities (https://github.com/ieee-ceda-datc). As a part of RobustDesignFlow,  we are calling ORFS from an outside folder. This is the only variable that was hard coded. So removing the hard code will be helpful.

maliberty (approved):

oharboe:
What is the use case for DESIGN_HOME?

A variable implies some use case: take on some other value, change in the future, document a convention or save typing.

I dont think this improves legibility(or makes it considerably worse).

ORFS is being used on projects inside ORFS or outside ORFS.

The "designs" folder name is an arbitrary folder name picked in ORFS for testcase projects, it has no special significance outside of ORFS test projects, nor is ORFS suggestive in folder structure for non-ORFS designs.

PR #2551, opened 2024-11-05
add flag to enforce cell footprint honoring in all sizing operations

AcKoucher:
Resolves one of the top level tasks in [#5577](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5577).

Changes:
- New helper for repair_design like the one that already exists for repair_timing including the -match_cell_footprint new flag.
- Include -match_cell_footprint new flag in repair_timing_helper
- Change the "raw" calls for repair_design and repair_timing inside RESYNTH_TIMING_RECOVER (@precisionmoon I think you worked on this part, could you please take a look if that's correct?)
- Add -match_cell_footprint flag in the repair_timing call inside RECOVER_POWER - We should be careful to use the helper throughout elsewhere when/if adding new repair_timing calls, this is an exception.

### DIFF
--- a/flow/scripts/util.tcl
+++ b/flow/scripts/util.tcl
@@ -27,8 +27,21 @@ proc repair_timing_helper { {hold_margin 1} } {
   append_env_var additional_args SKIP_GATE_CLONING -skip_gate_cloning 0
   append_env_var additional_args SKIP_BUFFER_REMOVAL -skip_buffer_removal 0
   append_env_var additional_args SKIP_LAST_GASP -skip_last_gasp 0
+  append_env_var additional_args HONOR_CELL_FOOTPRINT -match_cell_footprint 0
   puts "repair_timing [join $additional_args " "]"
   repair_timing {*}$additional_args
+}
+
+proc repair_design_helper {} {
+  puts "Perform buffer insertion and gate resizing..."
+
+  set additional_args ""
+  append_env_var additional_args CAP_MARGIN -cap_margin 1
+  append_env_var additional_args SLEW_MARGIN -slew_margin 1
+  append_env_var additional_args HONOR_CELL_FOOTPRINT -match_cell_footprint 0
+  puts "repair_design [join $additional_args " "]"
+
+  repair_design {*}$additional_args
 }
 
 proc recover_power {} {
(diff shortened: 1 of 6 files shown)


flow/scripts/variables.yaml
```diff
@@ -622,3 +622,10 @@ GLOBAL_ROUTE_ARGS:
   stages:
     - grt
   default: -congestion_iterations 30 -congestion_report_iter_step 5 -verbose
+HONOR_CELL_FOOTPRINT:
+  description: >
+    Enforce sizing operations to only swap cells that have the same layout boundary.
+  stages:
+    - floorplan
+    - place
+    - cts
```
maliberty, line 631: You have added it to global_route.tcl which is in route not cts

flow/scripts/util.tcl
```diff
@@ -27,10 +27,23 @@ proc repair_timing_helper { {hold_margin 1} } {
...
   repair_timing {*}$additional_args
 }
 
+proc repair_design_helper {} {
+  puts "Perform buffer insertion and gate resizing..."
+
+  set additional_args ""
+  append_env_var additional_args CAP_MARGIN -cap_margin 1
+  append_env_var additional_args SLEW_MARGIN -slew_margin 1
+  append_env_var additional_args HONOR_CELL_FOOTPRINT -match_cell_footprint 0
```
maliberty: Is there a reason to have differing names (honor vs match)?

maliberty (changes requested):
Its is good that you made the various places consistent about CAP_MARGIN/SLEW_MARGIN

flow/scripts/variables.yaml
AcKoucher: Indeed. But we can do repair_timing in cts:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/c01383bed7ea05208642610dde3740d5d4e8394a/flow/scripts/cts.tcl#L67-L72

I'll add route

flow/scripts/util.tcl
AcKoucher: I just followed the proposal literally. I think it's best to keep consistency between flag / orfs variable. I'll fix it.

maliberty (approved):

maliberty:
@precisionmoon do you want footprint matching to apply throughout the flow or just certain steps?

precisionmoon:
> @precisionmoon do you want footprint matching to apply throughout the flow or just certain steps?

We want to match footprint for all steps..   When this is not honored in one step, there is no way to recover from this.

precisionmoon:
@AcKoucher , we want to enforce match cell footprint for the following PDKs:
1) sky130hd
2) sky130hs
3) ihp
flow/platforms/{sky130hd,sky130hs,...}/config.mk needs to be modified also.

PR #2457, opened 2024-10-13
[Docs] CI for variables.yaml + CI for docs update

luarss:
Tracked in #2452 

Fixes https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2444

PR #2555, opened 2024-11-08
Added some missing flow variables and descriptions

jeffng-or:


oharboe (changes requested):
Just to check: This documentation is autogenerated by scripts/generatedocs.py

Did you update variables.py and rerub scripts/generatedocs.py?

jeffng-or:
> Just to check: This documentation is autogenerated by scripts/generatedocs.py
> 
> Did you update variables.py and rerub scripts/generatedocs.py?

Thanks for the pointer. I'll update variables.yaml and the generate-variables-docs.py script.

jeffng-or:
Updated variables.yaml and re-generated FlowVariables.md. It looks like FlowVariables.md wasn't updated in a while, since there are a bunch of new additions in the generated file.

oharboe:
@jeffng-or @maliberty Autogenerating .md files under git source control is not great, but github does not have any dynamic rendering capabilities in .md files.

So, pick a poison: keep generated docs in git, or you dont get .md docs displayed when browsing github and the user needs to consult some other source.

oharboe:
Unrelated to this PR that is just regenerating:

RTLMP_FLOW should be deleted from variables.yaml

USE_FILL default in variables.yaml needs a look

oharboe:
@maliberty @jeffng-or I am of two minds of too much complication in CI for obscure cases, but it would be possible to check that FlowVariables.md, if changed, is not  different after scripts/generatedocs.py has been run: this would cause failure when a PR manually modiies the autogenerated section.

maliberty:
I imagine something like "Check That ODB Files Are Generated" in OR could be used for the docs.  @vvbandeira @luarss any thoughts?

maliberty:
I don't see any change to variables.yaml in this PR

luarss:
@maliberty There is some initial CI (WIP) for `variables.yaml` here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2457

oharboe:
> @maliberty There is some initial CI (WIP) for `variables.yaml` here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2457
> 

Thanks for the reference!

jeffng-or:
> I don't see any change to variables.yaml in this PR

Sorry, committed to the wrong branch. Updated the PR.

maliberty:
> Sorry, committed to the wrong branch. Updated the PR.

I don't see any update - did you push?

maliberty (approved):

oharboe:
@jeffng-or @maliberty Defaults can originate in ORFS or OpenROAD.

If they originate in ORFS, the single - source of truth is variables.yaml

However, if the default values originate in OpenROAD, there is no way to automatically update FlowVariables.md, nor ensure that variables.yaml is up to date.

The choices I can think of are:

1. duplicate information into variables.yaml. Fine as long as default never changes in OpenROAD. This is done today, but not consistently.
2. do not document default in variables.yaml: the user is forced to look up default in OpenROAD. The advantage is that the value is known to be correct(single source of truth).
3. seperate the concern of defaults in OpenROAD and ORFS. Modify ORFS to always specify OpenROAD option with default value from variables.yaml

As for recommendation, I am inclined to leaving as-is and switching to 3 for those cases where it turns out that it matters to anyone.

PR #2566, opened 2024-11-17
detail_route: call fast_route

oharboe:
There's no comment as to why fast_route is not invoked here, this is a copy&paste vestige, call fast_route.


oharboe:
@maliberty My commit comment is a guess, I found this when spelunking, the commit comment needs to be verified in review.

maliberty (approved):
It would be simpler to move it to load_design.  It's also terribly named.

PR #2573, opened 2024-11-19
Yosys dfflibmap now maps flip flops with enable - demo

widlarizer:
This PR demonstrates functionality added in https://github.com/YosysHQ/yosys/pull/4698/. We expect a QoR improvement on sky130hd and sky130hs and no change on other PDKs


maliberty:
I'm confused - you are updating the yosys submodule which points to our fork but somehow getting upstream changes?  I'm confused as to what is happening to make that possible.

widlarizer:
I suspect that this is an effect of fork networks: "Commits to any repository in a network can be accessed from any repository in the same network, including the upstream repository, even after a fork is deleted." - [About permissions and visibility of forks](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/about-permissions-and-visibility-of-forks)

povik:
This can be closed now that Yosys has been updated to 0.48 in #2620 and enable type flip-flop inference happens by default.

PR #2579, opened 2024-11-21
etc: update klayout version

vvbandeira:
Klayout org only provides the latest versions for new OSes such as RHEL9/RockyLinux9/Ubuntu24

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -10,7 +10,7 @@ else
 fi
 
 # package versions
-klayoutVersion=0.28.17
+klayoutVersion=0.29.8
 verilatorVersion=5.026
 
 _versionCompare() {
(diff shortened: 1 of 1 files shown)


etc/DependencyInstaller.sh
```diff
@@ -158,9 +158,14 @@ _installUbuntuPackages() {
...
             if [[ $1 == 20.04 ]]; then
-                klayoutChecksum=15a26f74cf396d8a10b7985ed70ab135
+                klayoutChecksum=bfa011330bc9e1f61b46f668379cc741
+            elif [[ $1 == 22.04 ]]; then
+                klayoutChecksum=60839f6b5b2b1c1e76bf0c1bc17bf18c
+            elif [[ $1 == 24.04 ]]; then
+                klayoutChecksum=b8909538985b15135b49e8d32a089d72
             else
-                klayoutChecksum=db751264399706a23d20455bb7624264
+                echo "Could not find checksum for klayout"
```
rovinski: A more actionable message would be better for the user
```suggestion
                echo "KLayout does not provide binaries for detected Ubuntu version $1. Please install KLayout manually."
```

maliberty: I could imagine a case where 26.04 comes out and we haven't updated yet.  klayout might provide binaries and our script is just behind.  Perhaps "Unrecognized version of Ubuntu $1.  Please install KLayout manually".

@vvbandeira this has been lingering a while.

```suggestion
                echo "Unrecognized version of Ubuntu $1. Please install KLayout manually."
```

vvbandeira: The CI failed with the update, and I did not have the time to take a closer look at the error. I will try to get back later this week.

maliberty (approved):

PR #2581, opened 2024-11-21
added RTLMP* variables and cleaned up per yamllint

jeffng-or:
1. Added docs for some RTLMP* variables
2. Cleaned up content per yamllint
3. Fixed variables in macro_place_util.tcl

### DIFF
--- a/flow/scripts/variables.yaml
+++ b/flow/scripts/variables.yaml
@@ -1,3 +1,4 @@
+---
 GENERATE_ARTIFACTS_ON_FAILURE:
   description: >
     For instance Bazel needs artifacts (.odb and .rpt files) on a failure to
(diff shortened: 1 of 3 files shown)


flow/scripts/variables.yaml
```diff
@@ -659,10 +685,136 @@ GLOBAL_ROUTE_ARGS:
...
+  description: >
+    Maximum depth of the physical hierarchy tree.
+  default: 2
+  stages:
+    - floorplan
+RTLMP_MAX_INST:
+  description: >
+    Maximum number of standard cells in a cluster. If unset, rtl_macro_placer
+    will calculate a value based on the design attributes.
+  default: 0
```
maliberty: Do we need to give a default since unset is handled by OR and 0 isn't informative?  Likewise min/max inst/macro.

jeffng-or: done!

flow/scripts/variables.yaml
```diff
@@ -108,7 +114,7 @@ PROCESS:
     Technology node or process in use.
 CORNER:
   description: >
-    Library to select based on corner BC/TC/WC.
+    PVT corner library selection. Only available for ASAP7, GF55 and GF180 PDKs.
```
maliberty: gf55 isn't a public platform so please remove it.  That should fix the scan error

jeffng-or: removed and pushed up a squashed commit

maliberty (approved):

PR #2587, opened 2024-11-23
global route: write out .v file and .spef file

oharboe:
Can be used to simulate power based on global route, instead of having to sit through detailed routing.

https://github.com/The-OpenROAD-Project/OpenROAD/discussions/6219#discussioncomment-11359128


maliberty:
These are large files, I would rather their generation was conditional as they have no use in the general flow.

oharboe:
Out of sight and out of mind until https://github.com/The-OpenROAD-Project/OpenROAD/issues/6225 is fixed.

PR #2588, opened 2024-11-25
mock-array: fix make verilog

oharboe:
upgrade to later version of scala that works


maliberty (approved):

oharboe:
@maliberty unrelated pr-head/merge failures

PR #2589, opened 2024-11-25
[DO NOT REVIEW] Mock array with input output registers

oharboe:


oharboe:
CTS created a surprising amount of skew here:

image

maliberty:
That's pretty ugly - where were the registers placed?  @arthurjolo is working a similar looking issue in another design.

oharboe:
> That's pretty ugly - where were the registers placed? @arthurjolo is working a similar looking issue in another design.

Perhaps this is a downstream effect of [the macro placement problem](https://github.com/The-OpenROAD-Project/OpenROAD/issues/6242)?

There are flip flops on all inputs/outputs around of what should be an 8x8 array centered element array:

image

maliberty:
I would wait for the macro to be resolved and see

oharboe:
> I would wait for the macro to be resolved and see

Yes. No rush: there are several layers to this onion PR before it can be merged...

oharboe:
The mock-array has accumulated an impressive number of layers of use-cases by now... That's both good and bad. Ideally there would be single concern unit-tests for each use-case.

oharboe:
@maliberty This PR is extending testcases, close it so it is out of sight out and out of mind until fixes in OpenROAD land:

- https://github.com/The-OpenROAD-Project/OpenROAD/issues/6181
- https://github.com/The-OpenROAD-Project/OpenROAD/issues/6242

oharboe:
@maliberty @eder-matheus The IO placement is funny with this change(once macro placement is fixed) :-)

The pins should be just across from the rows/columns, but the pins are bundled into the upper right and lower left corner.

This of course messes up placement and routing. The CTS looks more reasonable with a good macro placmenet though.

image

eder-matheus:
@oharboe Could you send a tar file with the artifacts for this run?

oharboe:
> @oharboe Could you send a tar file with the artifacts for this run?

Never mind :-) Mea culpa, I swapped x/y in manual placement

oharboe:
To make it build, rebase on master, disable automatic macro placement and disable hold repair(false violations in floorplan will stop the build).

Will [fail in drt](https://github.com/The-OpenROAD-Project/OpenROAD/issues/6267)

```
make DESIGN_CONFIG=designs/asap7/mock-array/config.mk RTLMP_FLOW=0 HOLD_SLACK_MARGIN=-200
```

image



image

PR #2590, opened 2024-11-26
detail route: remove broken check that is handled inside detailed route

oharboe:
Detailed route will fail once it loads the guides and see them congested.

Remove this fail-early, but broken check.

Perhaps detailed route will fail earlier in the future, but this check isn't doing anything now and is broken as is, so remove it.


oharboe:
Or is this moot with https://github.com/The-OpenROAD-Project/OpenROAD/pull/6237/ ?

eder-matheus:
> Or is this moot with [The-OpenROAD-Project/OpenROAD#6237](https://github.com/The-OpenROAD-Project/OpenROAD/pull/6237) ?

Yes, the PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/6237 should fix the broken check.

PR #2591, opened 2024-11-26
Add `REPAIR_TIMING_MAX_PASSES_PER_ITER` parameters

eszpotanski:
This PR adds new variables for max number of repairs per iteration in `rsz::repairSetup`, introduced in https://github.com/The-OpenROAD-Project/OpenROAD/pull/6244. It also sets the best found values for a few designs.

It adds:
* `FP_REPAIR_TIMING_MAX_PASSES_PER_ITER` - max number of repairs per iteration during floorplan,
* `CTS_REPAIR_TIMING_MAX_PASSES_PER_ITER` - max number of repairs per iteration during CTS,
* `GRT_REPAIR_TIMING_MAX_PASSES_PER_ITER` - max number of repairs per iteration during GRT.


flow/scripts/cts.tcl
```diff
@@ -69,7 +69,7 @@ if {![env_var_equals SKIP_CTS_REPAIR_TIMING 1]} {
       write_eqy_verilog 4_before_rsz.v
   }
 
-  repair_timing_helper
+  repair_timing_helper "cts"
```
oharboe, line 72: remove quotes, not needed in .tcl

flow/scripts/variables.yaml
```diff
@@ -287,6 +287,12 @@ IO_PLACER_V:
   stages:
     - floorplan
     - place
+FP_REPAIR_TIMING_MAX_PASSES_PER_ITER:
+  description: >
+    Maximum number of repairs per iteration during repair_timing, called from floorplan.
```
oharboe, line 317: more magic values? Every option makes a feature twice as hard to use.

flow/scripts/variables.yaml
```diff
@@ -652,6 +658,12 @@ CTS_CLUSTER_SIZE:
   default: 50
   stages:
     - cts
+CTS_REPAIR_TIMING_MAX_PASSES_PER_ITER:
+  description: >
+    Maximum number of repairs per iteration during repair_timing, called from CTS.
+  default: 20
+  stages:
+    - cts
```
oharboe, line 722: A proliferation of the same variable for the same operation. This multiplies the number of variables... Not ideal...

oharboe (changes requested):
each magic variable makes the feature twice as hard to use...

Can we make some better automatic policy here?

PR #2592, opened 2024-11-26
update nangate45/ariane136 place density and metrics

gudeh:
Nightly is failing on this design after gpl integration with rsz repair_design. For this specific situation, it never converges, the overflow diverges up and down. With the increased density it converges as expected.

I understand the issue is happening because rsz does a substantial increase in area and gpl is currently unable to find a solution with a lower density. After we increase the area, there is less empty space and we are unable to reach the target density.

./designs/nangate45/ariane136/ --variant base --failing --tighten
| finish__timing__drv__hold_violation_count     |      835 |      259 | Tighten  |


maliberty:
Sounds like the same as https://github.com/The-OpenROAD-Project/OpenROAD/issues/6238.  We can merge this to get master passing but should revisit after fixing that issue.

maliberty (approved):

gudeh:
The two issues indeed are really similar but with a difference. GPL has fillercells which relate to the empty space available. On Peter's issue the filler cell area ends up being negative, meaning it is impossible to achieve the target density.

This one from Ariane the issue is more hidden, because there should be empty space available, but the algorithm itself could not find a solution.

Either way, I must find a solution for this.

maliberty:
@gadfort why are there filler cells at global placement?

gadfort:
> @gadfort why are there filler cells at global placement?

@maliberty I think @gudeh is referring to the "filler cells" that GPL uses, not actual filler cells. (answered in the wrong place)

gudeh:
yes, GPL has some unusual namings. Its filler cells are just "dummy" cells to occupy any empty space. I understand they influence the movement of real cells.

PR #2600, opened 2024-11-30
makefile: Ubuntu 24.04 wayland gui_ workaround

oharboe:


oharboe:
doesn't always work

oharboe:
@maliberty Doesn't always work...

oharboe:
Seems to work on small designs, but not on larger designs that run for longer.

maliberty:
It appears wayland limits the action to be in response to user activity and presumably it takes too long.

oharboe:
> It appears wayland limits the action to be in response to user activity and presumably it takes too long.

There are some good intentions in Wayland that have gone wrong for sure. I dont know exactly what the policy is, I can find counterexamples to all policies I have thought of, including the one you are guessing above: a very short pause should then work, it doesnt. We can make the GUI appear after a very long time(my PR).

maliberty:
My suggestion is to turn off wayland.  It seems they have made this impossible to do in a usable way.

oharboe:
> My suggestion is to turn off wayland. It seems they have made this impossible to do in a usable way.

So restore https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2600 ?

Not perfect, but I think we can close the issues as inactionable until someone comes forth with more actionable information or a PR.

oharboe:
Doesn't work on my laptop, even for `make DESIGN_CONFIG=designs/nangate45/gcd/config.mk gui_floorplan`

maliberty:
@gadfort told me this worked for him but then stopped working later for unknown reasons.

maliberty:
My suggestion is to turn off wayland completely, not just for OR.  From AI

```
To disable Wayland security on a Linux device, you can: 
Open the Wayland configuration file for your distribution:
sudo nano /etc/gdm3/custom.conf
sudo nano /etc/gdm/custom.conf
Edit the file with root privileges
Find the line #WaylandEnable=false and remove the # to change it to WaylandEnable=false
Save your changes
Reboot your computer
```

oharboe:
> My suggestion is to turn off wayland completely, not just for OR. From AI
> 
> ```
> To disable Wayland security on a Linux device, you can: 
> Open the Wayland configuration file for your distribution:
> sudo nano /etc/gdm3/custom.conf
> sudo nano /etc/gdm/custom.conf
> Edit the file with root privileges
> Find the line #WaylandEnable=false and remove the # to change it to WaylandEnable=false
> Save your changes
> Reboot your computer
> ```

Your thinking is that enough people is suffering from the same problem in other applications and that an updated Ubuntu/Wayland/Qt will fix this?

maliberty:
I'm thinking it is enough suffering on this issue already for us.  I see no way to make wayland work as it seems to have prevented this by design.  It is a security policy that the user can't opt out of apparently.  Other apps are just a side effect.

oharboe:
I see.

So close the issue: not actionable without significant new easily actionable information or ideally a PR.

oharboe:
Logging in on an X11 session does not fix the problem:

image

```
$ export | grep x11 
declare -x MEMORY_PRESSURE_WATCH="/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/session.slice/org.gnome.Shell@x11.service/memory.pressure"
declare -x XDG_SESSION_TYPE="x11"
```

oharboe:
> My suggestion is to turn off wayland completely, not just for OR. From AI
> 
> ```
> To disable Wayland security on a Linux device, you can: 
> Open the Wayland configuration file for your distribution:
> sudo nano /etc/gdm3/custom.conf
> sudo nano /etc/gdm/custom.conf
> Edit the file with root privileges
> Find the line #WaylandEnable=false and remove the # to change it to WaylandEnable=false
> Save your changes
> Reboot your computer
> ```

Doesn't fix the problem either.

The wayland theory is bereft of life, pushing up daisies, no more, seized to be, it is an ex-theory.

I'm not saying this can't be a problem on Wayland or worse on Wayland, but it does happen without Wayland.

maliberty:
It appears gnome also has focus stealing prevention.  I normally use xfce4 which doesn't appear to have this property.  Can you give that a try?

oharboe:
> It appears gnome also has focus stealing prevention. I normally use xfce4 which doesn't appear to have this property. Can you give that a try?

Yes, that works. I installed it and then clicked on my login + the gears and selected xfce4.

Good to know, but obviousl it is a case of solving one problem and introducing several others...

image

maliberty:
The simplest solution is the oldest, just show the window from the start and wait for things to finish loading. Whatever the convenience of hiding the window, it cuts against the grain of how window systems are evolving for security reasons.

oharboe:
Agreed.

maliberty:
> Agreed.

I will make that change to ORFS then

PR #2610, opened 2024-12-03
make: gui_ prerendering

oharboe:
less [click and wait](https://github.com/The-OpenROAD-Project/OpenROAD/issues/6074) when the GUI opens

Clock tree prerendering is still missing, because there is  no way yet to disable the clock tree rendering from .tcl after   the clock tree has been rendered.
    
Some progress added so that the user can tell what it is  that is taking time. Also added a final gui::unminimize  log item so the user knows when the GUI should have unminimized(which it doesn't on Ubuntu 24.04 with Wayland yet, known issue filed).

```    
make gui_route
[deleted]
estimate_parasitics -global_routing
Populating timing paths...OK
Prerendering Placement heatmap...
Prerendering Routing heatmap...
Prerendering RUDY heatmap...
Prerendering Power heatmap...
gui::select_chart "Endpoint Slack"
gui::update_timing_report
gui::unminimize
```


flow/scripts/open.tcl
```diff
@@ -60,12 +60,43 @@ proc read_timing {input_file} {
...
+  foreach clock [get_clocks *] {
+    if { [llength [get_property $clock sources]] > 0 } {
+      set clock_name [get_name $clock]
+      save_clocktree_image -clock $clock_name \
+          -width 1024 -height 1024 \
+          $::env(REPORTS_DIR)/cts_$clock_name.webp
+      gui::select_clockviewer_clock $clock_name
+      save_image -resolution $resolution $::env(REPORTS_DIR)/cts_${clock_name}_layout.webp
+    }
+  }
```
oharboe: @maliberty How do I unselect the clock tree and return the GUI to the default view?

`make gui_cts`

image

maliberty:
@gadfort "How do I unselect the clock tree and return the GUI to the default view?"

flow/scripts/open.tcl
```diff
@@ -80,12 +80,19 @@ if {[env_var_equals GUI_TIMING 1]} {
...
-        gui::set_heatmap $heatmap rebuild 1
-        gui::dump_heatmap $heatmap $::env(REPORTS_DIR)/dummy.png
-      } error
+    if {[string equal $heatmap Routing] && ![grt::have_routes]} {
+      puts "Skipping $heatmap heatmap, no routes available"
+      continue
     }
+    if {[string equal $heatmap IRDrop] && ![env_var_exists_and_non_empty PWR_NETS_VOLTAGES]} {
+      puts "Skipping $heatmap heatmap, no PWR_NETS_VOLTAGES available"
+      continue
```
oharboe: @maliberty How do I skip the IRDrop heatmap when there is no data?

```
 make gui_floorplan
ODB_FILE=./results/nangate45/gcd/base/2_floorplan.odb GUI_SHOW=1 /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad -gui  -minimize /home/oyvind/OpenROAD-flow-scripts/flow/scripts/open.tcl
QSocketNotifier: Can only be used with threads started with QThread
OpenROAD v2.0-17013-gf7f634f88 
Populating timing paths...OK
Prerendering...
Dumping Placement heatmap...
Dumping IRDrop heatmap...
[ERROR GUI-0072] "IR Drop" is not populated with data.
[ERROR GRT-0010] Instance ctrl.state.out\[0\]$_DFF_P_ is not placed.
[ERROR GUI-0070] Error: open.tcl, 101 GRT-0010
^Cmake: *** [Makefile:1013: gui_2_floorplan.odb] Interrupt
```

oharboe: How do I skip Placement heatmap?

[ERROR GRT-0010] Instance ctrl.state.out\[0\]$_DFF_P_ is not placed.
[ERROR GUI-0070] Error: open.tcl, 105 GRT-0010

maliberty: We don't save the ir drop information in odb so if you are reloading there isn't any.

oharboe: I don't know what that means... I haven't used IR Drop yet. I'm going to remove IRDrop for now if there isn't an easy way to skip it when prerendering should be skipped.

Placement heatmap is more interesting: how do I skip it when there's no placement?

maliberty: You could loop over the instances and see if any are placed.

oharboe:
@maliberty Should grt::have_routes return false silently if the instances are not placed?

```
>>> grt::have_routes
[ERROR GRT-0010] Instance ctrl.state.out\[0\]$_DFF_P_ is not placed.
[ERROR GUI-0070] GRT-0010
```

oharboe:
@maliberty FYI https://github.com/The-OpenROAD-Project/OpenROAD/issues/6293

flow/scripts/open.tcl
oharboe: Worked. I haven't tested how well it scales though.

oharboe:
@maliberty Tried on a larger design, works nicely. Nits need to be fixed, broken out as issues.

oharboe:
> @gadfort "How do I unselect the clock tree and return the GUI to the default view?"

How about https://github.com/The-OpenROAD-Project/OpenROAD/pull/6307 ?

oharboe:
@maliberty Ready for review. Preprendering clock tree is still missing, but that's a simple change once this is through and that missing clock tree feature is available in OpenROAD.

flow/scripts/open.tcl
```diff
@@ -60,12 +60,74 @@ proc read_timing {input_file} {
...
+  #
+  # foreach clock [get_clocks *] {
+  #   if { [llength [get_property $clock sources]] > 0 } {
+  #     set clock_name [get_name $clock]
+  #     save_clocktree_image -clock $clock_name \
+  #         -width 100 -height 100 \
+  #         $::env(OBJECTS_DIR)/dummy.png
+  #     break
+  #   }
+  # }
```
maliberty: open should not write files as a side effect.  This should be removed

oharboe: Agreed, but how else to precalculate this?

maliberty: Does it actually save time?  Most of the effort is to get sta warmed up.  Drawing the tree should be cheap.

oharboe: Yes, last I checked.

However, I can delete the commented code and if this PR is otherwise good to go, we can square away what we already have a solution for and I can then create a new PR with a narrower concern and some numbers.

A way to precalculate this that does not modify GUI state and also does not write files to disk is needed, which isnt available yet, right?

maliberty: Yes, if there really is a need.  I would be curious to see the time saved.

oharboe:
@maliberty Fixed, should be ready for merge now. This is how far we've come with prerendering for now. Will revisit more later in followup PRs.

flow/scripts/open.tcl
```diff
@@ -60,12 +60,62 @@ proc read_timing {input_file} {
...
+  foreach inst $insts {
+    set status [$inst getPlacementStatus]
+    # status is not in the list of PLACED, LOCKED
+    if {[lsearch {PLACED LOCKED} $status] == -1} {
+      set placed 0
+      break
+    }
+  }
+
+  set have_routes [expr {$placed && [grt::have_routes]}]
```
maliberty: Do we still need to compute/check $placed?  I thought this was fixed.

flow/scripts/open.tcl
```diff
@@ -60,12 +60,62 @@ proc read_timing {input_file} {
...
+      # Skipping $heatmap heatmap, no PWR_NETS_VOLTAGES available
+      continue
+    }
+    if {[lsearch {RUDY Placement} $heatmap] != -1 && !$placed} {
+      # Skipping $heatmap heatmap, not all instances are placed
+      continue
+    }
+    puts "Prerendering $heatmap heatmap..."
+    gui::set_heatmap $heatmap rebuild 1
+    gui::dump_heatmap $heatmap $::env(REPORTS_DIR)/dummy.png
```
maliberty: Can we avoid the file side effect?

flow/scripts/open.tcl
oharboe: Yes: there is no ppl::have_placed method. I can simplify the code a bit here though for the have_routes case.

flow/scripts/open.tcl
oharboe: yes: fixed

oharboe:
@maliberty Fixed, should be good to go.

oharboe:
good to go?

maliberty:
I think all this warming should have its own control and be off by default.  I don't want to spend time on all this if I know what I'm going to be looking at is a subset.

oharboe:
> I think all this warming should have its own control and be off by default. I don't want to spend time on all this if I know what I'm going to be looking at is a subset.

Would you be happy with a policy to unconditionally compute it at build time?

oharboe:
Meaning: would you agree that it is better to close this PR and then open a more longer term feature request of storing precomputed values unconditionally in the .odb file?

oharboe:
Every option halves the usefulness of a feature.

maliberty:
This is an unconditional penalty on the user who isn't interested in "global warming" 😉 

I would be a very long term / large enhancement to store all the sta data in odb.  I'm ok if you want to open it but expect the response to be very slow.

oharboe:
@maliberty broke out https://github.com/The-OpenROAD-Project/OpenROAD/issues/6457 and only the timing reports are now updated, but only if GUI_TIMING=1

maliberty (approved):

PR #2619, opened 2024-12-10
Simplify handling of Liberty file related variables

widlarizer:
- remove preprocessLib.py
  - yosys-abc accepts user-defined attribute `original_pin` used in Liberty file pin declarations of cells it actually uses to map to even if `original_pin` is not defined as user-defined. Probably from Synopsys PrimeTime folklore? Can't find much of a trace. Let me know if private PDKs break with this
  - yosys accepts the Liberty expression pattern described in the comment, see [here](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/373#discussion_r1872687473)
  - extract gz in Makefile instead (could be done more elegantly but I don't like fiddling with Make)
- rename DONT_USE_LIBS->LIB_FILES_EXTRACTED
  - it's the files listing cells that **should** be used, but it was named the same as the list of cells that **shouldn't** be used. This is because before #1813 there was a script that would add `dont_use` attributes to some of the cells in those files
  - the only difference between LIB_FILES_EXTRACTED and LIB_FILES is that they're extracted if gzip and copied to `objects/`
- rename DONT_USE_SC_LIB->SC_LIB_FILE

I tried this out with asap7 and nangate and it seems to work fine

cc @povik


widlarizer:
This can be followed up by integrating https://github.com/YosysHQ/yosys/pull/4797 and redoing #2139 if possible

flow/scripts/synth_stats.tcl
```diff
@@ -23,7 +23,7 @@ proc write_keep_hierarchy {} {
   if {[env_var_exists_and_non_empty DFF_LIB_FILE]} {
     dfflibmap -liberty $::env(DFF_LIB_FILE)
   } else {
-    dfflibmap -liberty $::env(DONT_USE_SC_LIB)
+    dfflibmap -liberty $::env(SC_LIB_FILE)
```
rovinski, line 26: The only reason for this if/else is because `dfflibmap` used to only take one liberty file. If it accepts multiple now, then you can simply pass in `LIB_FILES_EXTRACTED` and eliminate the `DFF_LIB_FILE` variable entirely. In fact, `DFF_LIB_FILE` was added exclusively to support ASAP7, no other PDK uses it.

flow/scripts/synth.tcl
```diff
@@ -67,7 +67,7 @@ foreach cell $::env(DONT_USE_CELLS) {
 if {[env_var_exists_and_non_empty DFF_LIB_FILE]} {
   dfflibmap -liberty $::env(DFF_LIB_FILE) {*}$dfflibmap_args
 } else {
-  dfflibmap -liberty $::env(DONT_USE_SC_LIB) {*}$dfflibmap_args
+  dfflibmap -liberty $::env(SC_LIB_FILE) {*}$dfflibmap_args
```
rovinski, line 70: Same

flow/Makefile
```diff
@@ -355,12 +354,16 @@ versions.txt:
...
 .SECONDEXPANSION:
-$(DONT_USE_LIBS): $$(filter %$$(@F) %$$(@F).gz,$(LIB_FILES))
+$(LIB_FILES_EXTRACTED): $$(filter %$$(@F) %$$(@F).gz,$(LIB_FILES))
 	@mkdir -p $(OBJECTS_DIR)/lib
-	$(UTILS_DIR)/preprocessLib.py -i $^ -o $@
+	@if echo "$<" | grep -q "\.gz$$"; then \
+		gunzip -c $< > $@; \
+	else \
+		cp $< $@; \
+	fi
```
rovinski, line 363: Does Yosys support reading in compressed .gz files and decompressing on-the-fly? This step could be entirely eliminated if so.

rovinski (commented):
Thanks for the work on this!

flow/Makefile
widlarizer: It does not, but it could. It requires zlib by default to support .fst wave files. I'll see what I can do

povik: I think it does when you supply the file to a frontend (e.g. `read_liberty`) but not e.g. in `dfflibmap -liberty`

maliberty: It would be nice to support gz in both

povik: We're working on it

PR #2629, opened 2024-12-18
Wrapped operators synthesis flow

povik:
This is a synthesis flow which

 * preserves arithmetic operator boundaries (boundaries of what Yosys considers an `$alu` and `$macc` cell)
 * generates alternative architectures for those preserved operators, and exports those as uninstantiated modules in the final netlist

It's gated on the `NEW_OPERATOR_SYNTHESIS` option which I left undocumented. This option will only work once Yosys is bumped to 0.48 (~pending in #2620~; done)

### DIFF
--- a/flow/scripts/synth.tcl
+++ b/flow/scripts/synth.tcl
@@ -35,7 +35,7 @@ json -o $::env(RESULTS_DIR)/mem.json
 # Run report and check here so as to fail early if this synthesis run is doomed
 exec -- python3 $::env(SCRIPTS_DIR)/mem_dump.py --max-bits $::env(SYNTH_MEMORY_MAX_BITS) $::env(RESULTS_DIR)/mem.json
 
-if {![env_var_exists_and_non_empty NEW_OPERATOR_SYNTHESIS]} {
+if {![env_var_exists_and_non_empty SYNTH_WRAPPED_OPERATORS]} {
   synth -top $::env(DESIGN_NAME) -run fine: {*}$::env(SYNTH_FULL_ARGS)
 } else {
   source $::env(SCRIPTS_DIR)/synth_wrap_operators.tcl
@@ -82,7 +82,7 @@ if {[env_var_exists_and_non_empty DFF_LIB_FILE]} {
 }
 opt
 
-if {![env_var_exists_and_non_empty NEW_OPERATOR_SYNTHESIS]} {
+if {![env_var_exists_and_non_empty SYNTH_WRAPPED_OPERATORS]} {
   log_cmd abc {*}$abc_args
 } else {
   scratchpad -set abc9.script scripts/abc_speed_gia_only.script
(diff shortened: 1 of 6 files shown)


flow/scripts/synth.tcl
```diff
@@ -31,7 +31,11 @@ if { [env_var_equals SYNTH_GUT 1] } {
   delete $::env(DESIGN_NAME)/c:*
 }
 
-synthesize_check mem $::env(SYNTH_FULL_ARGS)
+if {![env_var_exists_and_non_empty NEW_OPERATOR_SYNTHESIS]} {
+  synthesize_check $::env(SYNTH_FULL_ARGS)
+} else {
+  source $::env(SCRIPTS_DIR)/synth_wrap_operators.tcl
```
oharboe, line 41: why not move into synthesize_check?

oharboe (commented):
can we avoid copy and paste by having a single synthesize_check proc that covers this new use-case too?

povik:
> can we avoid copy and paste by having a single synthesize_check proc that covers this new use-case too?

I guess we can. I was in a hurry to put together a working prototype. After finishing #2552 I will return to this and see about cleaning it up.

povik:
I've rebased to pull in the Yosys bump to v0.48

maliberty:
> I've rebased to pull in the Yosys bump to v0.48

are you looking for some action - this is still marked do not merge

povik:
Originally I just wanted a working version of this branch to send to Cho for his experiments, but there's no reason to hold this back now

flow/scripts/variables.yaml
```diff
@@ -228,6 +228,13 @@ MAX_UNGROUP_SIZE:
     platform's standard cell library. The default value is platform specific.
   stages:
     - synth
+NEW_OPERATOR_SYNTHESIS:
```
povik: I've documented the option though I'd consider it experimental

maliberty (approved):

maliberty:
@precisionmoon FYI

povik:
There's a CI failure due to an unmapped latch in the ihp-sg13g2/ibex netlist.
```
(* src = "/tmp/workspace/flow-scripts-Public_PR-2629-head/flow/designs/src/ibex/prim_clock_gating.v:16.3-20.6" *)
  always @*
    if (!clk_i) \core_clock_gate_i.en_latch  = _00005_;
```
This might be a pre-existing issue which became visible now that we've removed `-noexpr` from the write_verilog command in the Yosys script. I'll investigate.

povik:
> There's a CI failure due to an unmapped latch in the ihp-sg13g2/ibex netlist.

This has been fixed elsewhere. I've rebased to get this fix, and also I've changed:

 * renamed `NEW_OPERATOR_SYNTHESIS` to `SYNTH_WRAPPED_OPERATORS`
 * made corrections inside synth_wrap_operators.tcl
 * added Booth multiplier option (only available if the RTL operation isn't fused into a sum-of-products)

povik:
@maliberty CI is passing, please take a look

PR #2637, opened 2024-12-22
docs: add some ROUTING_LAYER_ADJUSTMENT color

oharboe:


flow/scripts/variables.yaml
```diff
@@ -42,7 +42,18 @@ TNS_END_PERCENT:
...
+    improve detailed routing. High values ease detailed routing
+    but risk excessive detours and long global routing times,
+    while low values reduce global routing failure but can
+    complicate detailed routing.
+
+    The global routing running time normally reduces
+    dramatically(from hours to minutes) when the value is
+    very low. If the default value does not work, start with
+    a too low low value, such as 0.10, and bisect your way
+    to value that works by doing multiple global routing runs.
```
maliberty: The runtime is design dependent.  I don't wish to make any promise that it can be achieved in minutes.  'very low' is also ill-defined.  Please strike the first sentence.

What do you mean by 'does not work'.  Do you mean congestion errors or slow runtime?  If you have congestion errors then lowering the density further won't be helpful.

'low low' -> low.

maliberty (approved):

AcKoucher:
@oharboe I'm working on updating some floorplan variables' descriptions and when I ran generate-variables-docs.py in order to regenerate the table with my updates in variables.yaml, the description added in this PR also appeared in the changes.

Was it intentional to not update the flow variables table with the description here?

oharboe:
We don't always re-run generate-variable-docs.py when we merge changes to variables.yaml.

Some CI work has been done to require a rerun of generate-variable-docs.py upon variables.yaml changes, but what we have now with some ad hoc re-running of generate-variables-docs.py seems to do the job.

PR #2651, opened 2025-01-03
[DO NOT REVIEW YET] mock-array: run power.tcl from openroad and sta and diff

oharboe:


oharboe:
circle back to this later

PR #2652, opened 2025-01-03
make: fix my open_ gaffe

oharboe:
interact with GUI only when gui is enabled


oharboe:
@jeffng-or FYI, this fix is needed to update bazel-orfs

maliberty (approved):

PR #2653, opened 2025-01-03
makefile: make RUN_LOG_NAME_STEM=foo run to specify name of log

oharboe:
Short explanation: simpler scripting, better impedance match with makefiles and bazel.


flow/Makefile
```diff
@@ -1051,10 +1051,12 @@ test-unset-and-make-%: ; $(UNSET_AND_MAKE) $*
 klayout:
 	$(KLAYOUT_CMD)
 
+export RUN_LOG_NAME_STEM ?= run
```
maliberty: Should be documented.

maliberty (approved):

PR #2659, opened 2025-01-07
ci: enable AutoTuner CI

vvbandeira:
This will only affect PRs that have the label "autotuner" or the branch name has the prefix "at-".

This brings back the code removed at commit 1ce17607b1026f75130a19512ceacbd298139a86.

### DIFF
--- a/flow/test/test_helper.sh
+++ b/flow/test/test_helper.sh
@@ -77,7 +77,25 @@ if [ $ret -eq 0 ] && grep -q 'power:' <(echo $TARGETS); then
   ret=$(( ret + $? ))
 fi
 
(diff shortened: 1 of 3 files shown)


vvbandeira:
@jeffng-or @maliberty @luarss FYI

vvbandeira:
@luarss There is still instability with the requirements.

luarss:
May I suggest some changes from https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2538?

Basically the instability hash error happens when network is unstable, and the size of packages we have to install is large. So I set up a retry mechanism during pip install

PR #2663, opened 2025-01-08
Mock array fixes

oharboe:


flow/designs/src/mock-array/simulate.cpp
```diff
@@ -80,7 +80,7 @@ int main(int argc, char** argv) {
 
             for (int k = 0; k < 2; k++) {
                 top->eval();
-                vcd->dump(tick++);
+                vcd->dump(tick++ * 125);
```
maliberty, line 83: what does this do?

maliberty (approved):

flow/designs/src/mock-array/simulate.cpp
oharboe: half a clock period with 1ps resolution, clock period is 250ps.

Issue #2660, opened 2025-01-07
Running ORFS flow using docker_shell results in yosys-abc failure


jeffng-or:
### Subject

[Flow] for any util, flow Makefile, or flow script issues.

### Describe the bug

Found during Quickstart guide development.

If you run util/docker_shell make DESIGN_CONFIG=designs/asap7/gcd/config.mk, you'll encounter error 134 when running yosys-abc. The issue is caused because the merged.lib contains multiple entries for the same cell.

The root cause is that the MAKEFILES env var definition in docker_shell is causing the Makefile tree to be read twice, resulting in each Liberty file being defined twice in LIB_FILES. Removing the MAKEFILES env var solves the problem.

Also, we should add ". ../env.sh" to the docker_shell invocation to ensure that the ORFS environment is set properly.

### Expected Behavior

./util/docker_shell make DESIGN_CONFIG=designs/asap7/gcd/config.mk should complete successfully.

### Environment

```shell
v3.0-1990-g1cbce355
```

### To Reproduce

./util/docker_shell make DESIGN_CONFIG=designs/asap7/gcd/config.mk

PR #2665, opened 2025-01-08
docker shell mods to address reading makefiles in twice and enable op…

jeffng-or:
…enroad to be called without make

Removes MAKEFILES env var, which caused the flow/Makefile to be read in twice resulting in a yosys-abc failure when the merged.lib had duplicate cell entries.

Adds sourcing of the env.sh, so you can do a "util/docker_shell openroad -gui".


jeffng-or:
addresses https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2660

maliberty (approved):

PR #2670, opened 2025-01-10
makefile: make issue, retire buggy confusing .gdb and .tcl files

oharboe:
These are broken when FLOW_HOME is not defined(.tcl version is broken even with FLOW_HOME defined), which is indicative to me that these files are not being used.

They should either be fixed.

.gdb file can't be fixed because the .gdb format lacks the required sophistication.

.tcl file could be fixed, but why have more than a canonical robust well trodden path to reproduce issues?

Example of .tcl file not working:

```
$ . ~/OpenROAD-flow-scripts/env.sh 
OPENROAD: /home/oyvind/OpenROAD-flow-scripts/tools/OpenROAD
$ echo $FLOW_HOME
/home/oyvind/OpenROAD-flow-scripts/flow
$ openroad
OpenROAD v2.0-18021-g74c7c59d2 
Features included (+) or not (-): +GPU +GUI +Python
openroad> source vars-fs-mul-mod-q-asap7-base.tcl 
Error: vars-fs-mul-mod-q-asap7-base.tcl, 8 can't read "FLOW_HOME": no such variable
while evaluating source vars-fs-mul-mod-q-asap7-base.tcl 
openroad> 
```

### DIFF
--- a/flow/util/makeIssue.sh
+++ b/flow/util/makeIssue.sh
@@ -63,9 +63,7 @@ done
 ISSUE_CP_FILES+="${ISSUE_CP_FILES_PLATFORM} \
     $UTILS_DIR/def2stream.py \
     ${RUN_ME_SCRIPT} \
-    $VARS_BASENAME.sh \
-    $VARS_BASENAME.tcl \
-    $VARS_BASENAME.gdb"
+    $VARS_BASENAME.sh"
 
 cat > ${RUN_ME_SCRIPT} <<EOF
 #!/usr/bin/env bash


maliberty:
I usually just do a quick replace on the .gdb if needed.  What do you suggests replaces it?

oharboe:
> I usually just do a quick replace on the .gdb if needed. What do you suggests replaces it?

I don't have any great suggestions here, I think there are only unsatisfying choices to be made. Doing nothing here is also a choice. Increasing complexity or spending more time on this isn't great are also choices and not great.

Deleting things that is confusing and harming users is what I'm proposing here, at least it is simple. We have some evidence that the .tcl file is broken and is confusing users. It is rarely used and there's no regression test, so delete it.

What I do and what I can recommend to users, it works for most cases:

1. `source vars-*.sh`. This sets up the bash shell with all the variables for all apps. However, it also poisons the bash shell, if you invoke "make" on ORFS, you get into trouble.
2. Now launch the debugger
3. debug
4. quit debugger
5. close shell that was poisoned with `source vars-*.sh` above

oharboe:
Add CI tests for .tcl file and fix instead.

PR #2677, opened 2025-01-13
asap7: dff.v was missing internal signals

oharboe:
This caused .vcd files in mock-array not to have pin values for all pins.

### DIFF
--- a/flow/platforms/asap7/verilog/stdcell/dff.v
+++ b/flow/platforms/asap7/verilog/stdcell/dff.v
@@ -1,32 +1,41 @@
-// Quick and dirty reimplementation of DFFHQNx2_ASAP7_75t_R because
-// Verialtor doesn't support and has no plans to support 1995 UDP
-// tables.
+// Quick and dirty reimplementation of DFFHQNx2_ASAP7_75t_R
+// because verilator doesn't support and has no plans to
+// support 1995 UDP tables.
 //
 // https://github.com/verilator/verilator/issues/5243
 
 module DFFHQNx1_ASAP7_75t_R (QN, D, CLK);
-    output reg QN;
+    output QN;
     input D, CLK;
-
+    reg IQNN;
+    wire IQN;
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty Shooting from the hip here. I don't find IQNN, nor IQN exactly in the original behavioral Verilog(which Verilator can't support), but they do exist in the .lib file:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/0ff3a4b899de83f7a2fde97ff80d78d51e5c9891/flow/platforms/asap7/verilog/stdcell/asap7sc7p5t_SEQ_RVT_TT_220101.v#L44

```
  cell (DFFASRHQNx1_ASAP7_75t_R) {
    [deleted]
    pin (QN) {
      direction : output;
      function : "IQN";
      power_down_function : "(!VDD) + (VSS)";
      related_ground_pin : VSS;
    [deleted]
    ff (IQN,IQNN) {
      clear : "!SETN";
      clear_preset_var1 : L;
      clear_preset_var2 : L;
      clocked_on : "CLK";
      next_state : "!D";
      power_down_function : "(!VDD) + (VSS)";
      preset : "!RESETN";
    }
  }
```

maliberty:
I disagree with this change.  There is no need for these in Verilog.  The state nodes are internal to Liberty.  Perhaps you can expand on what the problem you are trying to solve is.

oharboe:
> I disagree with this change. There is no need for these in Verilog. The state nodes are internal to Liberty. Perhaps you can expand on what the problem you are trying to solve is.

I'm finding pins that are [not annotated](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2678/commits/76b22a624a7909ddbaf6b15dc7ff1f64bf3955f8#r1913536070) with activities by the .vcd file

maliberty:
We seem to be having this conversation in two PRs.  Let's close this and discuss in the other.

PR #2678, opened 2025-01-13
[DO NOT REVIEW] Mock array check pins

oharboe:
Needs an update after https://github.com/The-OpenROAD-Project/OpenSTA/pull/258 or some feature like it is available in OpenSTA.

### DIFF
--- a/flow/designs/asap7/mock-array/power.tcl
+++ b/flow/designs/asap7/mock-array/power.tcl
@@ -1,3 +1,5 @@
+source $::env(SCRIPTS_DIR)/util.tcl
+
 foreach libFile $::env(LIB_FILES) {
   if {[lsearch -exact $::env(ADDITIONAL_LIBS) $libFile] == -1} {
     read_liberty $libFile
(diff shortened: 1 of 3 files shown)


oharboe:
@maliberty Strange, any idea what it means that there are 366436 pin activities in the .vcd file when there are only 362274 pins in total?

(366436-362274)-(64*65) = 2

Trying to make some guesses: there are 65 pins per element not accounted for + 2 more pins.

I believe that `get_pins -hiearchical *` does not include all pins in the .vcd file as the below didn't list anything:

```
$ git diff
diff --git a/power/ReadVcdActivities.cc b/power/ReadVcdActivities.cc
index e6510c18..f5d3a86f 100644
--- a/power/ReadVcdActivities.cc
+++ b/power/ReadVcdActivities.cc
@@ -194,6 +194,9 @@ ReadVcdActivities::setVarActivity(const char *pin_name,
                duty);
     if (sdc_->isLeafPinClock(pin))
       checkClkPeriod(pin, transition_count);
+    if (annotated_pins_.find(pin) != annotated_pins_.end()) {
+      printf("Setting activity for pin more than once %s\n", pin_name);
+    }
     power_->setUserActivity(pin, activity, duty, PwrActivityOrigin::vcd);
     annotated_pins_.insert(pin);
   }
```

flow/designs/asap7/mock-array/power.tcl
```diff
@@ -18,9 +20,31 @@ for {set x 0} {$x < 8} {incr x} {
...
+    if {[llength $unannoted] >= 10} {
+      break
+    }
+  }
+}
+
+if {[llength $unannoted] > 0} {
+  puts "Error: Listing [llength $unannoted] pins without user power activity from $vcd_file:"
+  foreach pin $unannoted {
+    puts "[get_full_name $pin]"
```
oharboe, line 44: IQN and IQNN pins were reported here

maliberty: I don't think it will matter if they are annotated or not (@tspyrou confirm?).  I think it would be unusual to have the q/qn pins in the Verilog of any PDK so I wouldn't build a methodology around that assumption.

oharboe: Is there a way I can identify programatically the pins that don't need activities from the .vcd file?

maliberty: I haven't tried it but I would guess you could get the Liberty pin and check its direction != internal

oharboe: Will try some ChatGPT + hacking once I have an answer from @tspyrou whether these pins should have power activities from .vcd or not.

PR #2681, opened 2025-01-15
make: open/gui_ logs seconds for operations that take longer than 10 seconds

oharboe:
Example `make gui_cts` output for operations that take longer than 10 seconds.

This does make the logging output non-idempontent, let us see if that breaks any tests. For bazel-orfs it is not a problem, because the .log files are not a dependency on subsequent targets.

```
read_db .../4_cts.odb
  11 seconds
read_sdc .../4_cts.sdc
  13 seconds
estimate_parasitics -placement
  345 seconds
gui::update_timing_report 
  952 seconds
```

### DIFF
--- a/flow/scripts/open.tcl
+++ b/flow/scripts/open.tcl
@@ -2,6 +2,15 @@ source $::env(SCRIPTS_DIR)/util.tcl
 # Read liberty files
 source $::env(SCRIPTS_DIR)/read_liberty.tcl
 
+proc log_seconds {cmd args} {
+  set start_time_seconds [clock seconds]
+  log_cmd $cmd {*}$args
+  set total [expr {[clock seconds] - $start_time_seconds}]
+  if {$total > 0} {
+    puts "  $total seconds"
+  }
+}
+
 # Read def
 if {[env_var_exists_and_non_empty DEF_FILE]} {
     # Read lef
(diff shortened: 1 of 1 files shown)


maliberty:
This misses the read_lef/def path.  Also read_liberty can take more than 10s on large files but that code is shared and not open/gui_ specific.

oharboe:
> This misses the read_lef/def path. Also read_liberty can take more than 10s on large files but that code is shared and not open/gui_ specific.

So you think this could be generally useful? 

If so, I'll complete.

It was something I quickly cobbled together to understand where the time went.

maliberty:
Personally I don't need the timing printed but if it is useful to you I don't object.  If possible 
```
read_db .../4_cts.odb  (11 seconds)
```
would be more compact

oharboe:
> Personally I don't need the timing printed but if it is useful to you I don't object. If possible
> 
> ```
> read_db .../4_cts.odb  (11 seconds)
> ```
> 
> would be more compact

Yes, better. I'll let it lie for now and if we decide it is useful, I'll finish it up.

Initially I was interested in knowing what took so long when opening our design. I need the Timing->Update(even with 1 path it takes long), so there's nothing for it for now until some progress is made on performance or some method of mitigating the long running times(private issue filed with reproducible case).

oharboe:
One thought was to modify log_cmd directly to use your suggested syntax above, but perhaps that's a bit invasive.

I think a bit of pondering before making changes here is probably a good idea.

PR #2687, opened 2025-01-18
enabled drt antenna repair for ihp-sg13g2 designs except jpeg

jeffng-or:
Enabled DRT antenna repair for all ihp-sg13g2 designs, except jpeg. jpeg runs into GRT congestion issues when enabled.


jeffng-or:
Looks like Jenkins finally passed

maliberty:
You should update the metrics to capture the current state of antenna repair now that it is enabled.

jeffng-or:
abandoning this PR, since https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/b0cd4a380553be753b62ef38a201b43a7b5c0e97 addresses the missing metric

PR #2689, opened 2025-01-18
[DO NOT REVIEW] synthesis: add SYNTH_STRIP to improve canonicalization

oharboe:
@povik I was hoping that you could help me. I'm trying to reduce the number of rebuilds when working with bazel. I'm doing architectural exploration and this can move around the line numbers in the Verilog code, which results in an .rtlil file that changes, even when the underlying RTL does not change.

Is there a way to make the .rtlil maximally canonical so that I minimze the number of downstread ORFS builds?

```
make SYNTH_STRIP=1 clean_synth synth
```

Here are examples of lines that I'd like not to not emit or not change when I change line numbers in the Verilog files or .lib files when I invoke `write_rtlil`:

```
# Generated by Yosys 0.48 (git sha1 aaa534749, clang++ 18.1.3 -fPIC -O3)
attribute \src "/home/oyvind/OpenROAD-flow-scripts/flow/designs/src/gcd/gcd.v:109.1-320.10"
cell $_AND_ $auto$liberty.cc:108:parse_func_reduce$1
wire $eq$/home/oyvind/OpenROAD-flow-scripts/flow/designs/src/gcd/gcd.v:195$994_Y
connect \ZN $auto$rtlil.cc:2977:AndGate$2
```

### DIFF
--- a/flow/scripts/synth_canonicalize.tcl
+++ b/flow/scripts/synth_canonicalize.tcl
@@ -2,5 +2,11 @@ source $::env(SCRIPTS_DIR)/synth_preamble.tcl
 hierarchy -check -top $::env(DESIGN_NAME)
 # Get rid of unused modules
 opt_clean -purge
+
+if {$::env(SYNTH_STRIP)} {
+  puts "Stripping source code information"
+  attrmap -remove \src
+}
+
 # The hash of this file will not change if files not part of synthesis do not change
 write_rtlil $::env(RESULTS_DIR)/1_synth.rtlil
--- a/flow/scripts/variables.yaml
+++ b/flow/scripts/variables.yaml
@@ -166,6 +166,16 @@ SYNTH_GUT:
     SRAMs, for instance.
   stages:
     - synth
+SYNTH_STRIP:
+  description: >
+    Strip all debug information, such as source and line numbers.
+
+    Use-case: avoid many hours of rebuilds in bazel in architectural
+    exploration where Verilog line numbers can change without RTL
+    changing.
+  stages:
+    - synth
+  default: 0
 SYNTH_HIERARCHICAL:
   description: >
     Enable to Synthesis hierarchically, otherwise considered flat synthesis.


oharboe:
Will reopen/create new PR if there is a way forward

povik:
Hi Øyvind,

I'm happy to help but I'm sceptical we can make a dent without changing Yosys itself. FWIW there's issue YosysHQ/yosys#3713 about Yosys sensitivity to input.

Some suggestions:

 * we can add an option to write_rtlil to remove the first `# generated by` line, if that helps
 * about `cell $_AND_ $auto$liberty.cc:108:parse_func_reduce$1` we can remove those from the .il since they are reloaded in synth_preamble anyway for the actual synthesis run
 * with `wire $eq$/home/oyvind/OpenROAD-flow-scripts/flow/designs/src/gcd/gcd.v:195$994_Y` we might be able to get somewhere with `rename`; if not, we can look into changing the Verilog frontend to have a different ID generation scheme on request

So in your use case, the input is in the same order and equivalent, just the line numbers are slightly different?

povik:
> * about `cell $_AND_ $auto$liberty.cc:108:parse_func_reduce$1` we can remove those from the .il since they are reloaded in synth_preamble anyway for the actual synthesis run

I just tested `delete =A:liberty_cell` should work for this. `A:liberty_cell` selects all modules with the `liberty_cell` attribute (we ask for this attribute in the `read_liberty` command line); the `=` is there to select blackboxes/whiteboxes too, as those would be usually ignored in module selection

oharboe:
@povik Thanks!

The goal is that I should be able to make non-functional changes to the Verilog and have canonical .rtlil(same hash) so that bazel does not trigger a full ORFS flow.

The non-functional changes that I'd like to be able to ignore that I can think of are:

- The version of yosys should not be in the write_verilog, because ORFS should not rerun if the output from Yosys did not change. The version number of yosys in .rtlil should retrigger a rerun of Yosys synthesis.
- Source file names, order and line number changes should be ignored. The job of the .rtlil file as I've used in ORFS is to canonicalize the input. Verilog module names should retrigger a build.

I prefer som explicit support in Yosys for this(options to write_rtlil?), rather than try to hack it with deleting things or try to write hard to maintain ORFS scripts to rename variables.

povik:
> The version of yosys should not be in the write_verilog, because ORFS should not rerun if the output from Yosys did not change. The version number of yosys in .rtlil should retrigger a rerun of Yosys synthesis.

OK, so an option on write_verilog would suit you. Though I am not sure how often the version of Yosys would change and the synthesis result won't.

> Source file names, order and line number changes should be ignored. The job of the .rtlil file as I've used in ORFS is to canonicalize the input. Verilog module names should retrigger a build.

Ignoring order changes can be hard, as you need a way to identify equivalent elements based on some kind of graph isomorphism and attach consistent naming to them. Unless... we use RTL names of variables to base it on. Variable names triggering a build is OK too?

oharboe:
I need the option on write_rtlil to avoid rerunning synthesis, but I would also use a similar option on write_verilog to avoid pointless rebuilds.

Ordering of module names or files or verilog names should not cause a rebuild(so identical .rtlil), but yes, renamed modules would have to be rebuilt.

In my use-case I have a big .sv file generated and if unused Verilog modules (not reachable from current top) change, then this causes a change in .rtlil(even if I have tried in ORFS to purge unused modules), which causes  synthesis and a full ORFS flow.

What I'm doing is to create two RTL versions of a design across a range of architectural parameters(generating Verilog from Chisel), then I run a large number of full RTL flows on all of these variants and finally I plot PPA numbers for the A/B comparsion. This tells me which design is scales best across thes sweep of architectural parameters.

In this case, there is a lot to be saved when avoiding unecessary rebuilds. :-)

povik:
> but yes, renamed modules would have to be rebuilt.

Including a rename of a signal inside the module. A rebuild in that case is fine too, right?

oharboe:
> > but yes, renamed modules would have to be rebuilt.
> 
> Including a rename of a signal inside the module. A rebuild in that case is fine too, right?

Yes: that *should* trigger a rebuild to update the ORFS outputs.

povik:
What about changing the order of statements within a module?

So
```
    wire a = c && d;
    wire e = f && g;
```
vs
```
    wire e = f && g;
    wire a = c && d;
```

oharboe:
> What about changing the order of statements within a module?
> 
> So
> 
> ```
>     wire a = c && d;
>     wire e = f && g;
> ```
> 
> vs
> 
> ```
>     wire e = f && g;
>     wire a = c && d;
> ```

I think that is a bridge too far and I fear that going down this path opens up a can of worms.

I would be delighted if the same modules(after cleaning out unused ones), when identical, do not trigger a rebuild, regardless of source file names and which source files the modules are occur in.

oharboe:
To give you an idea: you can with Chisel(and presumably other alternatives such as Amaranth) choose to geneate individual or a single .sv file. A single .sv file is convenient when working with a build system where the automatically generated .sv file is fed directly into ORFS and the user doesn't care much if it is a single or multiple .sv files.

oharboe:
A single .sv file with Chisel or Amaranth is convenient as an interface to ORFS as which modules are added and removed is a result of changing Chisel or Amaranth files, the Verilog files and modules are computer generated.

flow/scripts/synth_canonicalize.tcl
```diff
@@ -2,5 +2,11 @@ source $::env(SCRIPTS_DIR)/synth_preamble.tcl
 hierarchy -check -top $::env(DESIGN_NAME)
 # Get rid of unused modules
 opt_clean -purge
+
+if {$::env(SYNTH_STRIP)} {
+  puts "Stripping source code information"
+  attrmap -remove \src
```
povik, line 8: `setattr -unset src` might be a better command for this since it doesn't log each attribute instance it removes. Also you need `setattr -mod -unset src` to strip the attribute on modules. With `attrmap` second command is needed too

povik:
@oharboe, we're considering pulling https://github.com/povik/yosys-slang into ORFS to get better SV support, but it might also behave better for your use case.

I just tested it on

```
module top(output [7:0] c, output [6:0] c2, input [7:0] clk); 
	register1 r1(c, c + 1, clk);

	register2 r2(c2, c2 + 2, clk);
endmodule

module register1(output reg [7:0] q, input [7:0] d, input clk);
	always_ff @(posedge clk)
			q <= d;
endmodule

module register2(output reg [6:0] q, input [6:0] d, input clk);
	always_ff @(posedge clk)
			q <= d;
endmodule
```

When I placed the modules in different files or the same file, when I added extra unused modules, when I added comments changing the line numbers, the exported `.il` was the same as long as I stripped the source attrbutes.

@maliberty FYI, yosys-slang might work better for Øyvind's use case here

oharboe:
@povik Very interesting! Nice to see that LLVM CIRCT and Yosys are now coming together :-) 

@hovind FYI

povik:
> Nice to see that LLVM CIRCT and Yosys are now coming together :-)

I know CIRCT are looking at slang but this is unrelated effort.

oharboe:
> > Nice to see that LLVM CIRCT and Yosys are now coming together :-)
> 
> I know CIRCT are looking at slang but this is unrelated effort.

Good to know! I'm having a bit of trouble understanding all the parts related to LLVM and CIRCT and now slang.

PR #2690, opened 2025-01-18
openroad: bump

oharboe:
Hopefully the computer says yes to getting latest OpenROAD so I can get a docker image.


oharboe:
autoupdated openroad bump is now newer

PR #2701, opened 2025-01-21
logging: stringly typed fixes

oharboe:
On master, wrong, missing space before "0":

```
repair_timing -setup_margin0
```

This PR fixes it:

```
repair_timing -setup_margin 0
```


maliberty:
Your description is confusing.  The command is report_checks and I would think you want the space between the flag and value.

oharboe:
> Your description is confusing. The command is report_checks and I would think you want the space between the flag and value.

Fixed pull request message.

maliberty:
I'm still confused - you want `repair_timing -setup_margin0` to be the correct result?

oharboe:
@maliberty the explanation in the PR should be legible now and this should be good to go

maliberty (approved):

PR #2694, opened 2025-01-20
synth: nudge .rtlil towards cleaner and more canonical

oharboe:
@povik A small step per your tip in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2689#issuecomment-2602747185 and also a non-trivial reduction in size(the effect is the largest on smallest designs).

```
$ wc -l results/nangate45/gcd/*/1_synth.rtlil 
   990 results/nangate45/gcd/base/1_synth.rtlil
  9524 results/nangate45/gcd/master/1_synth.rtlil
 10514 total
```

PR #2727, opened 2025-01-28
results: add 3_2_place_iop.tcl and 2_3_floorplan_macro.tcl to results

oharboe:
Useful when testing a failure in master to check if it is due to macro or pin placement

I wanted to try this on: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2694

Confirming that the failure is not in synthesis. A slight change in initial conditions cause a significant change in IO placement, which causes a grt failure.

First I run on master with https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2727

```
make FLOW_VARIANT=master DESIGN_CONFIG=designs/asap7/ethmac_lvt/config.mk place
make FLOW_VARIANT=master DESIGN_CONFIG=designs/asap7/ethmac_lvt/config.mk gui_3_2_place_iop
```

image

Then I run with this https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2694, but with macro and pin placement from master and pin placement is unchanged:

```
make DESIGN_CONFIG=designs/asap7/ethmac_lvt/config.mk IO_CONSTRAINTS=results/asap7/ethmac_lvt/master/3_2_place_iop.tcl do-3_2_place_iop
```

image

global route then passes:

image

### DIFF
--- a/flow/scripts/io_placement.tcl
+++ b/flow/scripts/io_placement.tcl
@@ -5,6 +5,7 @@ if {![env_var_equals IS_CHIP 1]} {
   load_design 3_1_place_gp_skip_io.odb 2_floorplan.sdc
   source $::env(SCRIPTS_DIR)/io_placement_util.tcl
   write_db $::env(RESULTS_DIR)/3_2_place_iop.odb
+  write_pin_placement $::env(RESULTS_DIR)/3_2_place_iop.tcl
 } else {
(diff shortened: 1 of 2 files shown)


maliberty (approved):
I'm not sure how this is more useful than the .odb but it shouldn't matter much

oharboe:
> I'm not sure how this is more useful than the .odb but it shouldn't matter much

It supports the use-case above, which can be useful in local testing.

oharboe:
@maliberty automerge reapply, I made a small fix

PR #2732, opened 2025-01-28
Check conn error config

jeffng-or:
committing checkConnectivity error configs for safe keeping
Do not merge


jeffng-or:
Confirmed fixes with OR 4b80e42bb

PR #2738, opened 2025-01-29
bazel: add cred_helper.py to be able to use artifact repository more easily

oharboe:
same scheme as in megaboom


maliberty:
The 'scan code' failed as you added this to flow/ where we white list files.  This doesn't really belong there - please move to util.

oharboe:
> The 'scan code' failed as you added this to flow/ where we white list files. This doesn't really belong there - please move to util.

Fixed

maliberty (approved):

PR #2741, opened 2025-01-29
added option for padding == 0 in asap7 mock-alu AutoTuner config

jeffng-or:
Updated asap7 mock-alu AutoTuner config to support detailed/global placement padding of 0.

63 AutoTuner errors all of type DRT-0255| (Maze Route cannot find path of net...)

### DIFF
--- a/flow/designs/asap7/mock-alu/autotuner.json
+++ b/flow/designs/asap7/mock-alu/autotuner.json
@@ -35,16 +35,16 @@
     "CELL_PAD_IN_SITES_GLOBAL_PLACEMENT": {
         "type": "int",
         "minmax": [
-            1,
-            5
+            0,
+            4
         ],
         "step": 1
     },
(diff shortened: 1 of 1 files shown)


flow/designs/asap7/mock-alu/autotuner.json
```diff
@@ -35,16 +35,16 @@
...
         ],
         "step": 1
     },
     "CELL_PAD_IN_SITES_DETAIL_PLACEMENT": {
         "type": "int",
         "minmax": [
-            1,
-            5
+            0,
+            4
```
maliberty, line 47: It doesn't make sense to allow 4 for detailed when global only goes up to 3

jeffng-or: oops. my bad. updated to both be 0,4

maliberty (approved):

PR #2749, opened 2025-01-30
mock-array: fix large negative slack for Element

oharboe:
we want slight negative slack for testing purposes.

Ended up with the simplest solution I could think of, just reduce the output width of the multiplier and let synthesis optimize away the rest.

The simpler Element also speeds up Verilator significantly.

image


maliberty:
I would assume this requires a metrics update

oharboe:
> I would assume this requires a metrics update

No, this affects reg2reg paths inside the Element only. There are no metrics for Element, only for mock-array.

maliberty (approved):

PR #2753, opened 2025-01-30
added valid config check

jeffng-or:
Added a valid config check to verify things like the global place padding >= detail place padding.

The check is done in the setup() method since we still have the data in a convenient form (dict) vs. dealing with the post processed data string that's held in the self.parameters. The action to skip the config is in the step() method, where we skip calling OR and just return ERROR_METRIC as the metric score.

### DIFF
--- a/tools/AutoTuner/src/autotuner/distributed.py
+++ b/tools/AutoTuner/src/autotuner/distributed.py
@@ -100,11 +100,18 @@ class AutoTunerBase(tune.Trainable):
         )
         self.step_ = 0
         self.variant = f"variant-{self.__class__.__name__}-{self.trial_id}-or"
+        # Do a valid config check here, since we still have the config in a
+        # dict vs. having to scan through the parameter string later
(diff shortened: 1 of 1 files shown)


maliberty (approved):

tools/AutoTuner/src/autotuner/distributed.py
luarss, line 1: In terms of valid config check, would you prefer to fail fast? I.e. fail at the argparse stage or wait till ray servers are started

jeffng-or: The failure would need to be at the individual run level, not at argparse.  In other words, someone could do:

```
global: 0, 5
detailed 0, 5
```
then, AT chooses the following for a specific run:
```
global: 2
detailed: 4
```
which is invalid and we shouldn't even bother running OR.

I've got a checker that I will commit at some point that handles what we can check at the config file level.

PR #2755, opened 2025-01-31
bazel cleanup and asap7/gcd bazel placement density study

oharboe:
Studied PLACE_DENSITY for asap7/gcd to see what happened as placement density was increased. I also wanted a clean example of how to add a module to bazel and to clean up the top level BUILD.bazel a bit.

The goal of asap7/gcd is to be a quick smoketest, it will actually run through route with PLACE_DENSITY=0.99

image

image


maliberty (approved):

oharboe:
Interestingly swerv_wrapper fails for 0.44 and 0.46 with routing congestion, but not for the other values between 0.20 to 0.49 inclusive. 0.44 and 0.46 just have some unfortunate placement w.r.t. routing.

image

image

image

Aha.... the macro placement for 0.44 and 0.46 is probably unfortunate, for 0.49 it works:

image

image

PR #2759, opened 2025-02-01
Update docker instructions to clarify that you need to specify the tag when running docker

mguthaus:
Clarify that you must specify the tag of the built image or else it will download the "latest" image from docker hub, which is, in fact, old. At the time of this, it is 7+ months old.


maliberty:
@vvbandeira is there a reason latest is so old?

oharboe:
add a link to officially published docker images?


use git describe tag version in example, not hash

mguthaus:
Is that a question for me?

mguthaus:
As far as I know, these are different than the published docker images. 

The current docker build script does not tag the image.

PR #2770, opened 2025-02-03
yosys: shuffle seed

oharboe:
If CI falls over because of shuffled seeds, then this merits closer study of whether the config.mk parameters are too tight or CI baseline is too strict.


oharboe:
@povik @widlarizer FYI

oharboe:
This was just to capture a list of test that fail with a changed --hash-seed to yosys

oharboe:
@maliberty Some grt DRC failures are gone now with OpenROAD upgrade, others are still there.

PR #2777, opened 2025-02-04
sky130hd/ibex: add bazel

oharboe:


maliberty:
Is there some CI for these bazel setups?  I afraid that updates to config.mk will not get reflected here and these will become stale.

oharboe:
I think for now I am just sharing some test use cases and I am learning how to use bazel where it shines, for now studying variations in parameter values.

Also, I am learning and adapting bazel-orfs to handle ORFS test cases nicely.

maliberty:
Does it make sense to merge this or treat it as a private experiment?

oharboe:
> Does it make sense to merge this or treat it as a private experiment?

If there is no sharing of examples, no-one is going to take a look at bazel-orfs with ORFS examples.

I'm thinking that a use-case / users might come along... Nice to have some cases to point to that are less complicated than megaboom.

maliberty (approved):

PR #2784, opened 2025-02-05
env: cleanup and comments

oharboe:


dev_env.sh
```diff
@@ -1,9 +1,16 @@
...
     export CMAKE_INSTALL_RPATH=$DIR/dependencies/lib:$DIR/dependencies/lib64
-}
 
-setpaths
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        export CMAKE_PREFIX_PATH="$(brew --prefix or-tools)"
+    fi
+}
+__setpaths
+unset -f __setpaths
```
maliberty, line 16: What is the purpose of this?

env.sh
```diff
@@ -1,24 +1,29 @@
...
-export PATH=${DIR}/tools/install/yosys/bin:$PATH
+  # Set user paths and environment variables here,
+  # developer settings go in ./dev_env.sh
+  export PATH=${DIR}/tools/install/OpenROAD/bin:$PATH
+  export PATH=${DIR}/tools/install/yosys/bin:$PATH
 
-if [[ "$OSTYPE" == "darwin"* ]]; then
-  export PATH="/Applications/KLayout/klayout.app/Contents/MacOS:$PATH"
-  export PATH="$(brew --prefix bison)/bin:$(brew --prefix flex)/bin:$(brew --prefix tcl-tk)/bin:$PATH"
-  export CMAKE_PREFIX_PATH="$(brew --prefix or-tools)"
```
maliberty, line 21: @stefanottili is this ok for mac?  I have no way to know?

dev_env.sh
oharboe: clean up the namespace

oharboe: I don't know how to create an anonymous function and hide the DIR variable. I've seen collisions with DIR in other scripts as the idiom is common.

oharboe: Since Copilot/ChatGPT knows what this means, I'm thinking a comment to the effect of "clean up the namespace" is noise at these days and likely a liability because the comment is likely to become outdated compared as the rest of the code is updated.

The code in the `env.sh` file does the following:

1. `__setpaths`: This line calls a function named `__setpaths`. The function is expected to be defined elsewhere in the script or in a sourced script. The purpose of this function is likely to set up environment paths.

2. `unset -f __setpaths`: This line removes the definition of the `__setpaths` function from the shell environment. The `-f` option specifies that a function is being unset.

In summary, the script calls the `__setpaths` function to set up paths and then removes the function definition to clean up the environment.

env.sh
vvbandeira: This is only required to correctly find `or-tools` in macOS during compilation, moving to `dev_env.sh` should be fine.

env.sh
```diff
@@ -1,24 +1,29 @@
...
+  else
+    DIR="$(dirname $(readlink -f "${BASH_SOURCE[0]}"))"
+  fi
 
-if [ -f /opt/rh/rh-python38/enable ]; then
-  source /opt/rh/rh-python38/enable
-fi
+  if [ -f /opt/rh/rh-python38/enable ]; then
+    source /opt/rh/rh-python38/enable
+  fi
```
vvbandeira: Since we are cleaning up, we can drop these lines as they were for old versions of Centos/RHEL.

dev_env.sh
vvbandeira: If the conflict is with `DIR` name, could we use another name? I think we only use the variable here and do not export it.

vvbandeira:
@oharboe, we had some issues in our CI infrastructure. I have restarted the CI build.

env.sh
oharboe: fixed

dev_env.sh
oharboe: No matter what name we pick here, there is a chance that this variable name is copied and pasted and used in another script that conflicts with this one.

Using a function here is the idiomatic way in bash to make sure that the global name space is not polluted.

vvbandeira (approved):

PR #2786, opened 2025-02-05
DependencyInstaller.sh defaults to no option

### LINKED ISSUE #2771: Danger! `sudo etc/DependencyInstaller.sh` poisons operating system when passed no arguments
### Subject

### Describe the bug

```
$ sudo etc/DependencyInstaller.sh
$ which -a yosys
/usr/local/bin/yosys
```

### Expected Behavior

/usr/local is now poisoned by etc/DependencyInstaller.sh, I have no robust and simple way to uninstall it. The etc/DependencyInstaller.sh should only set up files in ORFS/dependencies and use package installers(apt in Ubuntu), not mangle /usr/local.

Modifying /usr/local should only be possible if passed an explicit option to do so: I explicitly asked for it and I deserve the result.

As I understand it, modifying /usr/local is for the Docker use-case, not for developer or user installations.

### Environment

```shell
v3.0-2247-gce1384516
```

### To Reproduce

See above

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

vvbandeira:


vvbandeira:
@sombraSoft FYI

etc/DependencyInstaller.sh
```diff
@@ -309,6 +317,14 @@ while [ "$#" -gt 0 ]; do
     shift 1
 done
 
+if [[ "${option}" == "none"  ]]; then
+        echo "You must use one of: -all|-base|-common" >&2
+        _help
```
oharboe, line 320: Very good. 

Add "exit 1" here, no argument is an error

oharboe (changes requested):
nit

etc/DependencyInstaller.sh
vvbandeira: `_help` already does `exit 1`. Would you like for me to add it as a redundancy?

oharboe: Ah! Good! Much better.

Ignore my "add exit 1" comment here.

maliberty (approved):

PR #2794, opened 2025-02-06
synth: invalid SDC_FILE_CLOCK_PERIOD now gives error instead of silence

oharboe:


maliberty:
What error do you get?

oharboe:
> What error do you get?

before silence on invalid file name now an error from tcl trying to open the file

oharboe:
The code I deleted is copy and paste cruft, probably.

With the fix, I get a sensible error message, instead of broken behavior and silence:

```
$ make do-synth SDC_FILE_CLOCK_PERIOD=blah do-yosys
[deleted]
Extracting clock period from SDC file: blah
ERROR: TCL interpreter returned an error: couldn't open "blah": no such file or directory
```

maliberty:
I'm confused - if the file doesn't exist how do you get into the if block?

oharboe:
> I'm confused - if the file doesn't exist how do you get into the if block?

The problem is that if the file does not exist, then this file block is silently skipped, which is a bug.

The error is the solution here, not the problem.

If the file is specified, then it *MUST* exist or it is a bug and the build should not proceed.

maliberty (approved):

PR #2796, opened 2025-02-06
etc: use same base dir for all pkgs

vvbandeira:


etc/DependencyInstaller.sh
```diff
@@ -13,6 +13,8 @@ fi
 klayoutVersion=0.28.8
 verilatorVersion=5.026
 
+baseDir="/tmp/DependencyInstaller-ORFS"
```
oharboe: This will break on multiuser systems as this folder belongs to the user who created it.

etc/DependencyInstaller.sh
```diff
@@ -36,29 +38,26 @@ _installCommon() {
...
     fi
 
-    if [[ "$constantBuildDir" == "true" ]]; then
-        baseDir="/tmp/DependencyInstaller-ORFS"
-        if [[ -d "$baseDir" ]]; then
-            echo "[INFO] Removing old building directory $baseDir"
+    if [[ "${constantBuildDir}" == "true" ]]; then
+        if [[ -d "${baseDir}" ]]; then
+            echo "[INFO] Removing old building directory ${baseDir}"
+            rm -rf "${baseDir}"
```
oharboe: will fail on multiuser systems, see comment above

oharboe:
This will break on multiuser systems as the temp folder belongs to the user who created it.

oharboe (commented):
ad hoc file naming in /tmp is brittle, better to use idiomatic methods in bash like

tmpdir=$(mktemp -d /tmp/somefolder.XXXXXX)
echo "Temporary directory: $tmpdir"

etc/DependencyInstaller.sh
vvbandeira: This was already being set for the `constantBuildDir` option and I just moved the default value to here, but I see your point. Would adding `$(whoami)` as a suffix be enough?

oharboe: I only saw it because you moved it :-)

I recommend using something like the below such that the folder name is unique across versions of the script and users. This is idiomatic bash scripts and minimizes questions and surprises when reading the code and possibilities for things that can go wrong.

```
tmpdir=$(mktemp -d /tmp/somefolder.XXXXXX)
echo "Temporary directory: $tmpdir"
```

If you need to store something per user and have persistence across multiple invocations of the script, the best I can think of is to user a folder added to .gitignore in the source folder, which what the ORFS/dependencies folder is for.

etc/DependencyInstaller.sh
```diff
@@ -344,6 +323,17 @@ case "${platform}" in
         ;;
 esac
 
+if [[ "${constantBuildDir}" == "true" ]]; then
+    baseDir="/tmp/DependencyInstaller-ORFS-$(whoami)"
+else
+    baseDir=$(mktemp -d /tmp/DependencyInstaller-ORFS-XXXXXX)
+fi
```
vvbandeira, line 330: @oharboe, we use `mktemp` as the default behavior; the fixed path is only used if the user explicitly asks for `-constant-build-dir`. The reason for the option is described at: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2335, merged at https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2353

PR #2797, opened 2025-02-06
etc: only install klayout if needed

vvbandeira:


etc/DependencyInstaller.sh
```diff
@@ -93,17 +95,70 @@ _installUbuntuCleanUp() {
...
+        ruby ruby-dev
     echo "All dependencies installed successfully"
+    echo "Installing KLayout for aarch64 architecture"
+    git clone --depth=1 -b v${klayoutVersion} https://github.com/KLayout/klayout.git
+    cd klayout
+    ./build.sh -bin "${klayoutPrefix}"
+}
+
+_install_KLayout() {
+    klayoutPrefix=${PREFIX:-"/usr/local"}
```
oharboe, line 122: this will install a compiled klayout into /usr/local which will be a real pain to remove when a newer klayout package becomes available. It is possible to download newer packages from klayout as I recall.

I think it would better, safeer and consistent with yosys&OpenROAD to install compiled tools into ORFS/tools/install/klayout and update ORFS to prefer klayout from tools/install/klayout.

This script can then check system install of klayout and tools/install/klayout version. If klayout is installed into tools/install/klayout, then rebuilding and reinstalling into that same location should be fast as this script has already done so and if it needs updating it will be updated. Worst case, the user can easily purge tools/install/klayout to do a rebuild(puring klayout from /usr/local is much trickier).

oharboe (changes requested):
prefer system install of klayout, if needed build and install into ORFS/tools/klayout, consistent with ORFS/tools/OpenROAD and ORFS/tools/yosys. Update ORFS to prefer ORFS/tools/klayout, following yosys/OpenROAD pattern YOSYS_EXE and OPENROAD_EXE (without checking versions) in ORFS/flow/Makefile

Remove options for users in this script to mess around with /usr/local.

PR #2811, opened 2025-02-10
make: fix out of ORFS tree 'make metadata'

oharboe:


flow/util/genMetrics.py
```diff
@@ -19,7 +19,7 @@
 from glob import glob
 
 # make sure the working dir is flow/
-os.chdir(os.path.join(os.path.dirname(os.path.abspath(__file__)), ".."))
+flow_home = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..")
```
maliberty: Where is this used?

oharboe: Fixed.

flow/util/genMetrics.py
```diff
@@ -438,7 +420,7 @@ def extract_metrics(cwd, platform, design, flow_variant, output, hier_json):
         f.write(all_df.to_html())
 else:
     metrics_dict, metrics_df = extract_metrics(
-        args.flowPath,
+        os.getcwd(),
```
maliberty, line 423: How does this work?  You removed the chdir so cwd could be anywhere.

oharboe: All the ORFS example designs assume that FLOW_HOME is cwd, they use relative paths.

With this change, make metadata works for a design with relative paths with cwd outside of FLOW_HOME. This applies both to non-ORFS makefile and bazel-orfs use cases.

flow/util/genMetrics.py
```diff
@@ -102,7 +84,7 @@ def extractTagFromFile(
 
     # Open file
     try:
-        searchFilePath = os.path.join(args.flowPath, file)
+        searchFilePath = file
```
vvbandeira, line 87: It looks like we don't need this variable anymore with the change.

oharboe: true, I just wanted to keep the PR small

oharboe: leave this PR as is and refine more later? 

I dont want to mix concerns in a PR

vvbandeira: Sure, I can pick up the rest of the change.

vvbandeira (approved):

oharboe:
@vvbandeira Thanks!

PR #2816, opened 2025-02-11
[Autotuner] fix global args

luarss:


jeffng-or (approved):
Ran changes through locally and was able to run asap7 gcd through tune

vvbandeira (approved):

PR #2819, opened 2025-02-11
utils: add option to only change one rule

vvbandeira:


flow/util/genRuleFile.py
```diff
@@ -307,7 +311,15 @@ def gen_rule_file(design_dir, update, tighten, failing, variant, golden_metrics=
         else:
             rule_value = ceil(rule_value * 100) / 100.0
 
-        if OLD_RULES is not None and field in OLD_RULES.keys():
+        skip_metric = (
```
maliberty: Where is skip_metric used?

vvbandeira: I forgot to change in the if statements

maliberty (approved):

maliberty:
Needs a lint fix

PR #2827, opened 2025-02-12
Enabled code coverage in AutoTuner smoke tests

jeffng-or:
1. Enabled code coverage in AutoTuner smoke tests. Had to pull executable name into its own method so that there's one place that controls how we invoke the AutoTuner (current coverage is 66% statement: 100% in the test files, 66% in distributed.py and 45% in utils). I have a couple of branches that pulls code out into their own classes where we can unit test them more fully.
2. Includes fix to PPAImprov.evaluate which should allow us to add smoke_test_algo_eval.py to the CI tests

### DIFF
--- a/tools/AutoTuner/test/autotuner_test_utils.py
+++ b/tools/AutoTuner/test/autotuner_test_utils.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python3
+
+import os
+
+
+class AutoTunerTestUtils:
+    @staticmethod
+    def get_exec_cmd():
+        """
+        Returns the execution command based on whether this is a coverage run or
+        not.
+
+        Note that you need to run coverage combine after the runs complete to
+        get the coverage of the parent plus the child invocations
+        """
+
+        if "COVERAGE_RUN" in os.environ:
+            exec = "coverage run --parallel-mode --omit=*/site-packages/*,*/dist-packages/*"
+        else:  # pragma: no cover
+            exec = "python3"
+        return exec + " -m autotuner.distributed"
+
+
+if __name__ == "__main__":  # pragma: no cover
+    print(AutoTunerTestUtils.get_exec_cmd())
--- a/tools/AutoTuner/src/autotuner/distributed.py
+++ b/tools/AutoTuner/src/autotuner/distributed.py
@@ -216,11 +216,13 @@ class PPAImprov(AutoTunerBase):
         error = "ERR" in metrics.values() or "ERR" in reference.values()
         not_found = "N/A" in metrics.values() or "N/A" in reference.values()
         if error or not_found:
-            return ERROR_METRIC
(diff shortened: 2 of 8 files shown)


oharboe (commented):
have you considered python mock libraries  to mock dependencies?

this allows normal unit tests in seconds without manually crafting dependency injection

oharboe:
@jeffng-or [deltaDebug.py](https://github.com/The-OpenROAD-Project/OpenROAD/blob/3da73a523d9fa20f664f90e7210f25c9a370e7f1/etc/test_deltaDebug.py#L68) is a pain to test with an actual dataset, I used Python mocking instead. The tests runs in seconds instead of minutes/hours and I can launch them from the IDE.

I think an investment into AutoTuner for something similar would be a force-multiplier...

jeffng-or:
> @jeffng-or [deltaDebug.py](https://github.com/The-OpenROAD-Project/OpenROAD/blob/3da73a523d9fa20f664f90e7210f25c9a370e7f1/etc/test_deltaDebug.py#L68) is a pain to test with an actual dataset, I used Python mocking instead. The tests runs in seconds instead of minutes/hours and I can launch them from the IDE.
> 
> I think an investment into AutoTuner for something similar would be a force-multiplier...

@oharboe thanks for the feedback. Most of the smoketests are what I would call flow tests as opposed to unit tests. They need to run the real flow to verify that it works. But, for the unit tests that we'll want to add going forward, it's an interesting option to use Python mocking instead.

@luarss FYI.

oharboe:
Makes sense.. deltaDebug.py uses both, unittests with mocking and full scale integration tests.

tools/AutoTuner/test/autotuner_test_utils.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+
+import os
+
+
+class AutoTunerTestUtils:
```
luarss, line 6: Is the plan to make coverage command a dev command? If so I think we can add that to README

Also curious on what is the rough runtime - if its not too bad we can consider printing to CI logs in pretty print format.

jeffng-or: > Is the plan to make coverage command a dev command? If so I think we can add that to README
> 
> Also curious on what is the rough runtime - if its not too bad we can consider printing to CI logs in pretty print format.

The coverage run time is about the same as the normal run time to execute the tests (about 2h). Timings are from my VM (n2d-standard-32 (32 vCPUs, 128 GB Memory))

Also, from the Jenkins failures, it looks like it's not picking up the autotuner_test_utils.py, so I wonder if I need to update sys.path in the smoke tests.

jeffng-or:
Closing pull request since I need to address the use case when we call the smoke tests from test_autotuner.sh. The command line generated in AutoTunerTestUtils needs to be modified.

PR #2840, opened 2025-02-13
Added POSITIVE_SLACK variable to choose alternate SDC file for asap7 ibex

jeffng-or:
Most of our config.mk's are configured to have a negative slack value in order to continually improve the tools. In some cases, we want to have a version of a design that closes timing.

I've added a variable called POSITIVE_SLACK for the asap7 ibex example, which will choose an alternate SDC file for the run. The SDC has a larger clock period, which results in positive slack in the final design (4.12).

@tspyrou, FYI

### DIFF
--- a/flow/designs/asap7/ibex/config.mk
+++ b/flow/designs/asap7/ibex/config.mk
@@ -4,7 +4,14 @@ export DESIGN_NICKNAME        = ibex
 export DESIGN_NAME            = ibex_core
 
 export VERILOG_FILES         = $(sort $(wildcard $(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/*.v))
+
+# if FLOW_VARIANT == pos_slack, use an SDC file that has a larger clock
+# resulting in positive slack
+ifeq ($(FLOW_VARIANT),pos_slack)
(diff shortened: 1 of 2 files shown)


oharboe (changes requested):
use eq==1 for consistency in ORFS. We moved away from check against non empty a year or two back

vvbandeira:
May I suggest leveraging the `FLOW_VARIANT` variable instead of creating the new variable? With `FLOW_VARIANT`, we can add the alternative config to the CI easily to make sure we always can close timing and the sdc does not become stale.

oharboe:
> May I suggest leveraging the `FLOW_VARIANT` variable instead of creating the new variable? With `FLOW_VARIANT`, we can add the alternative config to the CI easily to make sure we always can close timing and the sdc does not become stale.

:relieved:

jeffng-or:
Thanks for the feedback, @vvbandeira and @oharboe . I changed the config.mk to key off FLOW_VARIANT == pos_slack. Does this address both of your concerns?

oharboe (approved):

vvbandeira (approved):

PR #2847, opened 2025-02-13
Removed invariant minmax for sky130hd gcd AutoTuner config

jeffng-or:
AutoTuner failed runs

|Stage|Error Code|Description|Count|
|-|-|-|-|
|DPL|DPL-0036|Detailed placement failed|6|

PR #2849, opened 2025-02-13
files to recreate GRT/DPL failure on sky130hd gcd

jeffng-or:
Created test case per https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2847


maliberty:
2487 is closed - is this still relevant?

jeffng-or:
> 2487 is closed - is this still relevant?

I never filed an issue for investigating why the AT jobs failed in DPL. Given the other projects we have going, we should focus on any DPL issues in those testcases vs. this one. Closing.

PR #2884, opened 2025-02-27
Mempool yosys slang

jeffng-or:
@povik , FYI. I've been trying to get mempool_group to work with https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2875 and the good news is that it wiggles. The bad news is that I'm running into some errors that I'm not sure how to resolve:

1. The two fake ram cells aren't getting defined prior to calling yosys-slang, so it can't find them
2. There are other misc bus size errors that I don't recall seeing in the earlier test, so not sure what's going on there

To repro, you can either do a:

```make DESIGN_CONFIG=designs/nangate45/mempool_group/config.mk```

or just call the mempool_group.tcl script in the flows directory.

```yosys -m slang < mempool_group.tcl ```


povik:
@jeffng-or I'll take a look, thanks for uploading

PR #2886, opened 2025-02-28
ci: remove stale and no longer used files

vvbandeira:


maliberty (approved):

maliberty:
Are the json files useful if we need to adjust a metrics rule and regenerate the values or do you expect to use a labelled PR for that?

vvbandeira:
These were reference files, i.e., to know the overall state of metrics (indirectly, the commit) when we updated the rules -- since the dashboard DB, we have abandoned updating these files. They should never be considered for generating rules, either with the label action or locally by hand. This PR does not block a user that wants to update the rules locally without using the label; they will need to run the design and call `make update_ok` as usual.

PR #2903, opened 2025-03-05
re-worked ref_file_check.py to test specific error conditions we are …

jeffng-or:
…looking for

Previous version of ref_file_check.py checked to see if the AT run failed, but not the specific failure we are looking for. In this rev, I've split test_files() into specific test methods, so that we can check the type of failure.

Had to update the test_autotuner.sh script to call the test class in ref_file_check.py, so that all of the test methods get executed now.


luarss:
Is `test_resume_check` flaky? I think we might have to comment that out as we figure out how to robustify that test.

jeffng-or:
Will replace with updated PR

PR #2915, opened 2025-03-07
Adding gen_at_report plus unit tests

jeffng-or:
Committing gen_at_report.py scripts for extracting information related to AutoTuner runs. Provided to customer as an example.

Includes unit tests which cover 90%+ of the statements


jeffng-or:
Tests were dying due to an import issue with utils. I wonder if we should rename either utils.py or the utils directory to avoid confusion. Usually, you don't have a file and a directory with effectively the same name.

tools/AutoTuner/src/autotuner/scripts/__init__.py
```diff
@@ -0,0 +1,4 @@
+# To get relative imports to work
+import os, sys
+
+sys.path.append(os.path.dirname(os.path.realpath(__file__)))
```
luarss, line 4: Instead of relative imports - why not add in pyproject toml?
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/tools/AutoTuner/pyproject.toml

tools/AutoTuner/src/autotuner/scripts/at_extractor_base.py
```diff
@@ -0,0 +1,67 @@
...
+## LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+## CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+## SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+## INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+## CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+## ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+## POSSIBILITY OF SUCH DAMAGE.
+###############################################################################
+
+from datetime import datetime
```
luarss, line 36: Could this be placed in `utils/`? Any user created libs that we have to import .

Then we can add `utils` to the `[tool.setuptools.packages.find]` section.

luarss:
We should probably rename the `utils.py` file as well - maybe a separate PR?

jeffng-or:
> We should probably rename the `utils.py` file as well - maybe a separate PR?

probably a good idea

tools/AutoTuner/src/autotuner/scripts/at_extractor_base.py
jeffng-or: yeah, I originally had it in utils, but moved it due to the possible name collision with utils. I'll try re-PR'ing after the dust settles

tools/AutoTuner/src/autotuner/scripts/__init__.py
jeffng-or: yeah, I wasn't real happy with the solution either, but was struggling to make it work with the unit tests plus calling from a script.

jeffng-or: @luarss , I remember now. I added the relative imports because I wanted to be able to run the script from within my local workspace. My understanding, which might be wrong, was the pyproject.toml works for when you publish the package, but maybe not within the local workspace. Is my understanding wrong?

luarss: No - it should work as long as you have the Python virtual environment activated. In your case, it should import something like:

To use the scripts:
```python
from autotuner.scripts import moduleA, moduleB
```

For this __init__.py, it then can be empty

jeffng-or:
Closing old PR given transition to other repo

PR #2918, opened 2025-03-08
makefile: make yosys-run target to explore and report on synthesis results

oharboe:
@povik FYI.

Motivated by my tinkering with parallel yosys synthesis in Bazel. To do parallel synthesis, the modules to be flattened need to be part of the source code like other build parameters and information (placement density, die area, etc, etc.)

There's precedent for extracting information from the flow and locking choices, keeping it in git and using it as a parameter for future builds:

- From a different domain, Python: requirements.txt and requirements_lock.txt. Requirements.txt is an ambigious definition of lists of modules to use(use a module equal to or newer than such and such). requirements_lock.txt contains a list of modules actually chosen at some point in time, kept in git. 
- lock automatic io pin placement
- write_macro_placement
- list of modules to keep, useful for running synthesis in parallel. I don't have a way to extract dependencies automatically kinda like .d files for .cpp files, but that could speed up synthesis even more in bazel.
- AutoTuner is in a sense a broader example of extracting information during automation/flow and use those as parameters to the design

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -1026,6 +1026,11 @@ run:
 	@mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR) $(OBJECTS_DIR)
 	($(OPENROAD_CMD) -no_splash $(if $(filter %.py,$(RUN_SCRIPT)),-python) $(RUN_SCRIPT) 2>&1 | tee $(abspath $(LOG_DIR)/$(RUN_LOG_NAME_STEM).log))
 
+export RUN_YOSYS_ARGS ?= -c $(SCRIPTS_DIR)/yosys_keep.tcl
+.phony: run-yosys
+run-yosys:
+	$(YOSYS_EXE) $(RUN_YOSYS_ARGS)
+
 # Utilities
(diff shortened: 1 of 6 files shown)


povik:
@oharboe There's potentially a simpler solution: just make the `abc` command inside Yosys parallelize over the modules.

oharboe:
> @oharboe There's potentially a simpler solution: just make the `abc` command inside Yosys parallelize over the modules.

Nice! 😌

This debug/exploration utility is generally useful and orthogonal though.

oharboe:
Also, it enables yosys automation.

oharboe:
@povik Also, in a scenario where one is tinkering with only part of the design, by splitting synthesis at the top level instead of as part of the build, there is the potential to only rebuild that part of the design. This gives the running time of that part of the design instead of the worst path. bazel also allows splitting work across machines, whereas parallel abc within yosys is limited to one machine. I dont have numbers, but for MegaBoom the difference could amount to minutes vs 30 minutes. LSU is the longest path in MegaBoom by far. It is easy to imagine lots of iterations on other parts of the design.

maliberty (approved):

PR #2920, opened 2025-03-09
synthesis: add a do-yosys-coarse stage

oharboe:
@povik Tinkering with faster/earlier synthesis reporting and parallel synthesis.

This is the minimum time to run synthesis before interesting things can be known about the design. Reports can be made about which modules will be kept, sram sizes to be realized as flip flops.

After this stage, it is possible to run the rest of synthesis in parallel for kept modules and combine the netlists.

Example of running times:

```
$ make DESIGN_CONFIG=designs/asap7/swerv_wrapper/config.mk clean_synth do-yosys-canonicalize
[deleted]
Elapsed time: 0:12.44[h:]min:sec. CPU time: user 12.02 sys 0.41 (99%). Peak memory: 1179692KB.
$ make DESIGN_CONFIG=designs/asap7/swerv_wrapper/config.mk do-yosys-coarse
[deleted]
Elapsed time: 0:31.94[h:]min:sec. CPU time: user 31.37 sys 0.55 (99%). Peak memory: 745904KB.
$ make DESIGN_CONFIG=designs/asap7/swerv_wrapper/config.mk do-yosys
[deleted]
Elapsed time: 4:04.65[h:]min:sec. CPU time: user 238.70 sys 5.90 (99%). Peak memory: 1148760KB.
```

For a larger design, like MegaBoom, this would of course be much longer and I'm not sure the relative running times are representative.

The Bazel dependency graph is static, so either a report of kept modules must be generated and stored in the bazel BUILD files, or bazel-orfs could set aside a configurable number of jobs, these jobs are hidden inside a rule to as not to pollute bazel query namespace, then each of these parallel jobs picks up its part of the synthesis work. This allows remote execution and parallel execution of the rest of the build.

The advantage of having a list of kept modules in bazel BUILD file is that the entire synthesis flow could be run in parallel. The coarse stage should run faster when run on individual modules. However, this requires maintaining that list(some automation is possible to update the list, but it is a slight cognitive load), whereas re-running the coarse synthesis would without further mindshare from the user always give the desired result.

In the case of MegaBoom, hierarchical synthesis is used to generate a macro placement, macro placement works better with hierarchical synthesis. The macro placement could be stored statically and updated semi-regularly with some automation, which would skip synthesis entirely for the purposes of creating a macro placement. There is something to be said for keeping macro placement stable during development to avoid strange effects with small seemingly unrelated RTL changes resulting a significantly different macro placement. Macro placement can be sensitive to initial conditions, a small change in input can yield a very different macro placement.

However, the actual MegaBoom build uses flattened synthesis for best results, though for everyday development hierarchical parallel synthesis would probably be a better choice sacrifizing some quality of results while iterating on RTL.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -445,8 +445,14 @@ yosys-dependencies: $(YOSYS_DEPENDENCIES)
 .PHONY: do-yosys
 do-yosys: $(DONT_USE_SC_LIB)
 	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR) $(OBJECTS_DIR)
-	(export VERILOG_FILES=$(RESULTS_DIR)/1_synth.rtlil; \
+	(export VERILOG_FILES=$(RESULTS_DIR)/1_yosys_coarse.rtlil; \
 	$(TIME_CMD) $(YOSYS_EXE) $(YOSYS_FLAGS) -c $(SYNTH_SCRIPT)) 2>&1 | tee $(abspath $(LOG_DIR)/1_1_yosys.log)
+
+.PHONY: do-yosys-coarse
+do-yosys-coarse: $(DONT_USE_SC_LIB)
+	mkdir -p $(RESULTS_DIR) $(LOG_DIR) $(REPORTS_DIR) $(OBJECTS_DIR)
+	(export VERILOG_FILES=$(RESULTS_DIR)/1_synth.rtlil; \
+	$(TIME_CMD) $(YOSYS_EXE) $(YOSYS_FLAGS) -c $(SCRIPTS_DIR)/synth_coarse.tcl) 2>&1 | tee $(abspath $(LOG_DIR)/1_1_yosys_coarse.log)
 
 .PHONY: do-yosys-canonicalize
 do-yosys-canonicalize: yosys-dependencies $(DONT_USE_SC_LIB)
@@ -456,8 +462,11 @@ do-yosys-canonicalize: yosys-dependencies $(DONT_USE_SC_LIB)
 $(RESULTS_DIR)/1_synth.rtlil: $(YOSYS_DEPENDENCIES)
 	$(UNSET_AND_MAKE) do-yosys-canonicalize
 
-$(RESULTS_DIR)/1_1_yosys.v: $(RESULTS_DIR)/1_synth.rtlil
+$(RESULTS_DIR)/1_1_yosys.v: $(RESULTS_DIR)/1_yosys_coarse.rtlil
 	$(UNSET_AND_MAKE) do-yosys
+
+$(RESULTS_DIR)/1_yosys_coarse.rtlil: $(RESULTS_DIR)/1_synth.rtlil
+	$(UNSET_AND_MAKE) do-yosys-coarse
 
 .PHONY: do-synth
 do-synth:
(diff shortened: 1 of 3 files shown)


oharboe:
@maliberty There are a ton of regressions that I suppose are due to variations in synthesis results, but I believe this is noise, I think I saw some substantial improvements. I seem to recall you have a web page that can show some useful metrics w.r.t. improvements as well as regressions for a PR?

povik:
@oharboe

> I seem to recall you have a web page that can show some useful metrics w.r.t. improvements as well as regressions for a PR?

That would be the dashboard (I've filled in the relevant commmit IDs): https://dashboard.precisioninno.com/compare?sourceAType=Commit&sourceBType=Commit&sourceBName=cc796b0dbcc51d472523fbc30487e0528352d967&sourceBID=2&sourceAName=fa7672b61c1efe8a6741ba54165a1ba8fec5ce63&sourceAID=5225

oharboe:
@povik Is this telling me that quality of results is on the balance much better? :exploding_head: 

image

povik:
Oh right, I've learned to ignore that part of the comparison -- I think it's broken. What is more reliable is the part at the bottom of the page:

1. you select a platform and all its designs

Screenshot 2025-03-09 at 17 38 51

2. you ask to filter down metrics to "PR metrics"

Screenshot 2025-03-09 at 17 39 38

3. you inspect the table at the bottom

Screenshot 2025-03-09 at 17 39 55

oharboe:
@povik This isn't really telling me much... I guess someone who stares at these numbers every day can get some sense, but there isn't a single metric that is understandable to me that I can translate into a go/no go on merging...

oharboe:
@povik It looks better than master: lots of green, some yellow and a few red. Presumably this means better(green), undecied(yellow) and worse(read).

povik:
> there isn't a single metric that is understandable to me that I can translate into a go/no go on merging...

You can at least get a sense of direction for "finish_timing_ws/tns" and "finish_design_instance_area".

>  Presumably this means better(green), undecied(yellow) and worse(read).

I think yellow is always worse by a small amount. In any case I agree with you in believing the metric changes are noise.

povik:
I don't understand why the area comparison is green here, while elsewhere it's orange for 0.11 % increase. Cc @sombraSoft:  this is nangate45/bp_fe on the link above. Is it possible the coloring is not from the comparison between "base" vs "test", but it comes from a comparison against some other hidden reference?

Screenshot 2025-03-09 at 17 57 15

oharboe:
Before, 271 seconds.

```
Log                            Elapsed seconds Peak Memory/MB
1_1_yosys                                  260           1122
1_1_yosys_canonicalize                      11           1151
```

After, 283 seconds.

```
Log                            Elapsed seconds Peak Memory/MB
1_1_yosys                                  241           1122
1_1_yosys_canonicalize                      12           1151
1_1_yosys_coarse                            30            727
```

So slightly worse running time when this information is not used for anything and no parallel synthesis takes place.

povik (approved):
The script changes for introducing the new stage look fine to me

oharboe:
@povik Is there a way to speed up the coarse synthesis to find list of modules to keep?

Quality of results does not matter, only that a reasonable list of modules to keep is arrived at because coarse synthesis is redone afterwards anyway.

Thoughts?

-nofsm, -noshare?

https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/synth.html

povik:
We could disable the `share` pass. If you remove `-v 3` from yosys arguments you can see which passes take the longest in the log footer.

povik:
> We could disable the share pass.

This is done by passing `-noshare` to `synth`

oharboe:
I guess we're waiting for multiplier and sram area estimates to be fixed in synthesis.

Out of sight, out of mind for now.

PR #2923, opened 2025-03-11
[DO NOT REVIEW] plot: plot yosys gate equivalents against area

oharboe:
@povik FYI, I tinkered with some reporting this morning out of curiosity...

Might be worth knocking into shape and merging?

```
make run DESIGN_CONFIG=designs/asap7/swerv_wrapper/config.mk floorplan
make run DESIGN_CONFIG=designs/asap7/swerv_wrapper/config.mk RUN_SCRIPT=scripts/report_area.tcl
python3 scripts/plot_keep.py reports/asap7/swerv_wrapper/base/1_keep.txt reports/asap7/swerv_wrapper/base/2_area.txt
```

image

### DIFF
--- a/flow/scripts/plot_keep.py
+++ b/flow/scripts/plot_keep.py
@@ -13,7 +13,7 @@ def load_data(filename):
     data = {}
     with open(filename) as f:
         for line in f:
-            key, value = line.split()
+            key, value = line.split()[0:2]
             data[key] = float(value)
     return data
 
@@ -32,8 +32,9 @@ def main():
     data1 = load_data(sys.argv[1])
     data2 = load_data(sys.argv[2])
     keys = set(data1.keys()) & set(data2.keys())
-    x = [data1[key] for key in keys]
-    y = [data2[key] for key in keys]
+    max_value = 3000
+    x = [data1[key] for key in keys if data1[key] < max_value and data2[key] < max_value]
+    y = [data2[key] for key in keys if data1[key] < max_value and data2[key] < max_value]
     plot_data(x, y)
 
 
--- a/flow/scripts/plot_keep.py
+++ b/flow/scripts/plot_keep.py
@@ -18,10 +18,14 @@ def load_data(filename):
     return data
 
 
-def plot_data(x, y):
+def plot_data(x, y, labels):
+    # add a label along with each data point
     plt.scatter(x, y)
-    plt.xlabel("Yosys gate equivalents")
-    plt.ylabel("OpenROAD area (um^2)")
+    for i, label in enumerate(labels):
+        plt.annotate(label, (x[i], y[i]))
+    plt.xlabel("Yosys coarse gate equivalents")
+    plt.ylabel("Yosys fine synthesis gate equivalents")
+#    plt.ylabel("OpenROAD area (um^2)")
     plt.show()
 
 
@@ -33,9 +37,13 @@ def main():
     data2 = load_data(sys.argv[2])
     keys = set(data1.keys()) & set(data2.keys())
     max_value = 3000
-    x = [data1[key] for key in keys if data1[key] < max_value and data2[key] < max_value]
-    y = [data2[key] for key in keys if data1[key] < max_value and data2[key] < max_value]
-    plot_data(x, y)
+    #keys = [key for key in keys if data1[key] > max_value and data2[key] < 1000]
+    #print(keys)
+    x = [data1[key] for key in keys]
+    y = [data2[key] for key in keys]
+    #print(list(zip(keys, x, y)))
+    # label = [key for key in keys]
+    plot_data(x, y, keys)
 
 
 if __name__ == "__main__":
--- a/flow/scripts/plot_keep.py
+++ b/flow/scripts/plot_keep.py
@@ -1,0 +1,41 @@
+# Load two files passed on the command line with dictionaries
+# of the format below, then plot XY scatter plot of the data
+#
+# rvdff_WIDTH13 13
+# rvdff_WIDTH14 14
+
+
+import sys
+import matplotlib.pyplot as plt
+
+
+def load_data(filename):
+    data = {}
+    with open(filename) as f:
+        for line in f:
+            key, value = line.split()
+            data[key] = float(value)
+    return data
+
+
+def plot_data(x, y):
+    plt.scatter(x, y)
+    plt.xlabel("Yosys gate equivalents")
+    plt.ylabel("OpenROAD area (um^2)")
+    plt.show()
+
+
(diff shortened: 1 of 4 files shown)


oharboe:
@vvbandeira some sort of CI trouble with docs?

image

povik:
I wonder if the outliers are due to missing estimation on `$macc`, let's fix it and revisit.

oharboe:
Out of sight, out of mind for now.

vvbandeira:
> @vvbandeira some sort of CI trouble with docs?

Yes, we use readthedocs to build our docs, not our own CI, and their "worker" failed. The steps below are defined internally. I am unsure if we have control over the steps, but we will look for alternatives if this keeps happening.

image

oharboe:
@povik FYI, another thing from megaboom:

```
51. Executing KEEP_HIERARCHY pass.
Warning: Can't determine cost of $mem_v2 cell (21 parameters).
```

oharboe:
megaboom 

image

oharboe:
@jeffng-or @maliberty @povik FYI, I wonder if the MAX_UNGROUP_SIZE policy is fundamelly broken on MegaBoom...

The problem is that there are lots of debug signals that get washed away in flattened synthesis, so it isn't possible to get a reasonable estimate of module size without actually flattening. We know this because e.g. Naja was able to find lots of unused signals that don't go away until you look at the entire design(flatten everything). We also did a fair amount of plotting to confirm and we've checked with the authors(RocketChip generates debug signals that are for simulation only).

So for MegaBoom, this policy is meaningless. The right approach, which is taken in MegaBoom, is to use hierarchical synthesis to get a decent automatic macro placement, then use that macro placement with flattened synthesis.


image

maliberty:
If you are willing to pay the runtime that is a viable approach.  The ideal would be for synthesis to be able to selectively uniquify without flattening so that you keep hierarchy but still get good QOR.  Perhaps naja will get good enough to allow fully hierarchical yosys output to be processed into a flat-synthesis level QOR without needing to go back through yosys again.

oharboe:
I see. There are other things holding  MegaBoom back where there is no soluuon on the horizon. The stickyiest is probably the lack of register retiming and duplication, which is documented to have avery negative impact on the FPU. I have not heard that anyone is working on that and there are good resons to let that dog lie, it fixes one problem and creates another in LEC. 

Though: retiming and register duplication would be useful in architectural exploration: it would tell us how much more mileage there is in the design, given sufficient effort.

oharboe:
The same plot as above but running synth :fine(same as master). Not materially different.


image

povik:
@oharboe Is the openroad area including the submodules or without submodules? The Yosys area is *without* kept submodules

oharboe:
DecodeUnit 3185 gates equivalents from synthesis and 74 um^2 in gui_synth. It looks like an interesting module to study(assuming my assumption that it is exhibiting pathological behavior is correct), it is really simple, no submodules, small and just basic logic and wires.

https://github.com/The-OpenROAD-Project/megaboom/blob/main/rtl/DecodeUnit.sv


Building this standalone:

```
make DESIGN_CONFIG=designs/asap7/minimal/config.mk DESIGN_NAME=DecodeUnit VERILOG_FILES=/home/oyvind/megaboom/rtl/DecodeUnit.sv
```




```
make DESIGN_CONFIG=designs/asap7/minimal/config.mk DESIGN_NAME=DecodeUnit VERILOG_FILES=/home/oyvind/megaboom/rtl/DecodeUnit.sv gui_synth
```

image


```
make DESIGN_CONFIG=designs/asap7/minimal/config.mk DESIGN_NAME=DecodeUnit VERILOG_FILES=/home/oyvind/megaboom/rtl/DecodeUnit.sv gui_floorplan
```

image


```
make DESIGN_CONFIG=designs/asap7/minimal/config.mk DESIGN_NAME=DecodeUnit VERILOG_FILES=/home/oyvind/megaboom/rtl/DecodeUnit.sv gui_final
```

image





image

oharboe:
> @oharboe Is the openroad area including the submodules or without submodules? The Yosys area is *without* kept submodules

OpenROAD can report both, my script followed the GUI: including submodules. However the DecodeUnit has no submodules.

povik:
If we have a disconnect in what's being reported then it's not surprising that a module with high Yosys/OpenROAD ratio is one without submodules. Is it easy to regenerate the plot using area without kept submodules for the OpenROAD value?

oharboe:
> If we have a disconnect in what's being reported then it's not surprising that a module with high Yosys/OpenROAD ratio is one without submodules. Is it easy to regenerate the plot using area without kept submodules for the OpenROAD value?

Here you go... I'm now plotting `$module getInsts` instead of `$module getLeafInsts`. That does appear to be what you're asking for BoomTile (the top level) is now tiny

```
  puts $f "[$module getName] [module_area [$module getInsts]] [module_area [$module getLeafInsts]]"
```

```
BoomTile 75 744302
BoomCore 2746 425469
FpPipeline 358 190072
IssueUnitCollapsing 16475 16475
FPUExeUnit 149 7901
....
```

image

povik:
Thanks, that's still not great correlation

oharboe:
@povik So, it appears DecodeUnit is worth a closer look... It doesn't have any submodules.

DecodeUnit synthesis is  3185.0 and floorplan 67.0, so a simple module with a large discrepancy:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2923#issuecomment-2715408843

povik:
> DecodeUnit synthesis is 3185.0 and floorplan 67.0, so a simple module with a large discrepancy:

That's not necessarily a discrepancy. Those are different units, and also there will be cases where fine-grained synthesis (which we don't run) is able to significantly optimize the module so we overestimate. I admit the correlation isn't great based on your plot but I'm not sure DecodeUnit is worth investigating.

oharboe:
> > DecodeUnit synthesis is 3185.0 and floorplan 67.0, so a simple module with a large discrepancy:
> 
> That's not necessarily a discrepancy. Those are different units, and also there will be cases where fine-grained synthesis (which we don't run) is able to significantly optimize the module so we overestimate. I admit the correlation isn't great based on your plot but I'm not sure DecodeUnit is worth investigating.

This the correlation between coarse and fine synthesis. Other than the multipliers, the gate equivalents seem to correlate quite well.

image

Labels work reasonably well for the outliners, the multipliers...

image

povik:
@oharboe Sorry, what I meant by fine-grained synthesis is the ABC script, I think your plots are about the correlation between `synth -run :fine;` and `procs; memory -nomap` so something else.

PR #2925, opened 2025-03-12
floorplan.tcl: writes out 2_1_area.txt report

oharboe:
@povik @maliberty Useful?

### DIFF
--- a/flow/scripts/floorplan.tcl
+++ b/flow/scripts/floorplan.tcl
@@ -193,5 +193,32 @@ if { [env_var_exists_and_non_empty POST_FLOORPLAN_TCL] } {
   source $::env(POST_FLOORPLAN_TCL)
 }
 
+set db [ord::get_db]
+set chip [$db getChip]
+set block [$chip getBlock]
+
+set scale_to_um [expr [$block getDbUnitsPerMicron] * [$block getDbUnitsPerMicron]]
+
+proc module_area {insts} {
+  global scale_to_um
+  set area 0
+  foreach inst $insts {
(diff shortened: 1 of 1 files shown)


maliberty:
Is this different from what the hierarchy browser shows?

oharboe:
> Is this different from what the hierarchy browser shows?

It should be the same. This is useful for automation, for instance correlating to nand gate equivalents in synthesis or using it as a policy to produce a list of modules to keep that is then put into config.mk

maliberty:
If it is useful to you I don't mind.  I don't have an immediate need myself.

flow/scripts/floorplan.tcl
```diff
@@ -193,5 +193,32 @@ if { [env_var_exists_and_non_empty POST_FLOORPLAN_TCL] } {
...
+    set area [expr $area + [$bbox getDX] * [$bbox getDY]]
+  }
+  return [expr $area / $scale_to_um]
+}
+
+# $REPORTS_DIR/2_1_area.txt has three columns:
+#
+# name - name of module
+# area - area of instances within module, excluding submodules
+# area_leaf - area of instances within module and submodules
```
maliberty, line 216: I think it is good to print a header line in the file itself.

oharboe:
> If it is useful to you I don't mind.  I don't have an immediate need myself.

If @povik finds it useful, then I can knock it into shape, otherwise, a closed PR keeps the script around after a fashion...

povik:
I think it's useful but saving it in a closed PR may be good enough

PR #2930, opened 2025-03-12
synth: use SYNTH_KEEP_MODULES for consistency

oharboe:
and introduce a test-case.

Also, the keep attribute is set at the same time as SYNTH_HIERARCHICAL=1 for consistency


oharboe:
@maliberty I checked the other asap7 designs, but none of them are suitable for testing SYNTH_KEEP_MODULES, they are too small, they have one or two kept modules.

designs/asap7/swerv_wrapper is too slow(ca. 1 hour) to run in CI, which is why it is excluded from CI, I presume. I ran "make metadata" locally: worked.

maliberty (approved):

oharboe:
Ah! This also fixed a problem that I ran into after this was merged, but before I upgraded locally: the keep attribute must be applied *after* do-yosys-canonicalize

PR #2938, opened 2025-03-14
Add Nix flake to ORFS.

rslawson:
Hopefully this should allow users to invoke `nix develop` (with the appropriate arguments) in order to drop into a shell with OpenROAD and Yosys. I'm uncertain whether the AutoTuner works with this since I don't have anything to test it with at the moment, but I'd be happy to get some feedback on that.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -232,15 +232,27 @@ endif
 # Priority is given to
 #       1 user explicit set with variable in Makefile or command line, for instance setting OPENROAD_EXE
 #       2 ORFS compiled tools: openroad, yosys
-export OPENROAD_EXE      ?= $(abspath $(FLOW_HOME)/../tools/install/OpenROAD/bin/openroad)
-export OPENSTA_EXE       ?= $(abspath $(FLOW_HOME)/../tools/install/OpenROAD/bin/sta)
+ifneq (${IN_NIX_SHELL}, "")
+  export OPENROAD_EXE := $(shell which openroad)
+else
+  export OPENROAD_EXE ?= $(abspath $(FLOW_HOME)/../tools/install/OpenROAD/bin/openroad)
+endif
+ifneq (${IN_NIX_SHELL}, "")
+  export OPENSTA_EXE := $(shell which opensta)
+else
+  export OPENSTA_EXE ?= $(abspath $(FLOW_HOME)/../tools/install/OpenROAD/bin/opensta)
+endif
 
 OPENROAD_ARGS            = -no_init -threads $(NUM_CORES) $(OR_ARGS)
 OPENROAD_CMD             = $(OPENROAD_EXE) -exit $(OPENROAD_ARGS)
 OPENROAD_NO_EXIT_CMD     = $(OPENROAD_EXE) $(OPENROAD_ARGS)
 OPENROAD_GUI_CMD         = $(OPENROAD_EXE) -gui $(OR_ARGS)
 
-YOSYS_EXE               ?= $(abspath $(FLOW_HOME)/../tools/install/yosys/bin/yosys)
+ifneq (${IN_NIX_SHELL}, "")
+  YOSYS_EXE := $(shell which yosys)
+else
+  YOSYS_EXE ?= $(abspath $(FLOW_HOME)/../tools/install/yosys/bin/yosys)
+endif
 
 # Use locally installed and built klayout if it exists, otherwise use klayout in path
 KLAYOUT_DIR = $(abspath $(FLOW_HOME)/../tools/install/klayout/)
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -232,15 +232,15 @@ endif
 # Priority is given to
 #       1 user explicit set with variable in Makefile or command line, for instance setting OPENROAD_EXE
 #       2 ORFS compiled tools: openroad, yosys
-ifneq (${IN_NIX_SHELL}, "")
-  export OPENROAD_EXE := $(shell which openroad)
+ifneq (${IN_NIX_SHELL},)
+  export OPENROAD_EXE := $(shell command -v openroad)
 else
   export OPENROAD_EXE ?= $(abspath $(FLOW_HOME)/../tools/install/OpenROAD/bin/openroad)
 endif
-ifneq (${IN_NIX_SHELL}, "")
-  export OPENSTA_EXE := $(shell which opensta)
+ifneq (${IN_NIX_SHELL},)
+  export OPENSTA_EXE := $(shell command -v sta)
 else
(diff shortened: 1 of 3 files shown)


maliberty (approved):
@donn @sombraSoft FYI

rslawson:
Should I also write in some docs somewhere? In my testing it may require more than just `nix develop`, for me it took `nix --extra-experimental-features "nix-command flakes" develop`.

maliberty:
Documentation is welcome.  It looks like this change fails in the CI so some more work is needed.

rslawson:
Realised just a bit ago that it wasn't actually using the same default for OpenSTA as before, so that should be fixed now. However, I'm still not sure why it fails in CI otherwise (or if it still does). I'm also failing to get the tests to run locally in Nix - the first test with AES ASAP7 also fails for me but with this error in the step `22.1. Executing DFFLEGALIZE pass (convert FFs to types supported by the target).` (I believe?):
```
ERROR: FF aes_cipher_top.dcnt[0]$_SDFFE_PN0P_ (type $_SDFFE_PN0P_) cannot be legalized: D flip-flops are not supported
Elapsed time: 0:05.08[h:]min:sec. CPU time: user 4.95 sys 0.12 (99%). Peak memory: 134552KB.
make[1]: *** [Makefile:460: do-yosys] Error 1
make: *** [Makefile:472: results/asap7/aes/base/1_1_yosys.v] Error 2
```
Does this look familiar? If not I'm actually fairly confused what's going on, since I'm pulling the same OpenROAD and Yosys versions in the Nix flake as the repository is currently using (commits starting with `ec1b` and `c4b5` respectively).

flow/Makefile
```diff
@@ -232,15 +232,27 @@ endif
 # Priority is given to
 #       1 user explicit set with variable in Makefile or command line, for instance setting OPENROAD_EXE
 #       2 ORFS compiled tools: openroad, yosys
-export OPENROAD_EXE      ?= $(abspath $(FLOW_HOME)/../tools/install/OpenROAD/bin/openroad)
-export OPENSTA_EXE       ?= $(abspath $(FLOW_HOME)/../tools/install/OpenROAD/bin/sta)
+ifneq (${IN_NIX_SHELL}, "")
```
maliberty: I think you want
```
ifneq ($(IN_NIX_SHELL),)
```
and likewise below.  That should fix the failures

rslawson: Oh no, did I actually forget to delete these lines? Amazing.

oharboe:
Is there a reason this needs to be in tree?

Can you add bazelisk to the nix env?

rslawson:
I fail to see why it wouldn't make sense to have in tree considering both Yosys and OpenROAD have their own `flake.nix`, and those two are the primary dependencies of this project, no?

As for bazelisk, probably can, but why? I don't see that in the project dependencies anywhere.

oharboe:
> I fail to see why it wouldn't make sense to have in tree considering both Yosys and OpenROAD have their own `flake.nix`, and those two are the primary dependencies of this project, no?
> 
> As for bazelisk, probably can, but why? I don't see that in the project dependencies anywhere.

I am only vaguely familiar with nix, it sets up a dev env here, right?


bazel is a build tool on the horizon for OpenROAD. Already in use in hdl_rules and bazel-orfs.

rslawson:
> I am only vaguely familiar with nix, it sets up a dev env here, right?

That's right, yes. The idea being that it's able to provide all the dependencies required for a project regardless of what system it's being used on so long as it has Nix installed. My use case here is to avoid having to fix up a non-trivial number of ebuilds for Gentoo and repackage it for a newer Ubuntu version in order to be able to use ORFS on my work and home desktops.

> bazel is a build tool on the horizon for OpenROAD. Already in use in hdl_rules and bazel-orfs

Had a peek at nixpkgs and yeah, this should be a trivial change. Just a one-liner `pkgs.bazelisk` added to `buildInputs` in the outputs of the flake.

oharboe:
> > I am only vaguely familiar with nix, it sets up a dev env here, right?
> 
> That's right, yes. The idea being that it's able to provide all the dependencies required for a project regardless of what system it's being used on so long as it has Nix installed. My use case here is to avoid having to fix up a non-trivial number of ebuilds for Gentoo and repackage it for a newer Ubuntu version in order to be able to use ORFS on my work and home desktops.
> 
> > bazel is a build tool on the horizon for OpenROAD. Already in use in hdl_rules and bazel-orfs
> 
> Had a peek at nixpkgs and yeah, this should be a trivial change. Just a one-liner `pkgs.bazelisk` added to `buildInputs` in the outputs of the flake.

Even given the bazelisk, not bazel, usage and the .bazelversion file?

Happy to give it a try when the PR has it set up.

We tried to use Nix in the past, but ran into difficulties. It is a powerful concept, but not trivial. It solves a lot of hard dependency problems.

Can docker or podman be set up too? 

bazel-orfs that is used in ORFS uses docker(or podman). The bazel-orfs podman/docker usage is just to extract executables and patchelf them, so podman being usermode is probably easier?

I fear, though, that there is going to be some "framwork inversion" problem here. podman/docker, nix and bazelisk are all package systems and/or build systems fighting to be front and center.

rslawson:
I'm afraid I'm not knowledgeable enough on Nix to be able to say definitively one way or another whether that would be possible or not. I've just started trying to learn it properly, and for various reasons I chose this repo to try and use as a learning project (namely that I want to use ORFS, but I don't want to maintain packaging for several distros). With that said, if there's a way for me to try out whether the Bazelisk stuff functions already then I'd be happy to give that a go and see if It Just Works™. Though that's probably at least somewhat predicated on figuring out why CI fails and why I can't get stuff to build locally, either.

oharboe:
I am a big fan, but not user, of Nix... My colleagues speak highly of it, but it has proven non trivial to deploy at scale.

Switching to a nix env dir can take some time, setup problems with nix itself(there is no apt get install) so install/uninstall has been flaky on ubuntu with mysterious errors only nix experts can fix and then with great difficulty.

maliberty:
The nix setup was created by Efabless for OpenLane2.  We just host it for them and don't test it at all.  I don't expect to use it in OR/ORFS development at this point.  Given the recent changes I'm not sure of its fate.

oharboe:
I am a big fan, but not user, of Nix... My colleagues speak highly of it, but it has proven non trivial to deploy at scale.

Switching to a nix env dir can take some time, setup problems with nix itself(there is no apt get install) so install/uninstall has been flaky on ubuntu with mysterious errors only nix experts can fix and then with great difficulty.

I would love to be a nix user if there is good support, but not maintainer... My learning efforts currently go into Bazel.

flow/Makefile
rslawson: Thank you very much for catching this, this fix made it work locally using Nix for me.

PR #2943, opened 2025-03-16
Synth params fix

oharboe:
Manually tested.

 I want to extend this [serv CPU bazel-orfs test case](https://github.com/olofk/serv/pull/146) with some plotting of PPA as a function of bit width.

I tried to add a test case and do some cleanup, but need some advice from @povik https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2944


maliberty:
I'll wait for @povik to respond

povik (approved):
Parameters can't be changed (so chparam won't work) once the design roundtrips through an .rtlil export. So indeed the right place for chparam is synth_canonicalize

oharboe:
> Parameters can't be changed (so chparam won't work) once the design roundtrips through an .rtlil export. So indeed the right place for chparam is synth_canonicalize

Thanks!

maliberty (approved):

PR #2944, opened 2025-03-16
[DO NOT REVIEW] Synth params test case

oharboe:
@povik Any advice on how to make the test case work?

```
$ make DESIGN_CONFIG=designs/gf180/uart-blocks/config.mk metadata
[deleted]
9. Executing AST frontend in derive mode using pre-parsed AST for module `\uart'.
9.1. Analyzing design hierarchy..
ERROR: Module `uart_rx' referenced in module `uart' in cell `uart_rx_inst' does not have a parameter named 'DATA_WIDTH'.
Elapsed time: 0:00.42[h:]min:sec. CPU time: user 0.38 sys 0.03 (99%). Peak memory: 62464KB.
make[1]: *** [Makefile:454: do-yosys-canonicalize] Error 1
make: *** [Makefile:457: results/gf180/uart-blocks/base/1_synth.rtlil] Error 2
```


povik:
We would need a Verilog wrapper around the macro block which exposes the expected parameter.

oharboe:
So the fix is fine and should go in, but the test needs future work as it is testing more than was fixed?

could you approve the PR that has only the fix?

oharboe:
out of sight, out of mind until Yosys supports this use case

povik:
I'm not sure how Yosys could support this use case better: once you compile the macro block, you have compiled it for a particular value of parameters and the `.lib` file which we supply to Yosys describes a non-parametrizable block.

oharboe:
Unclear to me too without modifying the Verilog...

This makes me think of a non sequitor: Yosys can identify memories that are not in standalone modules, so making macros out of them requires modifying Verilog.

PR #2950, opened 2025-03-18
variables: demonstrate migration path from make to tcl

oharboe:
@maliberty thoughts?


maliberty:
Is the goal here to write config.tcl in terms of envars but once the migration is complete to stop using envars and directly evaluate the file in OR when called from bazel-orfs?

oharboe:
yes, end goal. Ideally a single giant effort and be done with it, but realistically it will be a more piecemeal effort.

maliberty:
I think it is a good directional step.  To fully remove make you would have to deal with the non-tool operations too (eg megeLib.pl, various shell operations [cp, sed, etc]).  Also klayout uses python not tcl.  I wonder if it wouldn't be better to move towards something more neutral like json.

oharboe:
Indeed, but if I have time to work on this, it will have to be piecemeal....

oharboe:
I want to ponder it a bit more and see if I can come up with standalone yosys and openroad invocation examples to tie everything together.

out of sight out if mind for now.

PR #2951, opened 2025-03-18
Makefile: clean out settings.mk

oharboe:
I suspect no-one is using this


maliberty (approved):
You added this in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/5ef3dad2815b0cb179e6fd161619a0b4a526395e

povik:
> I suspect no-one is using this

People are in fact using this. I'm opening a revert (it's being synced to public)

PR #2963, opened 2025-03-19
synth: consistent with other .tcl scripts, it knows its input

oharboe:
other scripts, like floorplan.tcl, knows what inputs to load. This makes synth.tcl consistent with this convention rather than encoding this in the Makefile

### DIFF
--- a/flow/scripts/synth.tcl
+++ b/flow/scripts/synth.tcl
@@ -1,3 +1,5 @@
+set ::env(VERILOG_FILES) $::env(RESULTS_DIR)/1_synth.rtlil
+
 source $::env(SCRIPTS_DIR)/synth_preamble.tcl
 
 hierarchy -check -top $::env(DESIGN_NAME)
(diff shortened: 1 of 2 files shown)


oharboe:
@maliberty unrelated error on master:

```
[ERROR] finish__timing__drv__hold_violation_count fail test: 342.0 <= 228.0
```

flow/scripts/synth.tcl
```diff
@@ -1,3 +1,5 @@
+set ::env(VERILOG_FILES) $::env(RESULTS_DIR)/1_synth.rtlil
```
maliberty, line 1: I believe this will fail with a cached netlist.

oharboe: Good poin, though I dont see how.

Will test manually report/fix as needed

maliberty: I don't think that file will exists though I haven't tried it.

maliberty (approved):

PR #2964, opened 2025-03-19
makefile: make YOSYS_EXE and YOSYS_FLGAS available to scripting

oharboe:


oharboe:
@maliberty pr-merge is an unrelated error on master, I think.

PR #2965, opened 2025-03-19
io: more cleanup now that pin constraints are in .odb

oharboe:


oharboe:
out of sight, out of mind for now

PR #2967, opened 2025-03-19
[DO NOT REVIEW] variables: demonstrate migration path from make to bash

oharboe:
@maliberty Thoughts?

I think this path is more promising than my last attempt.

There needs to be a klayout.sh file added just like synth.sh and flow.sh.

The idea is to start here and piecemeal migrate everything to .sh files to the point where make is no longer needed for bazel-orfs.


maliberty:
While bash is better than make it still is obscure as a programming language.  The previous iteration with python seemed better in that regard.

oharboe:
Last time I used .tcl and there is a step where I have to use bash for complicated reasons. Let us do a call to discuss.

oharboe:
unsure about direction to take, found some flaws here.

Configuration operations is take a dictionary in, run some logic and modify dictionary, write dictionary out. However, the dictionary is the environment, which makes it very awkward to deal with. Also the plethora of programming paradigms and framework inversions make it hard....

platforms/config.mk files are an example of this sort of operation.

PR #2973, opened 2025-03-19
makefile: move BLOCKS section to include DESIGN_CONFIG time

oharboe:
This will assist in migration away from `make` dependency for the support for flow variables.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -89,6 +89,27 @@ DESIGN_CONFIG ?= ./designs/nangate45/gcd/config.mk
 # in this file. This allows the DESIGN_CONFIG to set different defaults than
 # this file.
 include $(DESIGN_CONFIG)
+export DESIGN_NICKNAME?=$(DESIGN_NAME)
+# default value "base" is duplicated from variables.yaml because we need it
+# earlier in the flow for BLOCKS. BLOCKS is a feature specific to Makefile
+# that will not be ported to bazel-orfs.
+export FLOW_VARIANT?=base
+ifneq ($(BLOCKS),)
+  # Normally this comes from variables.yaml, but we need it here to set up these variables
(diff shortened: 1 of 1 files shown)


flow/Makefile
```diff
@@ -89,6 +89,27 @@ DESIGN_CONFIG ?= ./designs/nangate45/gcd/config.mk
 # in this file. This allows the DESIGN_CONFIG to set different defaults than
 # this file.
 include $(DESIGN_CONFIG)
+export DESIGN_NICKNAME?=$(DESIGN_NAME)
+# default value "base" is duplicated from variables.yaml because we need it
+# earlier in the flow for BLOCKS. BLOCKS is a feature specific to Makefile
+# that will not be ported to bazel-orfs.
+export FLOW_VARIANT?=base
```
maliberty, line 96: Shouldn't this come from variables.yaml where the default is already specified?

oharboe: Yes, but it is an earth cooling problem. BLOCKS needs to be set up as part of make DESIGN_CONFIG. I think this is the lesser evil.

maliberty: Why do "BLOCKS needs to be set up as part of make DESIGN_CONFIG"?

oharboe: Because scripts later will do all sorts of messing about with wrap and libs and lefs and the bash script to set up variables need to run before make rules are set up

maliberty: Should the flow default values be setup before we even get here?

oharboe: no. Default values have priorities: first user, scripts/config.sh, then variables.yaml, then misc, then platform.

Now we can talk about changing this, but this is how it is now. One step at the time.

oharboe:
Merge? green...

maliberty (approved):

Issue #2990, opened 2025-03-23
asap7 ADDER_MAP_FILE is identical for ASAP7_USELVT=1 and ASAP7_USELVT=0


oharboe:
### Subject

[Flow] for any util, flow Makefile, or flow script issues.

### Describe the bug

This is as intended, I think:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/4004072bd74aa01fd60a76011bf64bca3c4e74f6/flow/platforms/asap7/config.mk#L83

```
$ make DESIGN_CONFIG=designs/asap7/aes/config.mk print-ADDER_MAP_FILE
ADDER_MAP_FILE = .../platforms/asap7/yoSys/cells_adders_R.v
```

I believe it was the intention to get `cells_adders_L.v` with `ASAP7_USELVT=1`

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/4004072bd74aa01fd60a76011bf64bca3c4e74f6/flow/platforms/asap7/config.mk#L178

```
$ make DESIGN_CONFIG=designs/asap7/aes_lvt/config.mk print-ADDER_MAP_FILE
ADDER_MAP_FILE = .../platforms/asap7/yoSys/cells_adders_R.v
```

For anyone debugging this, in addition to the `make print-FOO` tool, I find that it is useful to run thish command on master and on my branch:

```
make DESIGN_CONFIG=designs/asap7/aes_lvt/config.mk vars
```

This generates a vars.sh file in canonical form with all the variables the makefiles has. This vars.sh file can be compared against master and my branch:

```
$ diff vars.sh ../../OpenROAD-flow-scripts/flow/vars.sh 
5c5
< export ADDER_MAP_FILE="${FLOW_HOME}/platforms/asap7/yoSys/cells_adders_R.v"
---
> export ADDER_MAP_FILE="${FLOW_HOME}/platforms/asap7/yoSys/cells_adders_L.v"
40c40
...
```

### Expected Behavior

Different ADDER_MAP_FILE values

### Environment

```shell
ORFS: v3.0-2686-g4004072b
```

### To Reproduce

See above

### Relevant log output

```shell

```

oharboe:
It would probably be a good idea to check that everything that should be different is different+has the correct value and everything that should be the same is the same...

```
$ make DESIGN_CONFIG=designs/asap7/aes/config.mk ASAP7_USELVT=1 vars
[...]

PR #2991, opened 2025-03-23
asap7: fix ADDER_MAP_FILE default handling with ASAP7_USELVT and ASAP7_USESLVT

oharboe:
Just a quick vain attempt to demonstrate possible fix for #2990

#2990 needs some more investigation as there could be other similar issues in platforms/asap7/config.mk


oharboe:
better to work from issue, this was just a test and a broader solution is needed.

PR #2996, opened 2025-03-24
Re-factored asap7 platform config.mk - more consistent handling of Vt's

jeffng-or:
Re-factored asap7 platform config.mk to use parameterized approach for managing flow variables based on VT choice. Replaced ASAP_USE_LVT and ASAP_USES_SLVT with ASAP_USE_VT, which accepts LVT, SLVT, RVT (defaults to RVT).

### DIFF
--- a/flow/designs/asap7/aes_lvt/config.mk
+++ b/flow/designs/asap7/aes_lvt/config.mk
@@ -14,14 +14,6 @@ export CORE_MARGIN            = 2
 export PLACE_DENSITY          = 0.65
 export TNS_END_PERCENT        = 100
 
-export ASAP7_USELVT           = 1
-export ADDITIONAL_LIBS        = $(LIB_DIR)/asap7sc7p5t_AO_RVT_FF_nldm_211120.lib.gz \
-                                $(LIB_DIR)/asap7sc7p5t_INVBUF_RVT_FF_nldm_220122.lib.gz \
-			        $(LIB_DIR)/asap7sc7p5t_OA_RVT_FF_nldm_211120.lib.gz \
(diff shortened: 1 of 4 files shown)


oharboe (changes requested):
nice!

nit:

ifdef is flaky, relying on the distinction between blank and undefined is asking for trouble w make, env vars, command line options, tcl, etc.

ifeq is better.

jeffng-or:
> nice!
> 
> nit:
> 
> ifdef is flaky, relying on the distinction between blank and undefined is asking for trouble w make, env vars, command line options, tcl, etc.
> 
> ifeq is better.

I just pushed an update to change the CLUSTER_FLOPS check from ifdef to ifeq. Was that the one you were referencing?

oharboe (approved):

maliberty (approved):

maliberty:
Thanks, much nicer!

maliberty:
FYI, adding

Fixes #2996 

to your commit message is nice as when the commit is merged to master it automatically closes the issue.

jeffng-or:
@luarss or @vvbandeira , any idea why ResumeCheck is failing with this PR? Is it the flakiness of the test that Jack mentioned a couple of weeks ago? I just ran it locally and it passed.

luarss:
@jeffng-or Yes - we should disable it temporarily for now.

jeffng-or:
> @jeffng-or Yes - we should disable it temporarily for now.

Thanks, I've updated the PR to include disabling the test for now

jeffng-or:
@vvbandeira , it looks like this PR is approved and ready to go in, but auto merge is disabled. What's the best way to move this one forward? Thanks!

PR #2998, opened 2025-03-24
cva6 mods to support multiple SV front ends

jeffng-or:
Replaced some of the parameterized RAMs with a fake ram. The fake ram isn't perfect since it doesn't have a reset or a byte enable, but hopefully it's semi-close. These mods go through both SV frontends that are being tested.

### DIFF
--- a/flow/designs/asap7/cva6/config.mk
+++ b/flow/designs/asap7/cva6/config.mk
@@ -126,8 +126,8 @@ export VERILOG_FILES          = $(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/core/cvfpu
 	$(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/core/pmp/src/pmp.sv \
 	$(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/core/pmp/src/pmp_entry.sv \
 	$(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/core/pmp/src/pmp_data_if.sv \
+	$(PLATFORM_DIR)/verilog/fakeram7_256x32.sv \
 	$(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/common/local/util/tc_sram_wrapper.sv \
-	$(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/common/local/util/tc_sram_wrapper_cache_techno.sv \
 	$(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/vendor/pulp-platform/tech_cells_generic/src/rtl/tc_sram.sv \
 	$(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/common/local/util/sram.sv \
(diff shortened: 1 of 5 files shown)


flow/designs/src/cva6/common/local/util/fakeram7_256x32.sv
maliberty, line 1: I think this belongs in platforms/asap7/verilog

jeffng-or: moved!

maliberty (approved):

povik:
@widlarizer FYI should you be looking for testcases on long `share` runtime this design triggers it but only if read with Verific. We can provide instructions to reproduce if interested.

widlarizer:
Yes, I would appreciate details, possibly even if it's mitigated by https://github.com/YosysHQ/yosys/pull/4904 - reproducing the originally discovered problematic case also required verific

jeffng-or:
> Yes, I would appreciate details, possibly even if it's mitigated by [YosysHQ/yosys#4904](https://github.com/YosysHQ/yosys/pull/4904) - reproducing the originally discovered problematic case also required verific

Hi Emil,

Use the following flow/scripts/synth_preamble.txt
[synth_preamble.txt](https://github.com/user-attachments/files/19452873/synth_preamble.txt)

The ORFS command line is:

```
make DESIGN_CONFIG=designs/asap7/cva6/config.mk SYNTH_USE_VERIFIC=1 VERILOG_DEFINES=HPDCACHE_ASSERT_OFF
```

The "SHARE pass (SAT-based resource sharing)" step will take around 4h.

widlarizer:
There's sorting failures in `opt_expr` and then `check` takes 3 minutes instead of milliseconds, which is weird. I reproduced the hang in `share` with 0.51, but as of main, `share` finishes in a reasonable amount of time and doesn't dominate the runtime:

```
End of script. Logfile hash: 1fc3187cdf, CPU: user 330.40s system 0.56s, MEM: 752.28 MB peak
Yosys 0.51+41 (git sha1 0c689091e, ccache clang++ 16.0.6 -fPIC -O3 -flto=thin)
Time spent: 39% 2x abc (213 sec), 35% 4x check (193 sec), ...
Elapsed time: 9:06.38[h:]min:sec. CPU time: user 541.56 sys 1.23 (99%). Peak memory: 893388KB.
```

widlarizer:
```
cell $pmux $flatten\csr_regfile_i.$verific$select_1747$/home/emil/pulls/openroad-benchmarks/or-scripts/flow/designs/src/cva6/core/csr_regfile.sv:1744$38441
    parameter \S_WIDTH 238
    parameter \WIDTH 512
...
```
So, the regfile creates a `$pmux` that concatenates 100k bits (many very much like `\csr_regfile_i.rvfi_csr_o.pmpcfg_q[63].locked`, others less descriptively named, there's a 222x repeated sequence of these arrays of structs in there) into its B port. This is what causes `check` to take so long. `share` took so long before the aforementioned PR because there's 53k muxes in this 125k cell design. On the surface level, both cases make sense given the RTLIL, but I might take a look at why the RTLIL is the way it is

PR #3001, opened 2025-03-26
added blackbox annotation to cva6 fakeram

jeffng-or:
Also updated place density and die area in order to complete GRT/DRT.

@povik , I'm able to see the fakerams in the hierarchy browser via the yosys-slang flow.

### DIFF
--- a/flow/platforms/asap7/verilog/fakeram7_256x32.sv
+++ b/flow/platforms/asap7/verilog/fakeram7_256x32.sv
@@ -1,3 +1,4 @@
+(* blackbox *)
 module fakeram7_256x32 (
    output reg [31:0] rd_out,
    input [7:0] addr_in,
--- a/flow/designs/asap7/cva6/config.mk
+++ b/flow/designs/asap7/cva6/config.mk
@@ -179,7 +179,7 @@ export VERILOG_INCLUDE_DIRS = $(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/core/include
 	$(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/core/cvfpu/src/common_cells/include \
 	$(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/core/cache_subsystem/hpdcache/rtl/include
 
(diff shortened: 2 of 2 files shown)


maliberty (approved):

flow/platforms/asap7/verilog/fakeram7_256x32.sv
```diff
@@ -1,3 +1,4 @@
+(* blackbox *)
```
maliberty, line 1: Just curious what this does and why its necessary

povik: Without the blackbox attribute yosys-slang sees an empty module, so it substitutes an empty module for fakeram instances. The reason this file is in the file list in the first place is for Verific, since it doesn't see the fakeram coming from `.lib`, but at the same time it's confusing yosys-slang unless the blackbox attribute is placed on it.

This comes down to the following behavior: Verific interprets an empty module as a blackbox, yosys-slang interprets an empty module as an empty module. yosys-slang has a flag to mimic Verific's behavior though

maliberty: What does yosys do in the same situation with standard Verilog?

povik: It sides with Verific

maliberty: Is there a good reason to be different?

povik: I'm not sure anymore. I thought it's cleaner not to make the assumption but I see it causing confusion. My thinking is you can have an empty module as an edge case not meaning for the module to be a blackbox.

maliberty: I don't think people write empty modules intending to have no implementation.  I think it would be better to have the same defaults as others.

povik: I can think of some fringe situations like programatically-generated Verilog where an empty module is an empty module. I don't like this kind of guessing by a tool but I might need to cave in.

PR #3006, opened 2025-03-28
introduce variables.mk

oharboe:
NOTE! This looks like a giant change, but really it only moves variables in `Makefile` into `scripts/variables.mk`.

Defines an interface to use ORFS scripts + variables implemented by make from non-make build systems or where the user's Makefile is the top makefile.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -84,16 +84,16 @@
 
 # Default design
 DESIGN_CONFIG ?= ./designs/nangate45/gcd/config.mk
-
-# Include design and platform configuration before setting default options
-# in this file. This allows the DESIGN_CONFIG to set different defaults than
-# this file.
+export DESIGN_CONFIG
 include $(DESIGN_CONFIG)
-export DESIGN_NICKNAME?=$(DESIGN_NAME)
+
+export DESIGN_DIR  = $(dir $(DESIGN_CONFIG))
+
 # default value "base" is duplicated from variables.yaml because we need it
-# earlier in the flow for BLOCKS. BLOCKS is a feature specific to Makefile
-# that will not be ported to bazel-orfs.
+# earlier in the flow for BLOCKS. BLOCKS is a feature specific to the
+# ORFS Makefile.
 export FLOW_VARIANT?=base
+# BLOCKS is a ORFS make flow specific feature.
 ifneq ($(BLOCKS),)
   # Normally this comes from variables.yaml, but we need it here to set up these variables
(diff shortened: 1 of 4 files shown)


flow/Makefile
```diff
@@ -84,16 +84,16 @@
...
 DESIGN_CONFIG ?= ./designs/nangate45/gcd/config.mk
-
-# Include design and platform configuration before setting default options
-# in this file. This allows the DESIGN_CONFIG to set different defaults than
-# this file.
+export DESIGN_CONFIG
 include $(DESIGN_CONFIG)
-export DESIGN_NICKNAME?=$(DESIGN_NAME)
+
+export DESIGN_DIR  = $(dir $(DESIGN_CONFIG))
```
maliberty, line 90: What criteria determines which variables stay here and which move to variables.mk?

oharboe: If think we can refine that, but in this first cut if a variable was exclusively used in flow/designs or was in some sense flow/Makefile specific I would leave it in  flow/Makefile.

Also, there are some problems with ordering in make. include statements struggle with variables defined other include files.

Over time, I would want to limit variables.mk to variables documented in variables.yaml

oharboe: Ideas for future refinements would be to convert generate-variables.sh to python, read in variables.yaml and include only vars that are documented.

We could also modify make issue to use variables.yaml + python to read in and create a standalone package. metainformation about variables could be added to .yaml to indicate which are a list of files so that no file will be left behind and we can retire hacks such as ADDITIONAL_FILES.

But... I think this is a step in the right direction. It is more of a policy than code change: ORFS supports other build systems as having control and the ORFS Makefile or a user's makefile or bazel-orfs are just exmaple use-cases that are all first class citizens.

oharboe:
@maliberty CI passed :-)

flow/Makefile
maliberty: There will be derived variables like the log/results/object/reports dirs.  Do you have a though on where they should be?

maliberty (approved):

flow/Makefile
oharboe: I think the simple answer is that anything that is referenced in the .tcl files are in reality part of the interface, so need to be in variables.mk.

I believe variables.mk could be pruned in the future, some variables can be moved back to the Makefile since they are only used by ORFS/Makefile and the test-scripts, but I think some thought needs to go into each variable and that this is a reasonable conservative starting point. Better one too many than too few in variables.mk.

PR #3010, opened 2025-03-28
mods to repro yosys-abc cap_load_unit restriction

jeffng-or:
Modifications to ASAP7 platform to reproduce yosys-abc capacitive_load_unit assertion failure.


jeffng-or:
doesn't need to be a PR

PR #3020, opened 2025-03-29
bazel: use local ORFS files

oharboe:
Another snapshot for further discussion and collaboration. ORFS Bazel support is not under CI and not officially relied upon yet.

It is now possible to modify PDK + make scripts and test them without having to spin a new docker image first. The idea is to gradually wean ORFS Bazel support off the docker image.

To build all ORFS targets for which there are BUILD.bazel files:

```
$ cd flow
$ bazelisk build ...
```

Bring up a design in the GUI:

```
$ bazelisk run //designs/asap7/gcd:gcd_route /tmp/route gui_route
```

image

### DIFF
--- a/flow/BUILD.bazel
+++ b/flow/BUILD.bazel
@@ -1,0 +1,62 @@
+load("@bazel-orfs//:openroad.bzl", "orfs_pdk")
+
+# files shared between scripts/synth.sh and scripts/flow.sh steps
+MAKEFILE_SHARED = [
+    "scripts/*.py",
+    "scripts/*.sh",
+    "scripts/*.yaml",
+    "scripts/*.mk",
+]
+
+# makefile and scripts used by script/synth.sh steps
+filegroup(
+    name = "makefile_yosys",
+    srcs = ["Makefile"],
+    data = glob(MAKEFILE_SHARED + [
+        "scripts/*.script",
+        "scripts/util.tcl",
+        "scripts/synth*.tcl",
+         "platforms/common/**/*.v",
+    ]) +
+    ["//util:makefile_yosys"],
+    visibility = ["//visibility:public"],
+)
+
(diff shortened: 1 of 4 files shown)


oharboe:
@hzeller Thoughts?

maliberty:
How are you bringing up the GUI from a bazel build when we don't compile with the GUI enabled?

oharboe:
> How are you bringing up the GUI from a bazel build when we don't compile with the GUI enabled?

I'm using the openroad binary from the docker image. When spelunking in bazel, you're likely to come across some [patchelf wizardy](https://github.com/The-OpenROAD-Project/bazel-orfs/blob/48cc8783f285d2a274b1613004dcbea21195f49c/docker.bzl#L74) that makes it possible to press binaries into service in this way... 

My plan is to migrate away from using the docker image to using the checked out ORFS and tools/OpenROAD folders as it becomes possible.

maliberty:
I don't have much to say about this.  I can merge it if you want as you are the only user.

oharboe:
> I don't have much to say about this. I can merge it if you want as you are the only user.

One day, hopefully, there will be more users :-)

maliberty (approved):

hzeller:
Drive-by comment: I'd only rarely use `BUILD.bazel`, the canonical name is `BUILD` in bazel projects.

In OR, we use `BUILD.bazel` only in the toplevel as some broken operation systems might get confused with a lowercase build/ directory. In every other place we use `BUILD`.

oharboe:
> Drive-by comment: I'd only rarely use `BUILD.bazel`, the canonical name is `BUILD` in bazel projects.
> 
> In OR, we use `BUILD.bazel` only in the toplevel as some broken operation systems might get confused with a lowercase build/ directory. In every other place we use `BUILD`.

Good to know! @hovind @nekronos FYI

PR #3021, opened 2025-03-29
variables.mk: PLATFORM_DIR can be provided from some non-ORFS location

oharboe:
TL;DR useful for local `bazel build ...` in ORFS when testing PDK changes.

Initial application is to use it in bazel builds in ORFS where local files instead of docker image files are used.

### DIFF
--- a/flow/scripts/variables.mk
+++ b/flow/scripts/variables.mk
@@ -37,14 +37,16 @@ ifeq ($(origin DESIGN_NAME), undefined)
   $(error DESIGN_NAME variable net set.)
 endif
 
-ifneq ($(wildcard $(PLATFORM_HOME)/$(PLATFORM)),)
-  export PLATFORM_DIR = $(PLATFORM_HOME)/$(PLATFORM)
-else ifneq ($(findstring $(PLATFORM),$(PUBLIC)),)
-  export PLATFORM_DIR = ./platforms/$(PLATFORM)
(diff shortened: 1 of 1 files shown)


flow/scripts/variables.mk
```diff
@@ -37,14 +37,16 @@ ifeq ($(origin DESIGN_NAME), undefined)
...
-ifneq ($(wildcard $(PLATFORM_HOME)/$(PLATFORM)),)
-  export PLATFORM_DIR = $(PLATFORM_HOME)/$(PLATFORM)
-else ifneq ($(findstring $(PLATFORM),$(PUBLIC)),)
-  export PLATFORM_DIR = ./platforms/$(PLATFORM)
-else ifneq ($(wildcard ../../$(PLATFORM)),)
-  export PLATFORM_DIR = ../../$(PLATFORM)
-else
-  $(error [ERROR][FLOW] Platform '$(PLATFORM)' not found.)
+ifeq ($(PLATFORM_DIR),)
+  ifneq ($(wildcard $(PLATFORM_HOME)/$(PLATFORM)),)
```
maliberty: You can avoid the extra indentation by just making this the first case with no action:
```
ifeq ($(PLATFORM_DIR),)
# already set
else ifneq ($(wildcard $(PLATFORM_HOME)/$(PLATFORM)),)
...
```

oharboe: Fixed

maliberty (approved):

PR #3043, opened 2025-04-04
Bazel: Build OpenROAD and run a few ORFS flows

oharboe:
Baseline for further development, improvements and discussions.

## Build OpenROAD and run a few ORFS flows with Bazel

Local use case:

- Install Bazelisk and no other dependencies, no need to run `sudo ./setup.sh`
- Modify & build OpenROAD
- Test built OpenROAD with a few ORFS flows

The Bazel support in OpenROAD and ORFS is work in progress and some Bazel experience is recommended before going spelunking in the Bazel builds.

Contributions welcome!

To build `designs/asap7/gcd:gcd_floorplan`:

    cd flow
    (cd ../tools/OpenROAD && bazel build :openroad -c opt) && bazelisk build designs/asap7/gcd:gcd_floorplan

Or to run all flows currently available in Bazel

    cd flow
    (cd ../tools/OpenROAD && bazel build :openroad -c opt) && bazelisk build ...

Note! ORFS uses the OpenROAD Bazel built binary in stop-gap way until OpenROAD has been switched to bzlmod, after which to build all flows becomes simpler as ORFS will build the requisite OpenROAD directly:

    cd flow
    bazelisk build ...

ORFS uses [bazel-orfs](https://github.com/The-OpenROAD-Project/bazel-orfs) to implement the flow and gets some depedencies, like yosys, from the Docker image. Over time, all dependencies should be built with Bazel and the dependency on the ORFS Docker image will be phased out.

### Upgrading MODULE.bazel with the latest bazel-orfs and ORFS Docker image

Run:

    bazelisk run @bazel-orfs//:bump

Then commit MODULE.bazel and MODULE.bazel.lock.


oharboe:
@hzeller @QuantamHD FYI

maliberty:
`Install Bazelisk and no other dependencies` applies to OR but not to yosys or klayout.

maliberty (approved):

oharboe:
> `Install Bazelisk and no other dependencies` applies to OR but not to yosys or klayout.

Actually no. It gets those from the docker image, so there is  dependency on podman or docker.

The documentation will eventually be accurate though: it describes the  end goal.

Either nobody reads it :-), they read it and discover some other dependency and either cares enough to report or fix it or not...

maliberty:
I'm not planning to own converting and maintaining all tools in bazel.  Both provide binary releases which could be used  as an alternative to docker if needed.

oharboe:
> I'm not planning to own converting and maintaining all tools in bazel.  Both provide binary releases which could be used  as an alternative to docker if needed.

👍

That's what I am doing in Bazel now, effectively: using a binary release. 

Bazel idiomatically has ways to use binary releases, a key strength. sometimes with binary patching shenanigans, which to my mind is fine. It works, a bit of an eyesore, but moving on.

I see no point in adding Yosys or Klayout bazel builds in ORFS/OR, binary releases exist.

PR #3044, opened 2025-04-04
floorplan: move metrics code into report_metrics utility fn

oharboe:
Heeds SKIP_REPORT_METRICS=1 now, useful with Bazel testing since report_units_metrics isn't supported in Bazel OpenROAD builds yet.


maliberty:
This will report the units in every step of stage 2.  I think it would be better to check if the command exists (temporarily).

maliberty:
Unneeded due to https://github.com/The-OpenROAD-Project/OpenROAD/pull/7000

PR #3049, opened 2025-04-05
final: save images if GUI is enabled

oharboe:


flow/scripts/final_report.tcl
```diff
@@ -66,6 +66,6 @@ report_cell_usage
 report_metrics 6 "finish"
 
 # Save a final image if openroad is compiled with the gui
-if {[expr [llength [info procs save_image]] > 0]} {
+if {[gui::enabled] && [expr [llength [info procs save_image]] > 0]} {
```
maliberty, line 69: This isn't right as OR in text mode will give gui::enabled=0 but images can still be saved if compiled with the GUI.

oharboe: What solution is needed here then?

maliberty: Doesn't the `info procs` cover this case?

oharboe: No, because it is checking if save_image exists, but it is gui::show that doesnt exist and fails

PR #3052, opened 2025-04-07
Use gcc instead of clang to fix compilation issues

bragadeesh153:


maliberty:
What issues are you fixing?

bragadeesh153:
I'm on ubuntu 22:04. By default when I build locally using clang, it gives the following compilation error. Makes sense to move to gcc as default instead ? 

/build_openroad.sh --local

[INFO FLW-0028] Compiling with 32 threads.
[INFO FLW-0027] Saving logs to build_openroad.log
[INFO FLW-0028] ./build_openroad.sh --local
[INFO FLW-0002] Updating git submodules.
[INFO FLW-0001] Using local build method. This will create binaries at 'tools/install' unless overwritten.
[INFO FLW-0017] Compiling Yosys.
make: Entering directory '/home/bragadeesh/Projects/chip/temp/OpenROAD-flow-scripts/tools/yosys'
[  0%] Building kernel/version_c4b519022.cc
[  0%] Building kernel/driver.o
[  0%] Building techlibs/common/simlib_help.inc
[  0%] Building techlibs/common/simcells_help.inc
[  1%] Building kernel/rtlil.o
[  1%] Building kernel/log.o
[  1%] Building kernel/calc.o
In file included from kernel/driver.cc:20:
In file included from ./kernel/yosys.h:40:
./kernel/yosys_common.h:23:10: fatal error: 'map' file not found
#include <map>
         ^~~~~
In file included from kernel/rtlil.cc:20:
In file included from ./kernel/yosys.h:40:
./kernel/yosys_common.h:23:10: fatal error: 'map' file not found
#include <map>
         ^~~~~

oharboe:
Since we're moving to bazel and bazel is using LLVM, I think it makes more sense to see if this can't be fixed in a way that works on both gcc + llvm.

maliberty:
`'map' file not found` sounds like you have a bad compiler install.  We build with the current setup nightly without issue on this platform.  I'm not open to switching compilers as it is unnecessary and a lot of validation.

PR #3055, opened 2025-04-08
[TESTING] bazel: build OpenROAD and run ORFS flows

oharboe:


oharboe:
@hzeller Thoughts?

hzeller:
Looks good. The final step that shows the size of what is in ~/.cache/bazel is not really needed but a nice FYI while things are still somewhat evolving.
Removing the repository cache I found useful in the OR build as it cleaned up enough storage for the cache compression to work, it was running out of disk before (but that might've also been a bad CI machine I was hitting)

oharboe:
> Looks good. The final step that shows the size of what is in ~/.cache/bazel is not really needed but a nice FYI while things are still somewhat evolving. Removing the repository cache I found useful in the OR build as it cleaned up enough storage for the cache compression to work, it was running out of disk before (but that might've also been a bad CI machine I was hitting)

This is of course a copy and paste + modify of what you did in OpenROAD. I didn't make an decisions or considerations as to what is a good way to do this under github...

oharboe:
@hzeller More disk space needed... `write (No space left on device)`

oharboe:
@maliberty What are the chances of a larger runner, possibly self-hosted on google cloud, runner?

maliberty:
> @maliberty What are the chances of a larger runner, possibly self-hosted on google cloud, runner?

I believe we already are running self-hosted but perhaps we need to scale the machine up.  Most of the heavy lifting is done in Jenkins currently so the self-hosted runners are just doing light work like clang-tidy/format/etc.  Probably we should move this action to Jenkins as well.

@vvbandeira thoughts?

oharboe:
@vvbandeira Will re-open when there is a solution for build capacity/runner.

PR #3070, opened 2025-04-13
[AutoTuner] Restore resume check 

### LINKED ISSUE #3005: Re-enable AutoTuner ResumeCheck tests
### Subject

### Describe the bug

We disabled the AutoTuner ResumeCheck tests due to instability. Adding an issue to track re-enabling the test after addressing the instability.

### Expected Behavior

Test passes consistently

### Environment

```shell
Latest ORFS
```

### To Reproduce

Uncomment the line in test_autotuner.sh and run through CI

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

luarss:
This pull request includes several changes to the `AutoTuner` tests to re-enable previously disabled tests, improve the test setup, and add new utility methods.

Fixes #3005 

Re-enabling tests and improving setup:

* [`flow/test/test_autotuner.sh`](diffhunk://#diff-d56eec4dd48b40016002ea5ffe97f45482b50e63a53f9fef7edb9065883ca489L33-R33): Re-enabled the `test_tune_resume` test in the `AutoTuner` script by uncommenting the test execution line.

Codebase simplification and improvements:

* [`tools/AutoTuner/test/resume_check.py`](diffhunk://#diff-018940b904a40ee28b33e9d60e767ffafe7a8b5dbc291b8cb609667bbac376c2R35): Removed unnecessary directory changes and imported the `glob` module to facilitate file pattern matching. [[1]](diffhunk://#diff-018940b904a40ee28b33e9d60e767ffafe7a8b5dbc291b8cb609667bbac376c2R35) [[2]](diffhunk://#diff-018940b904a40ee28b33e9d60e767ffafe7a8b5dbc291b8cb609667bbac376c2L44-L46)
* [`tools/AutoTuner/test/resume_check.py`](diffhunk://#diff-018940b904a40ee28b33e9d60e767ffafe7a8b5dbc291b8cb609667bbac376c2L90-R127): Introduced the `check_trial_times` method to check the modification times of trial iterations, improving the robustness of the `test_tune_resume` test.

### DIFF
--- a/tools/AutoTuner/test/resume_check.py
+++ b/tools/AutoTuner/test/resume_check.py
@@ -90,9 +90,9 @@ class ResumeCheck(unittest.TestCase):
             for c in options
         ]
 
-    def check_trial_times(self, iteration: int = 0) -> int:
+    def get_trial_times(self, iteration: int = 0) -> int:
         """
-        Checks the nth iteration time of a trial.
+        Returns the nth iteration time of a trial.
 
         :param iteration: The iteration to check.
         :return: The latest modified UNIX time of the nth iteration.
--- a/tools/AutoTuner/test/resume_check.py
+++ b/tools/AutoTuner/test/resume_check.py
@@ -119,7 +119,7 @@ class ResumeCheck(unittest.TestCase):
             time.sleep(30)
             # Check if first config is complete
             while True:
-                cur_modified_time = self.check_trial_times()
+                cur_modified_time = self.get_trial_times()
                 print(f"Current modified time: {cur_modified_time}")
                 print(f"Latest modified time: {latest_modified_time}")
                 if abs(cur_modified_time - latest_modified_time) < 1e-3:
--- a/tools/AutoTuner/test/resume_check.py
+++ b/tools/AutoTuner/test/resume_check.py
@@ -76,9 +76,9 @@ class ResumeCheck(unittest.TestCase):
         # Cast to 1 decimal place
         res_per_trial = float("{:.1f}".format(self.num_cpus / self.samples))
         options = ["", "--resume"]
-        self.exec = AutoTunerTestUtils.get_exec_cmd()
+        self.executable_command = AutoTunerTestUtils.get_exec_cmd()
         self.commands = [
-            f"{self.exec}"
+            f"{self.executable_command}"
             f" --design {self.design}"
             f" --platform {self.platform}"
             f" --config {self.config}"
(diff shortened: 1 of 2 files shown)


tools/AutoTuner/test/resume_check.py
```diff
@@ -65,58 +63,93 @@ class ResumeCheck(unittest.TestCase):
...
+    def check_trial_times(self, iteration: int = 0) -> int:
+        """
+        Checks the nth iteration time of a trial.
+
+        :param iteration: The iteration to check.
+        :return: The latest modified UNIX time of the nth iteration.
+                 If no folders are found, returns a default value of 9e99.
+        """
+        if iteration < 0 or iteration >= self.iterations:
+            raise ValueError("Iteration must be between 0 and iterations - 1")
```
vvbandeira:
```suggestion
            raise ValueError("Iteration must be between 0 and (iterations - 1)")
```

tools/AutoTuner/test/resume_check.py
```diff
@@ -65,58 +63,93 @@ class ResumeCheck(unittest.TestCase):
...
+
+        experiment_dir = os.path.join(
+            cur_dir,
+            f"../../../flow/logs/{self.platform}/{self.design}/{self.experiment_name}-tune",
+        )
+        folders = glob.glob(os.path.join(experiment_dir, f"variant-*-or-{iteration}"))
+        return max((os.path.getmtime(folder) for folder in folders), default=9e99)
```
vvbandeira: This is a little obfuscated:

1. The function's name does not match the expected return value type/format. A check should return True/False.
2. Returning 9e99 is not clear on what is going on.
3. Creating a folder does not confirm the run status. For this test to be true to its purpose, we need to guarantee that it has not finished, not just started.

We should consider using a `get_experiment_status` function to check if all iterations have finished running; this function would return at least "RUNNING" and "FINISHED", other states might be helpful but are not required now.

vvbandeira (changes requested):

tools/AutoTuner/test/resume_check.py
luarss: 1. The function's name does not match the expected return value type/format. A check should return True/False.
> Can possibly change it to `get_trial_times`
2. Returning 9e99 is not clear on what is going on.
> It is just a dummy value, to compare for latest modified time in while loop line 121-128

```python
            # Check if first config is complete
            while True:
                cur_modified_time = self.check_trial_times()
                print(f"Current modified time: {cur_modified_time}")
                print(f"Latest modified time: {latest_modified_time}")
                if abs(cur_modified_time - latest_modified_time) < 1e-3:
                    break
                latest_modified_time = cur_modified_time
                time.sleep(10)
```

3. Creating a folder does not confirm the run status. For this test to be true to its purpose, we need to guarantee that it has not finished, not just started.
> This function returns the latest modified time of a given iteration (folder names are matched using `iteration` glob) - so if a run is completed the folder should no longer be modified.


4, We should consider using a get_experiment_status function to check if all iterations have finished running; this function would return at least "RUNNING" and "FINISHED", other states might be helpful but are not required now.
> `get_experiment_status` is helpful in general, but might not be too useful in resume_check because we need to check iteration completion, as opposed to experiment completion (or all_iterations completion)

PR #2939, opened 2025-03-14
Enabled cva6 on asap7 using yosys-slang

jeffng-or:
Enabling snapshot of cva6 on asap7. yosys-slang was used as the front end.

Metrics:
score: 141368.0
clk_period: 17.0
worst_slack: -1396.68
effective_clk_period: 1413.68
num_drc: 0

cva6_200

TODOs as part of yosys-slang enablement: https://github.com/povik/yosys-slang/issues/74

PR #3076, opened 2025-04-14
Integrate yosys-slang

povik:
Add yosys-slang as a submodule and provide `SYNTH_USE_SLANG` option.

Changes from #2875 (now superseded):

 * new `VERILOG_DEFINES` variable piped into both yosys-slang and Yosys frontend
 * split `RTLIL_FILE` from `VERILOG_FILES`
 * updated yosys-slang with build system changes
 * convert all Ibex configs

I've tested with @jeffng-or's cva6 config from #2939


povik:
Marking this as do not merge pending the resolution of https://github.com/povik/yosys-slang/issues/120 which breaks `build_openroad.sh --local`

EDIT: resolved

oharboe:
non-sequitor: 

Longer term I'd like to see builds moved to bazel where there's no need to add submodules, just specify version and URLs + any patches of dependencies in MODULE.bazel.

flow/scripts/synth.tcl
```diff
@@ -1,4 +1,4 @@
-set ::env(VERILOG_FILES) $::env(RESULTS_DIR)/1_synth.rtlil
+set ::env(RTLIL_FILE) $::env(RESULTS_DIR)/1_synth.rtlil
```
oharboe: Switch to proc w/arguments instead of passing information via RTLIL_FILE globals?

flow/scripts/variables.yaml
```diff
@@ -169,6 +169,9 @@ SYNTH_MEMORY_MAX_BITS:
   default: 4096
   stages:
     - synth
+SYNTH_USE_SLANG:
+  description: >
+    Use yosys-slang for read-in of SystemVerilog design.
```
oharboe: add `default: 0`, makes it clearer that this is a 1 or 0 option.

flow/scripts/variables.yaml
```diff
@@ -569,8 +572,7 @@ DESIGN_NAME:
 VERILOG_FILES:
   required: true
   description: >
-    The path to the design Verilog files or JSON files providing a description
-    of modules (check `yosys -h write_json` for more details).
+    The path to the design Verilog files providing a description of modules.
```
oharboe: SystemVerilog or Verilog

flow/designs/asap7/ibex/config.mk
```diff
@@ -3,7 +3,32 @@ export PLATFORM               = asap7
 export DESIGN_NICKNAME        = ibex
 export DESIGN_NAME            = ibex_core
 
-export VERILOG_FILES         = $(sort $(wildcard $(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/*.v))
+export VERILOG_FILES = $(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/ibex_pkg.sv \
```
oharboe: any reason not to use a glob pattern here and have more terse code?

Makefiles will never have anything but the crudest dependency notions by 2025 standards.

oharboe: DESIGN_DIR is terser

flow/scripts/variables.yaml
oharboe: add

  stages:
    - synth

povik:
Because we've reordered commands in the synthesis script, the change impacts non-slang designs too. `mock-array` fails with the following. Should I package a reproducer?

```
asap7 mock-array (base)
 Last log file 5_1_grt.log
 Found 2 errors in the logs.
     [ERROR DRT-0073] No access point for ces_1_0/io_ins_left[10].
     [ERROR DRT-0073] No access point for ces_0_0/io_ins_left[10].
```

oharboe:
> Because we've reordered commands in the synthesis script, the change impacts non-slang designs too. `mock-array` fails with the following. Should I package a reproducer?
> 
> ```
> asap7 mock-array (base)
>  Last log file 5_1_grt.log
>  Found 2 errors in the logs.
>      [ERROR DRT-0073] No access point for ces_1_0/io_ins_left[10].
>      [ERROR DRT-0073] No access point for ces_0_0/io_ins_left[10].
> ```

My thinking is that this is already reported https://github.com/The-OpenROAD-Project/OpenROAD/issues/6991

povik:
Thanks, looks to be the same issue

flow/designs/asap7/ibex/config.mk
povik: Not really

oharboe (changes requested):
Not too many comments...

How about leaving src/ibex unmodified and adding a src/ibex_sv?

Those really are different test cases and it could be interesting to do an A/B comparison in the future.

povik:
What's there to test? I don't want to support sv2v processed sources when I'm already aware of the drawbacks of that method, and it's the motivation behind yosys-slang existence. I'll let @maliberty make the call.

oharboe:
I see. I would consider using a new folder name as it is not a new version of the same, but something completely new.

povik:
To report here's how Ibex QoR is impacted by the switch (base: sv2v+read_verilog/test: yosys-slang)

asap7
Screenshot 2025-04-15 at 11 30 11

ihp-sg13g2
Screenshot 2025-04-15 at 11 31 31

nangate45
Screenshot 2025-04-15 at 11 32 02

gf180
Screenshot 2025-04-15 at 11 30 51

sky130hd
Screenshot 2025-04-15 at 11 32 30

sky130hs
Screenshot 2025-04-15 at 11 34 33

povik:
> I would consider using a new folder name as it is not a new version of the same, but something completely new.

It should be the same design in the same version with a changed flow (instead of sv2v+read_verilog we do read_slang). In any case we can put the sources into a new folder.

jeffng-or:
@povik, we might have talked about this a while back, but I can't remember. But, can we change the flow variable SYNTH_USE_SLANG to something that could be used for both yosys-slang and Verific? In other words, something like SYNTH_SV_FRONTEND which could be yosys-slang or verific (can default to yosys-slang).

povik:
@jeffng-or Definitely. Let me make the change

flow/scripts/variables.yaml
```diff
@@ -169,6 +169,13 @@ SYNTH_MEMORY_MAX_BITS:
   default: 4096
   stages:
     - synth
+SYNTH_HDL_FRONTEND:
```
povik, line 172: @jeffng-or See here. I went with "HDL frontend" since we might start supporting VHDL through Verific or GHDL

maliberty (approved):

maliberty:
> I see. I would consider using a new folder name as it is not a new version of the same, but something completely new.

Is this truly something new or just the same thing in sv instead of v?  I won't hold up this PR but I am unclear if this rename was needed.

povik:
Hooray! Now we can enable asap7/cva6 in the CI. I'm not sure about the steps required.

maliberty:
What is the runtime for asap7/cva6?

povik:
Jeff reported 50 minutes for the full flow

jeffng-or:
Nice! A few post-merge notes:

1. I noticed that if you build_openroad.sh twice, the following slang files get re-compiled:

```
[ 94%] Building CXX object src/CMakeFiles/yosys-slang.dir/abort_helpers.cc.o
[ 94%] Building CXX object src/CMakeFiles/yosys-slang.dir/async_pattern.cc.o
[ 95%] Building CXX object src/CMakeFiles/yosys-slang.dir/blackboxes.cc.o
[ 96%] Building CXX object src/CMakeFiles/yosys-slang.dir/builder.cc.o
[ 97%] Building CXX object src/CMakeFiles/yosys-slang.dir/diag.cc.o
[ 98%] Building CXX object src/CMakeFiles/yosys-slang.dir/naming.cc.o
[ 99%] Building CXX object src/CMakeFiles/yosys-slang.dir/slang_frontend.cc.o
```

2. nangate45/mempool_group run time is ~1h45m
3. asap7/cva6 run time is ~54m
4. I'll work with the DevOps team to get both enabled in CI

povik:
> 1. I noticed that if you build_openroad.sh twice, the following slang files get re-compiled:

I confirm seeing the same on my end. I'll take a look.

povik:
The cause is Yosys headers getting updated so yosys-slang sees it needs to recompile all the `.cc` which depend on them. This should be fixable inside Yosys by adjusting the install script.

PR #3077, opened 2025-04-15
Add RULES_JSON and makefile rules more consistent with the do-something pattern

oharboe:
This enables rules-base.json files to be used from bazel-orfs tests.

### DIFF
--- a/flow/scripts/generate-variables-docs.py
+++ b/flow/scripts/generate-variables-docs.py
@@ -17,10 +17,10 @@ with open(yaml_path, "r") as file:
 preferred_order = ["synth", "floorplan", "place", "cts", "grt", "route", "final"]
 stages = {stage for value in data.values() for stage in value.get("stages", [])}
 # convert set of stages to stages in a list in the preferred order, but
-# list all stages
-stages = [stage for stage in preferred_order if stage in stages] + [
-    stage for stage in stages if stage not in preferred_order
-]
+# list all stages and sort the rest for a stable order
+stages = [stage for stage in preferred_order if stage in stages] + sorted(
+    [stage for stage in stages if stage not in preferred_order]
+)
 
 markdown_table = ""
 
--- a/flow/util/genRuleFile.py
+++ b/flow/util/genRuleFile.py
@@ -7,6 +7,7 @@ from re import sub
 import argparse
 import json
 import operator
+import os
 import sys
 
 
@@ -20,6 +21,7 @@ def gen_rule_file(
     metrics_file=None,
     metrics_to_consider=[],
 ):
+    print(f"{os.path.normpath(rules_file)} updates:")
 
     with open(metrics_file, "r") as f:
         metrics = json.load(f)
--- a/flow/util/genRuleFile.py
+++ b/flow/util/genRuleFile.py
@@ -355,7 +355,7 @@ if __name__ == "__main__":
         "--new-rules",
         type=str,
         default=None,
-        help="Rules input file.",
+        help="Rules output file.",
     )
     parser.add_argument(
         "-m",
--- a/flow/util/genRuleFile.py
+++ b/flow/util/genRuleFile.py
@@ -11,36 +11,28 @@ import sys
 
 
 def gen_rule_file(
-    design_dir,
+    rules_file,
+    new_rules_file,
     update,
     tighten,
     failing,
     variant,
-    metrics=None,
+    metrics_file=None,
     metrics_to_consider=[],
 ):
 
-    golden_metrics = f"metadata-{variant}-ok.json"
-    if isinstance(metrics, str) and isfile(metrics):
-        with open(metrics, "r") as f:
-            metrics = json.load(f)
-    elif isfile(golden_metrics):
-        with open(golden_metrics, "r") as f:
-            metrics = json.load(f)
+    with open(metrics_file, "r") as f:
+        metrics = json.load(f)
     if not isinstance(metrics, dict):
-        print(f"[ERROR] Invalid format for reference metrics {design_dir}")
+        print(f"[ERROR] Invalid format for reference metrics {metrics_file}")
         sys.exit(1)
 
-    original_directory = getcwd()
-    chdir(design_dir)
-    rules_file = f"rules-{variant}.json"
     rules = dict()
-
     if isfile(rules_file):
         with open(rules_file, "r") as f:
             OLD_RULES = json.load(f)
     else:
-        print(f"[WARNING] Rules file not found {design_dir}")
+        print(f"[WARNING] No old rules file found {rules_file}")
         OLD_RULES = None
 
     # dict format
@@ -189,10 +181,7 @@ def gen_rule_file(
     change_str = ""
     for field, option in rules_dict.items():
         if field not in metrics.keys():
-            print(
-                f"[ERROR] Metric {field} not found in "
-                f"metrics file: {metrics_file} or golden metrics."
-            )
+            print(f"[ERROR] Metric {field} not found")
             sys.exit(1)
 
         if isinstance(metrics[field], str):
@@ -305,15 +294,12 @@ def gen_rule_file(
         rules[field] = dict(value=rule_value, compare=option["compare"])
 
     if len(change_str) > 0:
-        print(design_dir)
         print(format_str.format("Metric", "Old", "New", "Type"), end="")
         print(format_str.format("------", "---", "---", "----"), end="")
         print(change_str)
 
-    with open(rules_file, "w") as f:
+    with open(new_rules_file, "w") as f:
         json.dump(rules, f, indent=4)
-
-    chdir(original_directory)
 
 
 def comma_separated_list(value):
@@ -326,7 +312,6 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser(
         description="Generates or updates rules file for CI."
     )
-    parser.add_argument("dir", help="Path to the design directory.")
     parser.add_argument(
         "-v", "--variant", default="base", help='Flow variant [default="base"].'
     )
@@ -359,6 +344,18 @@ if __name__ == "__main__":
         help="Reference metadata file.",
     )
     parser.add_argument(
+        "--rules",
+        type=str,
+        default=None,
+        help="Rules input file.",
+    )
+    parser.add_argument(
+        "--new-rules",
+        type=str,
+        default=None,
+        help="Rules input file.",
+    )
+    parser.add_argument(
         "-m",
         "--metrics",
         type=comma_separated_list,
@@ -376,7 +373,8 @@ if __name__ == "__main__":
         sys.exit(1)
 
     gen_rule_file(
-        args.dir,
+        args.rules,
+        args.new_rules,
         args.update,
         args.tighten,
         args.failing,
--- a/flow/util/utils.mk
+++ b/flow/util/utils.mk
@@ -43,7 +43,10 @@ update_rules:
 
 .PHONY: update_rules_force
 update_rules_force:
-	$(UTILS_DIR)/genRuleFile.py $(DESIGN_DIR) --variant $(FLOW_VARIANT) --update
+	$(UTILS_DIR)/genRuleFile.py $(DESIGN_DIR) \
+	    --reference $(REPORTS_DIR)/metadata.json \
+		--variant $(FLOW_VARIANT) \
+		--update
 
 .PHONY: update_metadata_autotuner
 update_metadata_autotuner:
--- a/flow/util/utils.mk
+++ b/flow/util/utils.mk
@@ -33,24 +33,37 @@ update_metadata:
 	cp -f $(REPORTS_DIR)/metadata.json \
 	      $(DESIGN_DIR)/metadata-$(FLOW_VARIANT)-ok.json
 
-.PHONY: update_rules
-update_rules:
+.PHONY: do-update_rules
+do-update_rules:
 	$(UTILS_DIR)/genRuleFile.py \
 	    --rules $(DESIGN_DIR)/rules-$(FLOW_VARIANT).json \
-	    --new-rules $(DESIGN_DIR)/rules-$(FLOW_VARIANT).json \
+	    --new-rules $(RESULTS_DIR)/rules-$(FLOW_VARIANT).json \
 		--reference $(REPORTS_DIR)/metadata.json \
 		--variant $(FLOW_VARIANT) \
 		--failing \
 		--tighten
 
-.PHONY: update_rules_force
-update_rules_force:
+.PHONY: do-copy_update_rules
+do-copy_update_rules:
+	cp -f $(RESULTS_DIR)/rules-$(FLOW_VARIANT).json \
+	      $(DESIGN_DIR)/rules-$(FLOW_VARIANT).json
+
+.PHONY: update_rules
+update_rules: do-update_rules do-copy_update_rules
+
+.PHONY: do-update_rules_force
+do-update_rules_force:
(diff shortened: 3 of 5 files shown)


oharboe:
@maliberty I'm thinking unrelated metrics baseline update failure

image

oharboe:
@vvbandeira The latest commit of this PR is https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3077/commits/6b0483407443893ecf14f9f3d99b271f1a342681 but the job checked out 5113d57d0c68b0e999fe96d44313289975050a9e
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/actions/runs/14465282230/job/40566142916?pr=3077

vvbandeira:
> @vvbandeira The latest commit of this PR is [6b04834](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/6b0483407443893ecf14f9f3d99b271f1a342681) but the job checked out [5113d57](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/5113d57d0c68b0e999fe96d44313289975050a9e) https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/actions/runs/14465282230/job/40566142916?pr=3077

The action is running like the `pr-merge` in Jenkins. The action gets your changes from this PR and makes a merge commit into the master before running the tests. If you look into the commit hash https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/5113d57d0c68b0e999fe96d44313289975050a9e you get:

image

And that is the latest master:
image

flow/util/genRuleFile.py
```diff
@@ -358,6 +343,18 @@ def comma_separated_list(value):
...
+        "--rules",
+        type=str,
+        default=None,
+        help="Rules input file.",
+    )
+    parser.add_argument(
+        "--new-rules",
+        type=str,
+        default=None,
+        help="Rules input file.",
```
vvbandeira: This would be the output, right?

flow/util/genRuleFile.py
```diff
@@ -189,10 +181,7 @@ def gen_rule_file(
     change_str = ""
     for field, option in rules_dict.items():
         if field not in metrics.keys():
-            print(
-                f"[ERROR] Metric {field} not found in "
-                f"metrics file: {metrics_file} or golden metrics."
-            )
+            print(f"[ERROR] Metric {field} not found")
```
vvbandeira, line 186: Printing the name of the file is helpful because we have utilities in the CI that run this in parallel.

vvbandeira (commented):
Just two minor nits.
@maliberty: when the PR is ready, please leave it to me to merge it. I need to change the UpdateRules label action as the signature of `gen_rule_file` is changing.

flow/util/genRuleFile.py
oharboe: yes: I have tried to be consistent with the "old" vs "new" terminology in the code. However, I was thinking that perhaps later, I'd prefer to have an input/output terminology which is less surprising than the "new" "old" terminology in this file.

flow/util/genRuleFile.py
oharboe: Can you explain a bit more?

This would be inconsistent with all other error messages in OpenROAD/ORFS.

Error and progress messages do not contain context in which they run, because it doesn't scale.

Rather the CI system has to present log files in a way that only a single job and context is shown and that output from multiple jobs are not intermingled.

If we are to address this, then I would:

1. switch to exceptions (context and stack trace)
2. add a finally statement that prints out some extra information, like file name in the event of an exception

However, I'm unsure this is a good idea. A good CI system would give adequate context that it would be obvious what is being processed and the extra logging cruft adds a maintenance burden and clutters the code.

oharboe:
> The action is running like the `pr-merge` in Jenkins.

I still don't understand. rules-json is a fast forward on top of master, even as of writing, so why is a new commit created and why is there an error?

vvbandeira:
> > The action is running like the `pr-merge` in Jenkins.
> 
> I still don't understand. rules-json is a fast forward on top of master, even as of writing, so why is a new commit created and why is there an error?

We don't allow fast-forward merges in this repo. Thus, GH creates the merge commit for testing.

I don't know the error, but I will need to look.

oharboe:
> > > The action is running like the `pr-merge` in Jenkins.
> > 
> > 
> > I still don't understand. rules-json is a fast forward on top of master, even as of writing, so why is a new commit created and why is there an error?
> 
> We don't allow fast-forward merges in this repo. Thus, GH creates the merge commit for testing.

I see.

flow/scripts/generate-variables-docs.py
```diff
@@ -17,10 +17,10 @@
...
 stages = {stage for value in data.values() for stage in value.get("stages", [])}
 # convert set of stages to stages in a list in the preferred order, but
-# list all stages
-stages = [stage for stage in preferred_order if stage in stages] + [
-    stage for stage in stages if stage not in preferred_order
-]
+# list all stages and sort the rest for a stable order
+stages = [stage for stage in preferred_order if stage in stages] + sorted(
+    [stage for stage in stages if stage not in preferred_order]
+)
```
oharboe, line 23: @vvbandeira I threw in a sorted() here to get a stable order, belt and suspenders. The order is a result of reading in the .yaml and the implementation details of set() in python, so the order is the result of a long line of assumptions and implementation details that I can't account for. A sorted() here puts my mind to ease and may even be necessary.

oharboe:
@vvbandeira Purely for my planning purposes: how long should I expect merge to take after review & fixes is complete?

I have various other pull requests in flight that will depend on ORFS docker images for them to be easy for collaborators to test(otherwise it requires much mucking about locally to test a combination of pull requests).

vvbandeira:
> @vvbandeira Purely for my planning purposes: how long should I expect merge to take after review & fixes is complete?
> 
> I have various other pull requests in flight that will depend on ORFS docker images for them to be easy for collaborators to test(otherwise it requires much mucking about locally to test a combination of pull requests).

Not long. Making the changes is quick, maybe a couple of hours just to run the integration tests.

oharboe:
> > @vvbandeira Purely for my planning purposes: how long should I expect merge to take after review & fixes is complete?
> > I have various other pull requests in flight that will depend on ORFS docker images for them to be easy for collaborators to test(otherwise it requires much mucking about locally to test a combination of pull requests).
> 
> Not long. Making the changes is quick, maybe a couple of hours just to run the integration tests.

:+1:

flow/util/genRuleFile.py
```diff
@@ -305,16 +294,13 @@ def gen_rule_file(
         rules[field] = dict(value=rule_value, compare=option["compare"])
 
     if len(change_str) > 0:
-        print(design_dir)
```
vvbandeira, line 308: Some users copy and paste the generated table into the PRs descriptions; if the platform/design tuple is present, it saves typing and helps prevent errors.

oharboe: Yes, but this comes at the cost of loss of abstraction and increased maintenance.

The design_dir is a concept that is specific to the layout of the test scripts in ORFS, this concept should not leak into genRuleFile.py, it is not a restriction/behavior of ORFS itself.

Reporting the context in which a utility failed, should not mean that a utility should become context specific.

oharboe: It is hard to give specific advice and come up with a good solution without an example.

Can you show an example in CI / bug report where a user would use this log?

vvbandeira:
```
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2802
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2872
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2896
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2975
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2981
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3038
```

oharboe: Do I have access to the tool that generates the reports above?

oharboe: This should restore the use-case without genRuleFile.py being aware of DESIGN_DIR, it will now work for use-cases where the ORFS test design folder layout is not used: private repositories using ORFS or bazel-orfs.

New output

```
$ make update_rules
/home/oyvind/OpenROAD-flow-scripts/flow/util/genRuleFile.py \
    --rules ./designs/nangate45/gcd//rules-base.json \
    --new-rules ./reports/nangate45/gcd/base/rules.json \
	--reference ./reports/nangate45/gcd/base/metadata.json \
	--variant base \
	--failing \
	--tighten
designs/nangate45/gcd/rules-base.json updates:
| finish__timing__wns_percent_delay             |    -25.2 |   -25.05 | Tighten  |

cp -f ./reports/nangate45/gcd/base/rules.json \
      ./designs/nangate45/gcd//rules-base.json
```

If you copy and paste, it will appear as:


designs/nangate45/gcd/rules-base.json updates:
| finish__timing__wns_percent_delay             |    -25.2 |   -25.05 | Tighten  |

flow/util/genRuleFile.py
```diff
@@ -358,6 +345,18 @@ def comma_separated_list(value):
...
+        "--rules",
+        type=str,
+        default=None,
+        help="Rules input file.",
+    )
+    parser.add_argument(
+        "--new-rules",
+        type=str,
+        default=None,
+        help="Rules input file.",
```
maliberty: copy-paste bug

flow/util/genRuleFile.py
maliberty: I think it reasonable to include the design name, something like "[ERROR] Metric {field} not found for nangate45/gcd"

flow/util/genRuleFile.py
maliberty: Similarly the design name is useful even if a specific path is not

oharboe: @vvbandeira Does this fix the use-case?

flow/util/utils.mk
```diff
@@ -1,42 +1,73 @@
...
+metadata-generate:
+	@mkdir -p $(REPORTS_DIR)
 	@echo $(DESIGN_DIR) > $(REPORTS_DIR)/design-dir.txt
 	@$(UTILS_DIR)/genMetrics.py -d $(DESIGN_NICKNAME) \
 		-p $(PLATFORM) \
 		-v $(FLOW_VARIANT) \
-		-o $(REPORTS_DIR)/metadata-$(FLOW_VARIANT).json 2>&1 \
-		| tee $(REPORTS_DIR)/gen-metrics-$(FLOW_VARIANT)-check.log
+		-o $(REPORTS_DIR)/metadata.json 2>&1 \
+		| tee $(abspath $(REPORTS_DIR)/metadata-generate.log)
```
maliberty: Why is abspath needed here and below?

flow/util/utils.mk
```diff
@@ -1,42 +1,73 @@
...
-	$(UTILS_DIR)/genRuleFile.py $(DESIGN_DIR) \
-		--reference $(REPORTS_DIR)/metadata-$(FLOW_VARIANT).json \
+.PHONY: do-update_rules
+do-update_rules:
+	@mkdir -p $(REPORTS_DIR)
+	$(UTILS_DIR)/genRuleFile.py \
+	    --rules $(RULES_JSON) \
+	    --new-rules $(REPORTS_DIR)/rules.json \
+		--reference $(REPORTS_DIR)/metadata.json \
 		--variant $(FLOW_VARIANT) \
```
maliberty: Indentation alignment

flow/util/utils.mk
```diff
@@ -1,42 +1,73 @@
...
+
+.PHONY: do-update_rules_force
+do-update_rules_force:
+	@mkdir -p $(REPORTS_DIR)
+	$(UTILS_DIR)/genRuleFile.py \
+	    --rules $(RULES_JSON) \
+	    --new-rules $(REPORTS_DIR)/rules.json \
+	    --reference $(REPORTS_DIR)/metadata.json \
+		--variant $(FLOW_VARIANT) \
+		--update
```
maliberty: alignment

maliberty (changes requested):

flow/util/utils.mk
oharboe: This is consistent with Makefile. It is useful when doing `ps -Af | grep tee` in a system/server/bazel when there are many processes to `tail -f` and stare at logs. This runs so quickly that it was more a consistency reflex here.

flow/util/genRuleFile.py
oharboe: fixed

flow/util/utils.mk
oharboe: tab vs. spaces, not visible in editor. Don't know why I got tabs. Fixed.

flow/util/utils.mk
oharboe: fixed

flow/util/genRuleFile.py
oharboe: We don't have the design name, we only have input and output .json rule files.

Also, the command line is logged before the output, which gives context.

maliberty (approved):
Over to @vvbandeira to merge

oharboe:
@maliberty @vvbandeira 

Take a new look at the updated output. I think it covers the use cases that @vvbandeira showed me, has minimal complexity and testing requirements on the code and is pretty consistent with what to expect in other tools and ORFS generally.

1. First the command line is logged. This gives accurate context. What information is passed to the genRuleFile.py is very clear from the command line.
2. Then some information about which rule file is being updated is logged. This is superfluous as it is covered in 1, but as @vvbandeira points out it is useful in the github copy & paste use-case. The rules-base.json file is accurate about what is being updated and the information about which design this is is clear from the path for the ORFS use-case.
3. If we want more information, then that can come from `metadata.json`, just let me know what you want, it contains everything. However, I think the path to rules-base.json says it all(at least compared to what was previously).
4. I moved the `designs/nangate45/gcd/rules-base.json updates:` logging statement earlier in the code compared to before. This has the advantage that warnings during processing will fall under the correct `designs/nangate45/gcd/rules-base.json updates:` header.

```
$ make update_rules
/home/oyvind/OpenROAD-flow-scripts/flow/util/genRuleFile.py \
    --rules ./designs/nangate45/gcd//rules-base.json \
    --new-rules ./reports/nangate45/gcd/base/rules.json \
    --reference ./reports/nangate45/gcd/base/metadata.json \
    --variant base \
    --failing \
    --tighten
designs/nangate45/gcd/rules-base.json updates:
| finish__timing__wns_percent_delay             |    -25.2 |   -25.05 | Tighten  |

cp -f ./reports/nangate45/gcd/base/rules.json \
      ./designs/nangate45/gcd//rules-base.json
```

If you copy and paste, it will appear as:


designs/nangate45/gcd/rules-base.json updates:
**any warnings would appear here**
| finish__timing__wns_percent_delay             |    -25.2 |   -25.05 | Tighten  |

oharboe:
@vvbandeira pr-merge worked and pr-head failed. This is an instability on master as pr-head and pr-merge was identical at the time of this PR ran:

microwatt sky130hd:

```
08:53:59  [ERROR] detailedroute__antenna__violating__nets fail test: 5.0 <= 1.0
```

oharboe:
@vvbandeira ETA for my merge on this one for my planning purposes?

I have followon work hanging on this PR.

vvbandeira (approved):

PR #3081, opened 2025-04-16
flow: platforms: ihp-sg13g2: Add GDS_ALLOW_EMPTY

dnltz:
IHP's SRAM macros have empty placeholder cells included. ORFS will fail in the final step when merging all GDS files, because there are empty cells. According to IHP, there is nothing expected to be inside. So, add a regex to ignore excatly these empty cells.

### DIFF
--- a/flow/platforms/ihp-sg13g2/config.mk
+++ b/flow/platforms/ihp-sg13g2/config.mk
@@ -139,3 +139,11 @@ export CDL_FILE ?= $(PLATFORM_DIR)/cdl/sg13g2_stdcell.cdl
 
 #Temporary: skip post-DRT repair antennas
 export SKIP_ANTENNA_REPAIR_POST_DRT = 1
+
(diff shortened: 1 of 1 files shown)


dnltz:
@sergeiandreyev @KrzysztofHerman fyi

maliberty:
I'm confused as wouldn't you need real implementations of these cells to tapeout or does IHP substitute them post-tapeout?  Usually this mechanism is used for non-manufacturable cells (eg fake rams or academic pdks).

sergeiandreyev:
Hi @maliberty, these empty/dummy layouts are part of the output of our internal SRAM generator, the development of which is not supported currently, so we should stick with what we have, unfortunately. From what I checked internally, there were no complaints nor open bugs on this.
We can come back to this issue when there will be resources for its active development or a decision to move to another SRAM generator (open-source?).

maliberty (approved):

PR #3082, opened 2025-04-16
def2stream: log cleanup, terser and removed debug progress statements

oharboe:
The code was distinguishing between empty and not set GDS_ALLOW_EMPTY which led to confusing logs in the common case.

New output:

```
[INFO] GDS_ALLOW_EMPTY=Element
[INFO] All LEF cells have matching GDS/OAS cells
[INFO] No orphan cells in the final layout
```

### DIFF
--- a/flow/util/def2stream.py
+++ b/flow/util/def2stream.py
@@ -20,7 +20,6 @@ print("[INFO] Reporting cells prior to loading DEF ...")
 for i in main_layout.each_cell():
     print("[INFO] '{0}'".format(i.name))
 
-print("[INFO] Reading DEF ...")
 main_layout.read(in_def, layoutOptions)
 
 # Clear cells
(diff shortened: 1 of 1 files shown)


flow/util/def2stream.py
```diff
@@ -20,43 +20,40 @@
...
+regex = re.compile(allow_empty) if allow_empty else None
+
+if allow_empty:
+    print(f"[INFO] GDS_ALLOW_EMPTY={allow_empty}")
+
 for i in top_only_layout.each_cell():
     if i.is_empty():
         missing_cell = True
-        if regex is not None and re.match(regex, i.name):
+        if regex is not None and regex.match(i.name):
```
oharboe, line 56: Driveby fix: move regex compilation out of loop, don't think it matters, I have never seen def2stream take any time.

flow/util/def2stream.py
```diff
@@ -20,43 +20,40 @@
...
 top.copy_tree(main_layout.cell(design_name))
 
-print("[INFO] Checking for missing cell from GDS/OAS...")
 missing_cell = False
-regex = None
-if "GDS_ALLOW_EMPTY" in os.environ:
-    print("[INFO] Found GDS_ALLOW_EMPTY variable.")
-    regex = os.getenv("GDS_ALLOW_EMPTY")
+allow_empty = os.environ.get("GDS_ALLOW_EMPTY", "")
+regex = re.compile(allow_empty) if allow_empty else None
```
oharboe, line 48: stop distinguishing between env var set and empty env var

oharboe:
@maliberty Unrelated error on master:

```
sky130hd/microwatt
10:33:47  [ERROR] detailedroute__antenna__violating__nets fail test: 5.0 <= 0.0
```

flow/util/def2stream.py
```diff
@@ -101,7 +95,6 @@
             top.insert(pya.CellInstArray(cell.cell_index(), pya.Trans()))
 
 # Write out the GDS
-print("[INFO] Writing out GDS/OAS '{0}'".format(out_file))
```
maliberty, line 104: I think it is helpful to tell what file was written.

oharboe: This information is logged as the command line is logged.

This is consistent with e.g. cp, gcc, etc. not to repeat information tersely and accurately described by logging the command.

I dont mind making the change to the OR, but I think there is something to be said for the approach of cp, gcc, etc to log the command line and not add this logging.

maliberty (approved):

PR #3090, opened 2025-04-17
generate-docs: stable order

oharboe:
an ordering was dependent on set of dictionary keys from yaml parsing of a dictionary.

We want a stable ordering and that is a lot of
implementation detail to base that stable ordering feature on, so sort to make it clear in the code
that we have a stable order.


maliberty (approved):

oharboe:
@maliberty Merge? Pesky microwatt sky130hd rules-base.json problems on master.

maliberty:
FYI, the public CI is green this morning so merging master should solve failures on any older PRs

PR #3098, opened 2025-04-21
Obtain bash path through /usr/bin/env in flow scripts.

mminutoli:
Unfortunately, not all systems have bash in /bin. One notable example is nixos. When entering a development shell in nixos, the current version of the scripts will fail complaining that "something is wrong at line 1". The issue is that there is no /bin/bash on NixOS or more generally on the system.

### DIFF
--- a/flow/platforms/sky130ram/Makefile
+++ b/flow/platforms/sky130ram/Makefile
@@ -66,7 +66,7 @@ endif
 # Verbosity
 OPTS += -v
 
-SHELL ?= /bin/bash
+SHELL ?= /usr/bin/env bash
 export SHELL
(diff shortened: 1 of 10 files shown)


maliberty:
Thanks for contributing.  You need to sign your commits with `git commit -s`.  Please see the DCO link above for more info.

mminutoli:
I notice that my patch isn't complete. There are a few more that I'll promptly fix this morning and update this pull request.

maliberty:
Ready to go?

mminutoli:
> Ready to go?

I'd say so. the rest seems to be from automation that I don't want to be the one breaking :)

maliberty (approved):

PR #3103, opened 2025-04-23
make: metadata-generate now explicitly passes results, logs and reports folder

oharboe:
ORFS does not mandate a layout that genMetrics.py assumed, simply pass in folders to genMetrics.py to fix problem and have a single source of truth(RESULT/REPORTS/LOGS_DIR) in ORFS.

### DIFF
--- a/flow/util/genMetrics.py
+++ b/flow/util/genMetrics.py
@@ -51,6 +51,17 @@ def parse_args():
         "--output", "-o", required=False, default="metadata.json", help="Output file"
     )
     parser.add_argument("--hier", "-x", action="store_true", help="Hierarchical JSON")
+    parser.add_argument("--logs", help="Path to logs", default=None)
+    parser.add_argument(
+        "--reports",
+        help="Path to reports",
+        default=None,
+    )
+    parser.add_argument(
+        "--results",
+        help="Path to results",
+        default=None,
+    )
     args = parser.parse_args()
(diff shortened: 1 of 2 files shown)


flow/util/genMetrics.py
```diff
@@ -404,7 +413,20 @@ def extract_metrics(cwd, platform, design, flow_variant, output, hier_json):
                     continue
                 print(f"Extract Metrics for {plt}, {des}, {variant}")
                 file = "/".join(["reports", plt, des, variant, "metrics.json"])
-                metrics, df = extract_metrics(cwd, plt, des, variant, file, args.hier)
+                logPath = os.path.join(cwd, "logs", plt, des, variant)
+                rptPath = os.path.join(cwd, "reports", plt, des, variant)
+                resultPath = os.path.join(cwd, "results", plt, des, variant)
```
maliberty, line 418: Why are you recomputing these here?  I thought the point is to pass them in.

oharboe: This is the non-utils.mk use case where the .py file scans multiple projects. Is it used? I wanted to let that dog lie in this PR.

maliberty: I think its is used for a bulk update but not certain.  I'm glad to leave it alone.

maliberty (approved):

PR #3105, opened 2025-04-24
genMetrics: remove unreferenced brittle use-case

oharboe:
There is no reference to the use-case of genMetrics.py to scan all local designs, deleting it predicated on review confirmation that it is no longer in use.

genMetrics.py had code to scan designs, but this code would be brittle as it makes assumptions on layout. I ran it locally and it didn't look like it was maintained.

### DIFF
--- a/flow/util/genMetrics.py
+++ b/flow/util/genMetrics.py
@@ -6,7 +6,6 @@
 # -----------------------------------------------------------------------------
 
 import os
-from sys import exit
 from datetime import datetime, timedelta
 from collections import defaultdict
 from uuid import uuid4 as uuid
@@ -26,8 +25,7 @@ def parse_args():
     parser.add_argument(
         "--design",
         "-d",
-        required=False,
-        default="all_designs",
+        required=True,
         help="Design Name for metrics",
     )
     parser.add_argument(
--- a/flow/util/genMetrics.py
+++ b/flow/util/genMetrics.py
@@ -51,17 +51,9 @@ def parse_args():
         "--output", "-o", required=False, default="metadata.json", help="Output file"
     )
     parser.add_argument("--hier", "-x", action="store_true", help="Hierarchical JSON")
-    parser.add_argument("--logs", help="Path to logs", default=None)
-    parser.add_argument(
-        "--reports",
-        help="Path to reports",
-        default=None,
-    )
-    parser.add_argument(
-        "--results",
-        help="Path to results",
-        default=None,
-    )
+    parser.add_argument("--logs", help="Path to logs")
+    parser.add_argument("--reports", help="Path to reports")
+    parser.add_argument("--results", help="Path to results")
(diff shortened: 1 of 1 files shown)


vvbandeira (changes requested):
The `--design` argument should be modified to reflect the proposed changes.

oharboe:
> The `--design` argument should be modified to reflect the proposed changes.

In what way?

This PR is based on the assumption that there is some vestigal code that be deleted as I find no reference to its use and when I tried it, I got a dubious result, which strengthened my belief that there is some old unused code. Deleting it makes it easier to reason about and perform maintenance.

vvbandeira:
> > The `--design` argument should be modified to reflect the proposed changes.
> 
> In what way?
> 
> This PR is based on the assumption that there is some vestigal code that be deleted as I find no reference to its use and when I tried it, I got a dubious result, which strengthened my belief that there is some old unused code. Deleting it makes it easier to reason about and perform maintenance.

```python
    parser.add_argument(
        "--design",
        "-d",
        required=False,
        default="all_designs",
        help="Design Name for metrics",
    )
```

The default value is no longer used anywhere in the code.

oharboe:
> > > The `--design` argument should be modified to reflect the proposed changes.
> > 
> > 
> > In what way?
> > This PR is based on the assumption that there is some vestigal code that be deleted as I find no reference to its use and when I tried it, I got a dubious result, which strengthened my belief that there is some old unused code. Deleting it makes it easier to reason about and perform maintenance.
> 
> ```python
>     parser.add_argument(
>         "--design",
>         "-d",
>         required=False,
>         default="all_designs",
>         help="Design Name for metrics",
>     )
> ```
> 
> The default value is no longer used anywhere in the code.

Fixed.

As a separate PR: then --design name recorded is DESIGN_NICKNAME and not DESIGN_NAME. I don't know what value the DESIGN_NICKNAME has, it is used to give a name to a folder that forms part of e.g. RESULTS_DIR. I believe it would make more sense to record DESIGN_NAME.

    metrics_dict["run__flow__design"] = design

oharboe:
@vvbandeira Actually, it is correct the way it is. DESIGN_NICKNAME makes more sense than DESIGN_NAME.

Consider: 

    export DESIGN_NICKNAME = gcd-ccs

vs:

    export DESIGN_NICKNAME = gcd

In both cases DESIGN_NAME=gcd, but these are two distinct rules-base.json tests.

oharboe:
@vvbandeira @maliberty ready for review, unrelated pr-merge CI failures.

maliberty:
@vvbandeira do you use all the all designs case?  I have a feeling it is a manual use model.

oharboe:
ping?

maliberty:
Vitor is on vacation.  I'll merge and we can always revert if needed.

maliberty (approved):

Issue #3099, opened 2025-04-23
openroad/orfs image tag "latest" is 9+ months old


mguthaus:
### Subject

[Build] for compilation-related issues.

### Describe the bug

The image pushed to dockerhub for openroad/orfs with the tag "latest" is 9+ months old. It should be deleted or updated with the latest image on a regular basis. Otherwise, it causes problems when users use the old image and updated ORFS scripts.

### Expected Behavior

Use the actual latest image or return that an image tag is not specified.

### Environment

```shell
All environments
```

### To Reproduce

```
$ docker run -it openroad/orfs
Unable to find image 'openroad/orfs:latest' locally
latest: Pulling from openroad/orfs
3713021b0277: Already exists
8b2b0aa015e0: Already exists
fc8316ec5ca0: Already exists
2c3de6bea76c: Already exists
7ab2b07af581: Already exists
4364904fcba8: Already exists
12995494f124: Already exists
e88e3788d6db: Pull complete
8f0826d709c0: Pull complete
Digest: sha256:546fb1bfabbfec4fa03c3c25ff60dbf6478daf237cb0387b35e4a3218ad2c805
Status: Downloaded newer image for openroad/orfs:latest
root@0986436c2698:/OpenROAD-flow-scripts#
exit
$ docker images
REPOSITORY                          TAG                   IMAGE ID       CREATED         SIZE
openroad/orfs                       latest                87bbb9a9cffd   9 months ago    11.4GB
```
without a tag will return the 9+ month old version.

### Relevant log output

```shell

```

### Screenshots

![Image](https://github.com/user-attachments/assets/11ecb1ad-8438-4b1d-a1fd-2db709deed86)

jeffng-or:
Looks like the latest tag is getting updated now:

![Image](https://github.com/user-attachments/assets/787344bd-e518-42b8-b84b-d096774a33e8)

PR #3107, opened 2025-04-24
Added latest tag when pushing docker images

sombraSoft:
- Added `-push-latest` parameter to `DockerHelper.sh` so we can tag the image with `latest` only when running in master https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3099
- Updated and added new dry run logs


vvbandeira (changes requested):
You need to change the CI repo. This will update the latest image on every PR, even before it is merged.

vvbandeira (approved):

vvbandeira (approved):
Looks good, just need the new CI tag

PR #3114, opened 2025-04-28
include "-all" parameter to dependency installer description

gudeh:
If I run DependencyInstaller without any input flag it throws an error, so I believe this is the right description

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -232,7 +232,7 @@ _help() {
     cat <<EOF
 
 All arguments and flags are only applicable for OpenROAD dependencies
-Usage: $0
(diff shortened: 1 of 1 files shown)


etc/DependencyInstaller.sh
```diff
@@ -232,7 +232,7 @@ _help() {
     cat <<EOF
 
 All arguments and flags are only applicable for OpenROAD dependencies
-Usage: $0
+Usage: $0 -all
```
```suggestion
Usage: $0 [-all|-base|-common] [-<ARGS>]
```

vvbandeira (changes requested):

maliberty (approved):

vvbandeira:
Replaced by #3306

PR #3108, opened 2025-04-24
[AutoTuner] Fix metadata.json regression

luarss:
This pull request updates the `BaseAlgoEvalSmokeTest` class in `tools/AutoTuner/test/smoke_test_algo_eval.py` to introduce the concept of a `flow_variant` for better configurability and reuse. The changes primarily focus on refactoring the `setUp` method and related commands to use this new variable.

### Refactoring for `flow_variant` integration:

* Introduced a new class variable `flow_variant` with the default value `"smoke-test-algo-eval"` to represent the flow variant. (`[tools/AutoTuner/test/smoke_test_algo_eval.pyR47-R55](diffhunk://#diff-63a772b287b174f31b4335be42c7e910f197140112906da7b1bb169517dfacf8R47-R55)`)
* Updated the `setUp` method to use `flow_variant` in constructing the `experiment` name and the `reference` file path, improving flexibility and consistency. (`[tools/AutoTuner/test/smoke_test_algo_eval.pyR47-R55](diffhunk://#diff-63a772b287b174f31b4335be42c7e910f197140112906da7b1bb169517dfacf8R47-R55)`)
* Refactored the `make_base` method to include `FLOW_VARIANT` in all `make` commands, ensuring that the flow variant is explicitly passed to the build system. (`[tools/AutoTuner/test/smoke_test_algo_eval.pyL71-R79](diffhunk://#diff-63a772b287b174f31b4335be42c7e910f197140112906da7b1bb169517dfacf8L71-R79)`)

PR #3106, opened 2025-04-24
fixed CTS_BUF_DISTANCE handling

jeffng-or:
Splitting https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3066 into multiple PRs to try to get it through. This PR only fixes CTS_BUF_DISTANCE handling.

PR #3025, opened 2025-03-31
[Autotuner] remove git interface

luarss:
This is a companion PR to #2858. 

- Removes git interface, fixes #2836 
- Shifts helm/kubernetes code to `deprecated`

PR #3115, opened 2025-04-28
updated CTS_BUF_DISTANCE handling + AT genMetric.py call fix

jeffng-or:
Merge of https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3106 and part of https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3108.

- Fixes CTS_BUF_DISTANCE flow variable handling
- Fixed call to genMetrics.py in AutoTuner

@luarss , @vvbandeira , @arthurjolo FYI.

### DIFF
--- a/flow/scripts/cts.tcl
+++ b/flow/scripts/cts.tcl
@@ -18,9 +18,9 @@ set cts_args [list \
           -sink_clustering_enable \
           -balance_levels]
 
-# TODO: The first three are no-ops since the arg order is wrong, but hard to get
+append_env_var cts_args CTS_BUF_DISTANCE -distance_between_buffers 1
+# TODO: The next two are no-ops since the arg order is wrong, but hard to get
 # through CI since nine designs change metrics and the PR is blocked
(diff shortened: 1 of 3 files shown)


jeffng-or:
@luarss , @vvbandeira FYI. I removed sky130hd gcd from the sample iteration test because it continually fails with sample == 1. The AT can't find a successful run with only one sample, so it's a fragile test.

luarss:
Yep that is fine - I have pushed a fix in #3025 to be more tolerant of parameter tuning failure (error code 16), so that is a more long term fix

jeffng-or:
> @luarss , @vvbandeira FYI. I removed sky130hd gcd from the sample iteration test because it continually fails with sample == 1. The AT can't find a successful run with only one sample, so it's a fragile test.

It looks like the failure is a seg fault in DPL. I'm able to repro locally, so will file an issue. Closing this PR and will send up a separate PR for the genMetrics.py fix, so that we can get the AT working again.

PR #3118, opened 2025-04-29
test case for reproducing dpl seg fault

jeffng-or:
Test case for reproducing dpl seg fault in sky130hd gcd found by AutoTuner smoke test.

To repro:

```
make DESIGN_CONFIG=designs/sky130hd/gcd/config.mk
```


jeffng-or:
issue resolved. no need for the test case anymore

Issue #3129, opened 2025-05-04
Broken KLayout URL fails Docker build


erendn:
### Subject

[Build] for compilation-related issues.

### Describe the bug

Similar to issue https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1099, the Docker build fails because the current KLayout version was removed from the website.

I tried bumping to the latest version in DependencyInstaller.sh but it still doesn't compile, saying some dependencies are missing. Can someone take a look?

### Expected Behavior

`./build_openroad.sh` should build OpenROAD and all its dependencies.

### Environment

```shell
It is recommened to pull the latest changes.
If problem persists, file a github issue with the re-producible test case.
kernel: Linux 5.4.0-186-generic
os: Ubuntu 20.04.6 LTS (Focal Fossa)
cmake version 3.16.3
-- The CXX compiler identification is GNU 9.4.0
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-21009-ga8ef34ce2
-- System name: Linux
-- Compiler: GNU 9.4.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 9.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds - Success
[...]

PR #3131, opened 2025-05-05
updated klayout version to 0.30.1

jeffng-or:
Per https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3129, the KLayout version is no longer available on the downloads site. Updated to 0.30.1 (latest) and tested on clean Ubuntu 22.04 VM.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -10,7 +10,7 @@ else
 fi
 
 # package versions
-klayoutVersion=0.28.8
+klayoutVersion=0.30.1
 verilatorVersion=5.026
 
 _versionCompare() {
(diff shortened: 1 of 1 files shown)


etc/DependencyInstaller.sh
```diff
@@ -116,6 +116,7 @@ _installUbuntuPackages() {
         help2man \
         libfl-dev \
         libfl2 \
+        libgit2-dev \
```
maliberty, line 119: Does klayout need this now?

maliberty (approved):

etc/DependencyInstaller.sh
jeffng-or: Yes, all of the KLayout .deb files, including the latest 0.28 build, require libgit2-1.1, so I had to add this to the ORFS dependencies. The installation fails if it's not there.

jeffng-or:
@maliberty, looks like the PR failed on mock-array rules. OK to try adding UpdateRules to get the PR through?

maliberty:
mock-array is failing on master and I have a PR for that update.  In any case klayout can't affect metrics so I think we can just go ahead with this.

oharboe:
@jeffng-or @maliberty There's some breakage https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3150

PR #3141, opened 2025-05-07
Added yosys/verific support to synth_preamble

jeffng-or:
Also added updated rules files for asap7/cva6 and nangate45/mempool_group (results generated using yosys-slang, not verific)

### DIFF
--- a/flow/designs/nangate45/mempool_group/rules-base.json
+++ b/flow/designs/nangate45/mempool_group/rules-base.json
@@ -1,18 +1,18 @@
 {
     "synth__design__instance__area__stdcell": {
-        "value": 16477.72,
+        "value": 271098.37,
         "compare": "<="
     },
     "constraints__clocks__count": {
-        "value": 1,
+        "value": 2,
         "compare": "=="
     },
     "placeopt__design__instance__area": {
-        "value": 19790,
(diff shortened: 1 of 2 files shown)


povik (approved):

flow/designs/asap7/cva6/rules-base.json
```diff
@@ -52,7 +52,7 @@
         "compare": ">="
     },
     "finish__design__instance__area": {
-        "value": 16112,
+        "value": 20112,
```
maliberty, line 55: Is this expected?

maliberty (approved):

flow/designs/asap7/cva6/rules-base.json
povik: The old value might be from before #3001 in which case yes, it's expected the instance area has gone up

jeffng-or: Based on the mod dates of the config, the rules file, and the fakeram verilog files, my guess is that the original rules file was run without the blackboxed SRAMs. So, synth just optimized it away. I probably didn't go back and re-gen the rules file when I added the blackboxed SRAMs because yosys-slang wasn't in master yet.

PR #3145, opened 2025-05-07
Re-worked ref_file_check.py to test specific error conditions

jeffng-or:
Previous version of ref_file_check.py checked to see if the AT run failed, but not the specific failure we are looking for. In this rev, I've split test_files() into specific test methods, so that we can check the type of failure.

Had to update the test_autotuner.sh script to call the test class in ref_file_check.py, so that all of the test methods get executed now.

Replacement for https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2903


jeffng-or:
@maliberty, FYI this PR also passes everything except asap7/mock-alu. Seg fault in resizer/STA

maliberty (approved):

maliberty:
I see no path for a connection so it seems unrelated.  I did update sta today so perhaps this will be resolved.

PR #3165, opened 2025-05-14
inject macro definitions during compilation time

eder-matheus:
I've found this strategy to set the __TIME__, __DATE__ and __TIMESTAMP__ as constants without breaking Yosys build, and also without breaking the check for the C++14 support (the error that was happening with verilator).

I've tested it and I can see that what the C++ compiler sees is the constant "0", so I believe it will restore the behavior of having constant date/time in the binary.

### DIFF
--- a/docker/Dockerfile.builder
+++ b/docker/Dockerfile.builder
@@ -15,12 +15,23 @@ COPY --link build_openroad.sh build_openroad.sh
 
 FROM orfs-base AS orfs-builder-base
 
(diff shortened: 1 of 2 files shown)


eder-matheus:
@oharboe FYI.

maliberty (approved):

maliberty:
Ugly but effective. I'd like to see if we can upstream a change to have an build option that avoids the need for such tricks

PR #3172, opened 2025-05-17
[Autotuner] Refactor set_seed

luarss:
- Achieves separation of concerns: set_seed in another function
- cleanup unused imports
- reorganise utils.py with clearer comments

### DIFF
--- a/tools/AutoTuner/src/autotuner/utils.py
+++ b/tools/AutoTuner/src/autotuner/utils.py
@@ -42,7 +42,6 @@ import subprocess
 import sys
 import uuid
 import time
-from multiprocessing import cpu_count
 from datetime import datetime
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


tools/AutoTuner/src/autotuner/utils.py
```diff
@@ -71,6 +70,56 @@
...
+        print(process.stdout)
+
+    if fail_fast and process.returncode != 0:
+        raise RuntimeError
+
+
+def set_seed(seed: int):
+    """Set seed for reproducibility."""
+    import torch
+    import random
```
vvbandeira: Adding the import inside a `def` leads to late dependency resolution. I would prefer to fail fast in this case.

vvbandeira (changes requested):

PR #3178, opened 2025-05-22
ihp-sg13g2: Prepare Multi Corner Analysis

dnltz:
Define the libraries for the slow and fast corner. Not enabled by default. Simply add `export CONRNERS = slow fast` to the config.mk file.


dnltz:
@KrzysztofHerman @sergeiandreyev FYI

maliberty (approved):

PR #3182, opened 2025-05-23
Add variable for move sequence selection.

mguthaus:
Add variable to over-ride the new move sequences.

### DIFF
--- a/flow/scripts/util.tcl
+++ b/flow/scripts/util.tcl
@@ -31,6 +31,7 @@ proc repair_timing_helper { {hold_margin 1} } {
   if {$hold_margin || $::env(HOLD_SLACK_MARGIN) < 0} {
     append_env_var additional_args HOLD_SLACK_MARGIN -hold_margin 1
   }
+  append_env_var additional_args SETUP_MOVE_SEQUENCE -sequence 1
   append_env_var additional_args TNS_END_PERCENT -repair_tns 1
   append_env_var additional_args SKIP_PIN_SWAP -skip_pin_swap 0
   append_env_var additional_args SKIP_GATE_CLONING -skip_gate_cloning 0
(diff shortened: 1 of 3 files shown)


flow/scripts/util.tcl
```diff
@@ -31,6 +31,7 @@ proc repair_timing_helper { {hold_margin 1} } {
   if {$hold_margin || $::env(HOLD_SLACK_MARGIN) < 0} {
     append_env_var additional_args HOLD_SLACK_MARGIN -hold_margin 1
   }
+  append_env_var additional_args SETUP_MOVE_SEQUENCE -sequence 1
```
maliberty, line 34: Needs to be documented

mguthaus: Is this done in the `flow/scripts/variables.yaml`?

maliberty (approved):

flow/scripts/variables.yaml
```diff
@@ -484,6 +484,15 @@ SETUP_SLACK_MARGIN:
     - floorplan
     - grt
   default: 0
+SETUP_MOVE_SEQUENCE:
```
povik: Suggestion:  `SETUP_REPAIR_MOVES`

flow/scripts/util.tcl
povik: @mguthaus yes, and one needs to run `python3 flow/scripts/generate-variables-docs.py` to update the generated markdown

mguthaus: It's not just the moves though, it's an order (sequence) of moves. I think that sequence would be better, but I'm open to other suggestions.

povik: I thought it best to put "repair" into the name. One other idea: `SETUP_REPAIR_SEQUENCE`

mguthaus: Yeah that's better. I also didn't like making it four words which is too long.

maliberty (approved):

maliberty:
You need to fix the DCO

mguthaus:
> You need to fix the DCO

I fixed it

PR #3196, opened 2025-05-28
Updated SRAM sizes for asap7 cva6

jeffng-or:
Updated ram size to 256x256 to prevent RAMs from being optimized out in flat synthesis mode


maliberty (approved):

jeffng-or:
designs/asap7/cva6/rules-base.json updates:
| synth__design__instance__area__stdcell        | 16477.72 |  40692.1 | Failing  |
| placeopt__design__instance__area              |    19790 |    45043 | Failing  |
| placeopt__design__instance__count__stdcell    |   130789 |   164118 | Failing  |
| finish__design__instance__area                |    20112 |    45315 | Failing  |

maliberty:
I guess the jump in area is because we are no longer getting false dead logic elimination?

jeffng-or:
> I guess the jump in area is because we are no longer getting false dead logic elimination?

That and the RAMs are bigger. They were previously 256x32 and now they're 256x256.

PR #3203, opened 2025-05-30
added desired grouping for demo - hier synthesis

jeffng-or:
Tried to find a balance between visual, run time and results when running with SYNTH_HIERARCHICAL=1:

WNS: -415
TNS: -930978
modules: 15
skew: 189.922
total time: 2:36:16

### DIFF
--- a/flow/designs/asap7/cva6/config.mk
+++ b/flow/designs/asap7/cva6/config.mk
@@ -88,5 +88,8 @@ export MACRO_HALO             = 5 5
 # few last gasp iterations
 export SKIP_LAST_GASP ?= 1
 
+ifeq ($(SYNTH_HIERARCHICAL),1)
+	export SYNTH_MINIMUM_KEEP_SIZE ?= 40000
+endif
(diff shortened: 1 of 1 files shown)


maliberty (approved):

flow/designs/asap7/cva6/config.mk
```diff
@@ -88,5 +88,8 @@ export MACRO_HALO             = 5 5
 # few last gasp iterations
 export SKIP_LAST_GASP ?= 1
 
+ifeq ($(SYNTH_HIERARCHICAL),1)
+	export SYNTH_MINIMUM_KEEP_SIZE ?= 40000
```
povik, line 92: I think this can be unconditional and the value will be ignored if not in `SYNTH_HIERARCHICAL`

jeffng-or: true. I was just trying to be extra safe. I can take out the conditional next time I'm updating the config.

PR #3206, opened 2025-06-05
ihp-sg13g2: Minor fixes to enabled BLOCKS

dnltz:
- Update `sg13g2.map` file and use latest one from PDK
- Fix lib handling in ihp-sg13g2 config.mk
- Update I2C GPIO Expander and add a (dummy) macro


flow/scripts/generate_abstract.tcl
```diff
@@ -21,7 +21,15 @@ if {$design_stage >= 4} {
...
 puts "Generating abstract views"
-log_cmd write_timing_model $::env(RESULTS_DIR)/$::env(DESIGN_NAME).lib
+if {[env_var_exists_and_non_empty CORNERS]} {
+  # corners
+  foreach corner $::env(CORNERS) {
+    log_cmd write_timing_model -corner $corner $::env(RESULTS_DIR)/$::env(DESIGN_NAME)_$corner.lib
+  }
+  unset corner
+} else {
+  log_cmd write_timing_model $::env(RESULTS_DIR)/$::env(DESIGN_NAME)_typ.lib
```
oharboe: is the change in .lib file intentional? It will break all users, no?

oharboe (changes requested):
split out ORFS generic changes to a separate PR

flow/scripts/generate_abstract.tcl
dnltz: Yes, that's for purpose because no corner should mean typical. I updated the Makefile to handle that change. This script is only called when creating blocks.

dnltz:
> split out ORFS generic changes to a separate PR

can do that because I have to wait on feedback from IHP anyways.

dnltz:
Moved generic changes into https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3208

flow/scripts/generate_abstract.tcl
oharboe: This is a breaking change for users of ORFS, such as bazel-orfs. The output name is part of the ORFS interface/contract.

oharboe: Regarding changing the external API/contract of generate_abstract.

Reading the example in OpenSTA, I do see the _typ.lib naming convention:

image

dnltz:
@maliberty apparently the `.map` file was out-dated here. This PR now includes the most recent file from the IHP repository, minor fixes in LIB handling and an example block in the I2C GPIO Expander.

maliberty (approved):

maliberty:
Note that VSS is connected to your block through a single strap which could be problematic.

maliberty:
image

PR #3208, opened 2025-06-05
Minor fixes with block handling

dnltz:
- Add Multi Corner support to block generation and rename `.lib` file to `_typ.lib`.
- Add `WORK_HOME` variable to BLOCKS in the base Makefile
- Remove `GDS_HOME` in base Makefile when BLOCKS is set because this file should be assigned in the platform config

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -105,7 +105,6 @@ ifneq ($(BLOCKS),)
   export ADDITIONAL_LEFS += $(BLOCK_LEFS)
   export ADDITIONAL_LIBS += $(BLOCK_LIBS)
   export ADDITIONAL_GDS += $(BLOCK_GDS)
-  export GDS_FILES += $(BLOCK_GDS)
   ifneq ($(CDL_FILES),)
     export CDL_FILES += $(BLOCK_CDL)
   endif
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -97,11 +97,11 @@ export FLOW_VARIANT?=base
 ifneq ($(BLOCKS),)
   # Normally this comes from variables.yaml, but we need it here to set up these variables
   # which are part of the DESIGN_CONFIG. BLOCKS is a Makefile specific concept.
-  $(foreach block,$(BLOCKS),$(eval BLOCK_LEFS += ./results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/${block}.lef))
-  $(foreach block,$(BLOCKS),$(eval BLOCK_LIBS += ./results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/${block}.lib))
-  $(foreach block,$(BLOCKS),$(eval BLOCK_GDS += ./results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/6_final.gds))
-  $(foreach block,$(BLOCKS),$(eval BLOCK_CDL += ./results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/6_final.cdl))
-  $(foreach block,$(BLOCKS),$(eval BLOCK_LOG_FOLDERS += ./logs/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/))
+  $(foreach block,$(BLOCKS),$(eval BLOCK_LEFS += $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/${block}.lef))
+  $(foreach block,$(BLOCKS),$(eval BLOCK_LIBS += $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/${block}.lib))
+  $(foreach block,$(BLOCKS),$(eval BLOCK_GDS += $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/6_final.gds))
+  $(foreach block,$(BLOCKS),$(eval BLOCK_CDL += $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/6_final.cdl))
+  $(foreach block,$(BLOCKS),$(eval BLOCK_LOG_FOLDERS += $(WORK_HOME)/logs/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/))
   export ADDITIONAL_LEFS += $(BLOCK_LEFS)
   export ADDITIONAL_LIBS += $(BLOCK_LIBS)
   export ADDITIONAL_GDS += $(BLOCK_GDS)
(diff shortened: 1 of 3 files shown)


dnltz:
@oharboe FYI

oharboe:
@maliberty  My vision for ORFS is to make it *less* complex and leave all advanced flows to bazel and bazel-orfs.

BLOCKS in ORFS is inflexible and an eyesore, but useful for now for a few ORFS test cases, but as soon as you want to do something interesting it breaks apart as a solution.

Similarly, I wonder if ORFS already has everything for corners and that corners should be done in bazel-orfs on top of ORFS.

This PR has some breaking changes, but I want to first discuss direction.

flow/scripts/generate_abstract.tcl
```diff
@@ -21,7 +21,15 @@ if {$design_stage >= 4} {
 # write_timing_model includes the source latency in the model
 set_clock_latency -source 0 [all_clocks]
 puts "Generating abstract views"
-log_cmd write_timing_model $::env(RESULTS_DIR)/$::env(DESIGN_NAME).lib
+if {[env_var_exists_and_non_empty CORNERS]} {
+  # corners
+  foreach corner $::env(CORNERS) {
+    log_cmd write_timing_model -corner $corner $::env(RESULTS_DIR)/$::env(DESIGN_NAME)_$corner.lib
```
oharboe, line 27: so how does this corner information get into the .odb file or is it in additional_libs?

maliberty:
I think you can advocate for people to use bazel-orfs but we should not block people from using orfs is they choose.  The BLOCKS code already exists so having it not work well isn't great.  It would be easier to make such a case against a totally new feature.

Does bazel-orfs handle multi-corner analysis already?  This is a chance for a sale 😄

oharboe:
I believe that there are lots of more advanced flows that ORFS will struggle to handle and that ORFS has no good way of testing. You will be adding buggy code that complicates the codebase. Proof in point: this already breaks use-cases.

If I can understand corner support today, I dont quite understand the flow, I think I can create an example in bazel-orfs. This does not appear to be a big challenge for a garden variety user specific flow.

Here, for instance, I can imagine that there is a heterogenous flow where typical timing is used for some macros, but not others, which would not be supported by this code.

oharboe:
Another thing: I dont think new PDKs and their test cases should go into ORFS git, but their own git repository(perhaps one per tech or foundry). Example use cases for such PDKs should into their own git repository a la megaboom.

Fixes to ORFS should generally be PDK agnostic.

All of this is easy to tie together with bazel-orfs and getting easier.

maliberty:
This is getter further afield from this PR.  If you want to discuss the future broadly let's have a separate discussion.

oharboe:
> This is getter further afield from this PR. If you want to discuss the future broadly let's have a separate discussion.

I'm arguing that trying to push BLOCKS further will end in tears. Better to leave it as-is as a limited feature intended only for ORFS test flows of the existing complexity and move more advanced flow tests and usage to use ORFS as a component in a solution.

My next step is to understand how corners work in OpenROAD and provide an example use-case in bazel-orfs so that we have something specific to compare against and discuss.

flow/Makefile
```diff
@@ -89,10 +89,12 @@ include $(DESIGN_CONFIG)
 
 export DESIGN_DIR ?= $(dir $(DESIGN_CONFIG))
 
-# default value "base" is duplicated from variables.yaml because we need it
+# default value "base" for FLOW_VARIANT and "." for WORK_HOME are duplicated
+# from variables.yaml and variables.mk because we need it
 # earlier in the flow for BLOCKS. BLOCKS is a feature specific to the
 # ORFS Makefile.
 export FLOW_VARIANT?=base
+export WORK_HOME?=.
```
oharboe, line 97: This is in conflict with WORK_HOME default values further down in this same Makefile. If we go forward with a solution along these lines, then minimally we should move the existing WORK_HOME default values up here and not change the default value behavior. If we want to change the default WORK_HOME behavior, then that's fine, but I don't think the change in default behavior was intentional here.

flow/Makefile
```diff
@@ -105,7 +105,6 @@ ifneq ($(BLOCKS),)
   export ADDITIONAL_LEFS += $(BLOCK_LEFS)
   export ADDITIONAL_LIBS += $(BLOCK_LIBS)
   export ADDITIONAL_GDS += $(BLOCK_GDS)
-  export GDS_FILES += $(BLOCK_GDS)
   ifneq ($(CDL_FILES),)
```
oharboe, line 115: @maliberty Isn't this a standalone bugfix? Split out into separate PR?

I'm not very familiar with the gds part of the flow.

flow/Makefile
```diff
@@ -98,12 +98,17 @@ ifneq ($(BLOCKS),)
   # Normally this comes from variables.yaml, but we need it here to set up these variables
   # which are part of the DESIGN_CONFIG. BLOCKS is a Makefile specific concept.
   $(foreach block,$(BLOCKS),$(eval BLOCK_LEFS += $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/${block}.lef))
-  $(foreach block,$(BLOCKS),$(eval BLOCK_LIBS += $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/${block}.lib))
+  $(foreach block,$(BLOCKS),$(eval BLOCK_TYP_LIBS += $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/${block}_typ.lib))
+  $(foreach block,$(BLOCKS),$(eval BLOCK_FAST_LIBS += $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKNAME)_$(block)/$(FLOW_VARIANT)/${block}_fast.lib))
```
oharboe, line 104: I think this is problematic from the point of view that it is the macro config.mk that has to use CORNERS and then generate an abstract with corner .lib files. Makefiles break down here because CORNERS set up in the project defining BLOCKS can be different than CORNERS inside the macro listed in BLOCKS.

oharboe: In other words: if some of the macros in BLOCKS use corners and others don't, then Makefile can't express that here. I don't know if such a thing makes sense in the first place, but it is an example of a more customized flow that a user could need that ORFS won't have adequate tests for.

oharboe:
@maliberty Is there any OpenROAD GUI support for multiple corners?

flow/scripts/generate_abstract.tcl
oharboe: There's some sort of naming convention going on here in ORFS w.r.t. corners. Unsure if this has been used/tested or if it is just some early ideas/beginnings:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/5ac9869c28b20db2f5881ba20fb6223dcdd0b75e/flow/scripts/read_liberty.tcl#L2-L18

More CORNERS bits in ORFS:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/5ac9869c28b20db2f5881ba20fb6223dcdd0b75e/flow/scripts/report_metrics.tcl#L218-L228

oharboe:
@maliberty What about rules-base.json testing? Today it ORFS can set a single corner via CORNER for the entire flow and then test at the end, but rules-base.json would have to be compare against something against all CORNERS.

Does having multiple CORNERS defined have an effect on the quality of results or is it purely a reporting issue?

flow/Makefile
dnltz: Makes sense. I will remove it from `flow/scripts/variables.mk`

flow/Makefile
dnltz: Yes, platforms should set `GDS_FILES` and  use `ADDITIONAL_GDS`.

flow/scripts/generate_abstract.tcl
dnltz: This entire logic is basically copied from the `read_liberty.tcl` function. OpenSTA is already aware of corners, because it used `read_liberty.tcl` to read in .lib files for various corners. Afterwards, you cannot simply write just one timing model. You need to write timing models for each corner.

Regarding naming convention: I think it's pretty common to add the corner as last item to the file name.

dnltz:
> I think you can advocate for people to use bazel-orfs but we should not block people from using orfs is they choose. The BLOCKS code already exists so having it not work well isn't great. It would be easier to make such a case against a totally new feature.
> 
> Does bazel-orfs handle multi-corner analysis already? This is a chance for a sale 😄

I agree. We either remove this feature entirely or fix it. Otherwise someone else will give it a try and fix it (again).

dnltz:
@maliberty can you take a look at this PR and make a decision?

flow/Makefile
maliberty: Corners should be consistent across hierarchy

maliberty (approved):
I think this is ok.  The change to use _typ when no corners is defined simplifies the code but is slightly different than the current operation where no suffix is used.  I think the tradeoff is ok.

flow/scripts/variables.mk
```diff
@@ -18,11 +18,10 @@ export DESIGN_NICKNAME?=$(DESIGN_NAME)
 # Setup variables to point to other location for the following sub directory
 # - designs - default is under current directory
 # - platforms - default is under current directory
-# - work home - default is current directory
 # - utils, scripts, test - default is under current directory
 export DESIGN_HOME   ?= $(FLOW_HOME)/designs
 export PLATFORM_HOME ?= $(FLOW_HOME)/platforms
-export WORK_HOME     ?= .
+# WORK_HOME is set up in flow/Makefile
```
oharboe, line 24: Nit: I'm removing this comment in a followup PR. This is a comment left in place for a deleted line. Several variables are defined in flow/Makefile, WORK_HOME is not special in that regard.

Issue #3179, opened 2025-05-22
need support for equivalence checking using .lib models


Cronus-38:
### Description

The equivalence checker built into Yosys, eqy, supports the use of .lib models.  However, the ORFS flow currently only supports the use of verilog simulation models.  This is problematic since it's not uncommon for simulation models for cells to be represented as UDPs, which are _not_ supported by eqy.  This makes the equivalence checking feature in ORFS unusable for many vendor libraries -- unless the user hacks the simulation models and replaces the UDP descriptions with synthesizable representations, but that's not really an acceptable solution.

### Suggested Solution

To solve this, ORFS could provide a variable which controls what type of model is used for equivalence checking:  verilog simulation models or eqy.  This variable would be used to control branching logic in the write_eqy_script procedure in the load.tcl file.  For example, the variable named EQUIVALENCE_CHECK_USE_LIBS could contain a list of .lib files to use, and the code could look like this:

```
proc write_eqy_script {} {
    set top_cell [current_design]
    set cell_files [get_verilog_cells_for_design]
    set outfile [open "$::env(OBJECTS_DIR)/4_eqy_test.eqy" w]

    # Gold netlist
    puts $outfile "\[gold]"
    if { [env_var_exists_and_non_empty EQUIVALENCE_CHECK_USE_LIBS] } {
        puts $outfile "read_liberty -ignore_miss_func $::env(EQUIVALENCE_CHECK_USE_LIBS)"
    } else {
        puts $outfile "read_verilog -sv $cell_files"
    }
    puts $outfile "read_verilog -sv $::env(RESULTS_DIR)/4_before_rsz.v \n"
    puts $outfile "prep -top $top_cell -flatten\nmemory_map\n\n"

    # Modified netlist
    puts $outfile "\[gate]"
    if { [env_var_exists_and_non_empty EQUIVALENCE_CHECK_USE_LIBS] } {
        puts $outfile "read_liberty -ignore_miss_func $::env(EQUIVALENCE_CHECK_USE_LIBS)"
    } else {
        puts $outfile "read_verilog -sv $cell_files"
    }
    puts $outfile "read_verilog -sv $::env(RESULTS_DIR)/4_after_rsz.v \n"
    puts $outfile "prep -top $top_cell -flatten\nmemory_map\n\n"
...

```

maliberty:
@povik any opinions?

povik:
[...]

PR #1447, opened 2023-09-07
Eqy check

openroadie:
Adding eqy checks into ORFS with help from the eqy team on how to structure the input config file.

PR #3223, opened 2025-06-10
updated asap7 platform config.mk to support multi vt

jeffng-or:
- Updated jpeg_lvt to leverage ASAP7_USE_VT instead of manually adding the LVT asap7 lib files - more in line with aes_lvt and ethmac_lvt
- Extended ASAP7_USE_VT to support one or more Vt's.
- Default remains a single Vt (RVT)
- The first Vt specified is treated as the primary Vt and is used to set the tie hi/tie lo and other cells
- The other Vts are treated as secondary and their Lib, LEF, GDS, and fill cells are added to the appropriate lists
- Note that designs that use CCS will not be able to use multi-Vt unless we add the LVT and SLVT CCS files to the repo

### DIFF
--- a/flow/designs/asap7/jpeg_lvt/config.mk
+++ b/flow/designs/asap7/jpeg_lvt/config.mk
@@ -8,15 +8,6 @@ export VERILOG_INCLUDE_DIRS   = $(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/include
 export SDC_FILE               = $(DESIGN_HOME)/$(PLATFORM)/$(DESIGN_NICKNAME)/jpeg_encoder15_7nm.sdc
 export ABC_AREA               = 1
 
-export ADDITIONAL_LIBS        = $(LIB_DIR)/asap7sc7p5t_AO_LVT_FF_nldm_211120.lib.gz \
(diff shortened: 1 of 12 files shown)


flow/platforms/asap7/work_around_yosys/asap7sc6t_INVBUF_LVT_FF_210930.v
maliberty, line 1: Are we using 6T cells?

@povik I've forgotten why we need this workaround but it would be nice if we didn't.  Do you know/recall the history?

povik: I can see it was added in #1447 for use in the EQY check. If we switch to use Liberty models as discussed in #3179 we won't need this anymore.

jeffng-or: For the 6T question, it doesn't look like it. We only have only using 7.5T LEF/Lib/GDS in our repo. 

I ended up duplicating what was already there for RVT. but can remove all the 6T files as part of this PR.

PR #3227, opened 2025-06-12
updated asap7 platform config.mk to support multi vt

jeffng-or:
Replaces https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3223

- Updated jpeg_lvt to leverage ASAP7_USE_VT instead of manually adding the LVT asap7 lib files - more in line with aes_lvt and ethmac_lvt
- Extended ASAP7_USE_VT to support one or more Vt's.
- Default remains a single Vt (RVT)
- The first Vt specified is treated as the primary Vt and is used to set the tie hi/tie lo and other cells
- The other Vts are treated as secondary and their Lib, LEF, GDS, and fill cells are added to the appropriate lists
- Note that designs that use CCS will not be able to use multi-Vt unless we add the LVT and SLVT CCS files to the repo

### DIFF
--- a/flow/designs/asap7/jpeg_lvt/config.mk
+++ b/flow/designs/asap7/jpeg_lvt/config.mk
@@ -8,15 +8,6 @@ export VERILOG_INCLUDE_DIRS   = $(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/include
 export SDC_FILE               = $(DESIGN_HOME)/$(PLATFORM)/$(DESIGN_NICKNAME)/jpeg_encoder15_7nm.sdc
 export ABC_AREA               = 1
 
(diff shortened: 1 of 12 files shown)


jeffng-or:
@maliberty, I'm getting failures in non-asap7 designs for some reason.

both pr-head and pr-merge fail with the same failures:
- jpeg ihp-sg13g2: metadata update
- ariane136 nangate45: metadata update
- mampool_group nangate45: routing congestion too high
-  gcd sky130hd: metadata update
- microwatt sky130hd: routing congestion too high
- aes sky130hs: metadata update
- ibex sky130hs: metadata update
- riscv32i sky130hs: metadata update

Any thoughts on next steps?

jeffng-or:
@maliberty, yay! It passed after rebasing.

maliberty (approved):

PR #3228, opened 2025-06-12
update documentation on metal layers for pin placement

eder-matheus:


maliberty (approved):

maliberty:
@luarss what is the issue that the variables.yaml is flagging?  The output is unclear

rovinski:
I think it's saying `FlowVariables.md` needs to be updated as well

PR #3232, opened 2025-06-12
ihp-sg13g2: more consistent

oharboe:
with other PDKs and less brittle.


maliberty (approved):

oharboe:
@maliberty looks like unrelated errors, merge?

PR #3233, opened 2025-06-12
ihp-sg13g2: fix SDC_FILE lockup problem

oharboe:
`make DESIGN_CONFIG=designs/ihp-sg13g2/gcd/config.mk SDC_FILE= do-floorplan`  would
cause make to simply lock up, now it terminates.

Needed by https://github.com/aesc-silicon/ElemRV/pull/6


maliberty (approved):
Why would it lock up?

maliberty:
Is this only relevant to IHP?

oharboe:
> Is this only relevant to IHP?

Only IHP reimplemented this, so good enough for now. The code could do with some refactoring to have a non PDK part that extracts the clock period and a PDK scaling factor.

PR #3243, opened 2025-06-16
update klayout checksum for ubuntu 20.04

LucasYuki:
Update KLayout 0.28.17 checksum for Ubuntu 20.04.

Checksum found in https://www.klayout.de/build.html
image

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -170,7 +170,7 @@ _installUbuntuPackages() {
         fi
         else
             if [[ $1 == 20.04 ]]; then
-                klayoutChecksum=15a26f74cf396d8a10b7985ed70ab135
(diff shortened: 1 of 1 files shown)


etc/DependencyInstaller.sh
```diff
@@ -170,7 +170,7 @@ _installUbuntuPackages() {
         fi
         else
             if [[ $1 == 20.04 ]]; then
-                klayoutChecksum=15a26f74cf396d8a10b7985ed70ab135
+                klayoutChecksum=f78d41edf5bcfa5f1990bde1a9307e9e
```
maliberty, line 173: If this was wrong why wasn't there a failure?  (@vvbandeira thoughts?)

maliberty (approved):

PR #3248, opened 2025-06-17
docs: update

oharboe:
@vvbandeira @maliberty Strange, how can this be out of date? Why didn't CI catch this?


oharboe:
rolled into other pull requests where needed.

PR #3253, opened 2025-06-18
Allow user to specify a custom hierarchy separator for the synthesis `flatten` pass

magancarz:
This PR adds ability to specify a custom hierarchy separator which will be used for the `flatten` pass during synthesis.

### DIFF
--- a/flow/scripts/synth.tcl
+++ b/flow/scripts/synth.tcl
@@ -17,6 +17,10 @@ if {[env_var_exists_and_non_empty SYNTH_KEEP_MODULES]} {
   }
 }
 
+if {[env_var_exists_and_non_empty SYNTH_HIER_SEPARATOR]} {
+  scratchpad -set flatten.separator $::env(SYNTH_HIER_SEPARATOR)
+}
+
 set synth_full_args $::env(SYNTH_ARGS)
 if {[env_var_exists_and_non_empty SYNTH_OPERATIONS_ARGS]} {
   set synth_full_args [concat $synth_full_args $::env(SYNTH_OPERATIONS_ARGS)]
--- a/flow/scripts/variables.yaml
+++ b/flow/scripts/variables.yaml
@@ -677,6 +677,10 @@ SYNTH_ARGS:
   description: |
     Optional synthesis variables for yosys.
   default: -flatten
+SYNTH_HIER_SEPARATOR:
+  description: |
+    Separator used for the synthesis flatten stage.
+  default: .
 VERILOG_TOP_PARAMS:
   description: |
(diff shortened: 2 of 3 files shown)


flow/scripts/synth.tcl
```diff
@@ -17,6 +17,11 @@ if {[env_var_exists_and_non_empty SYNTH_KEEP_MODULES]} {
   }
 }
 
+set flatten_hier_separator .
+if {[env_var_exists_and_non_empty SYNTH_HIER_SEPARATOR]} {
```
povik, line 20: We should document this option

povik (commented):
Looks good apart from missing documentation

flow/scripts/synth.tcl
povik: There's a cleaner way to write this:

```
if {[env_var_exists_and_non_empty SYNTH_HIER_SEPARATOR]} {
  scratchpad -set flatten.separator $::env(SYNTH_HIER_SEPARATOR)
}
```

This applies it as a global Yosys setting. I think Silicon Compiler does the same. If we do it this way we don't need to take flatten out of synth.

flow/scripts/variables.yaml
```diff
@@ -676,7 +676,7 @@ SYNTH_KEEP_MODULES:
 SYNTH_ARGS:
   description: |
     Optional synthesis variables for yosys.
-  default: -flatten
+  default: ""
```
povik: No matter how we make the change to apply the separator I think it's good to remove `-flatten` from the default here

magancarz: Not specifying flattening stage as default causes some tests, i.e. `uart asap7` to fail. This can be fixed by adding `-flatten` as `SYNTH_ARGS` value in `config.mk`. Should we include these changes in the current PR? I initially removed this from the default `SYNTH_ARGS` value, to prevent running flattening stage twice.

magancarz:
Thanks for the review @povik. I applied some of your suggestions and it also helped with the failing test cases.

flow/scripts/variables.yaml
povik: I meant remove it here put pass it to synth explicitly. Let's leave it for another PR

povik (approved):

maliberty (approved):

flow/scripts/variables.yaml
maliberty: Who will make this other PR?

magancarz: I added this change in PR #3260.

povik: Thanks!

PR #3256, opened 2025-06-19
versions.txt: non-zero exit code when commands don't exist

oharboe:
use-case: in bazel-orfs when generating issues for floorplan, yosys isn't a dependency and hence isn't in e.g. the
`bazelisk build MockArray_floorplan --sandbox_debug` environment.


maliberty (approved):

oharboe:
@maliberty bugfix and refinement in latest commit

PR #3260, opened 2025-06-23
Add `-flatten` to synth pass explicitly

magancarz:
This PR explicitly adds `-flatten` flag as a default synth argument.

### DIFF
--- a/flow/scripts/variables.yaml
+++ b/flow/scripts/variables.yaml
@@ -683,7 +683,7 @@ SYNTH_KEEP_MODULES:
 SYNTH_ARGS:
   description: |
     Optional synthesis variables for yosys.
-  default: -flatten
+  default: ""
 SYNTH_HIER_SEPARATOR:
   description: |
     Separator used for the synthesis flatten stage.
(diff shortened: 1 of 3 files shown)


povik (approved):

maliberty (approved):

flow/scripts/variables.yaml
```diff
@@ -683,7 +683,7 @@ SYNTH_KEEP_MODULES:
 SYNTH_ARGS:
   description: |
     Optional synthesis variables for yosys.
-  default: -flatten
+  default: ""
```
oharboe, line 686: nit: for consistency delete this line since default is blank

povik: Do you agree we can leave this for whenever someone is touching the surroundings?

oharboe: Certainly. There is no functional change.

oharboe: [Fixed](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3263)

Removed examples that shouldn't be followed.

PR #3265, opened 2025-06-26
scripts: whitespace consistency, tabs not spaces and misc

oharboe:
less noisy future PRs.


maliberty:
I disagree that "tabs not spaces" is an improvement.  Python prefers spacing (see pep8).  Tabs are inconsistently handled across various tools.

oharboe:
> I disagree that "tabs not spaces" is an improvement.  Python prefers spacing (see pep8).  Tabs are inconsistently handled across various tools.

I would prefer "black" for .tcl, I prefer consistency over any particular style.

As it is now, I saw some space/tab thrashing in a PR.

maliberty:
In OR we are using https://github.com/nmoroze/tclint and the same could be applied here.  Unfortunately there is no black for tcl.  I would be glad to have a spaces only rule.

PR #3266, opened 2025-06-26
README: add bazel-orfs as an installation option

oharboe:
Not exactly expecting a rush :-), but increase bazel-orfs visibility.


oharboe:
@maliberty unrelated CI network error

maliberty (approved):

PR #3271, opened 2025-06-27
variables: add type info for tunable variables

oharboe:
DRY, AutoTuner type use-cases can use this information, also it is documentation.

### DIFF
--- a/flow/scripts/variables.yaml
+++ b/flow/scripts/variables.yaml
@@ -362,7 +362,7 @@ CELL_PAD_IN_SITES_GLOBAL_PLACEMENT:
     - place
     - floorplan
   default: 0
-  type: float
+  type: int
   tunable: 1
 CELL_PAD_IN_SITES_DETAIL_PLACEMENT:
   description: >
@@ -373,7 +373,7 @@ CELL_PAD_IN_SITES_DETAIL_PLACEMENT:
     - cts
     - grt
   default: 0
-  type: float
+  type: int
   tunable: 1
 PLACE_PINS_ARGS:
   description: |
--- a/flow/scripts/variables.yaml
+++ b/flow/scripts/variables.yaml
@@ -95,6 +95,7 @@ CORE_UTILIZATION:
   stages:
     - floorplan
   tunable: 1
+  type: float
 CORE_AREA:
   description: >
     The core area specified as a list of lower-left and upper-right corners in
@@ -361,6 +362,7 @@ CELL_PAD_IN_SITES_GLOBAL_PLACEMENT:
     - place
     - floorplan
   default: 0
+  type: float
   tunable: 1
 CELL_PAD_IN_SITES_DETAIL_PLACEMENT:
   description: >
@@ -371,6 +373,7 @@ CELL_PAD_IN_SITES_DETAIL_PLACEMENT:
     - cts
     - grt
   default: 0
+  type: float
   tunable: 1
 PLACE_PINS_ARGS:
   description: |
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty unrelated CI error `Found unhandled java.io.IOException exception: `

flow/scripts/variables.yaml
```diff
@@ -361,6 +362,7 @@ CELL_PAD_IN_SITES_GLOBAL_PLACEMENT:
     - place
     - floorplan
   default: 0
+  type: float
```
maliberty: int

flow/scripts/variables.yaml
```diff
@@ -371,6 +373,7 @@ CELL_PAD_IN_SITES_DETAIL_PLACEMENT:
     - cts
     - grt
   default: 0
+  type: float
```
maliberty: int

maliberty (changes requested):

flow/scripts/variables.yaml
oharboe: but it is in microns, no?

flow/scripts/variables.yaml
oharboe: it is in microns, no?

flow/scripts/variables.yaml
maliberty: It's right in the name, in sites

oharboe: Will fix, but could you extend the description a bit, to state what is obvious to those skilled in the art?

maliberty: Perhaps I'm too skilled as " Cell padding on both sides **in site widths** to ease routability during global placement." seems clear to me.  What do you suggest?

flow/scripts/variables.yaml
oharboe: This is copilot's elaboration on these variables. So if this explanation is not grossly misleading, then it means that the documentation is sufficiently constrainting AI that the documentation is effective.

# Explanation of Variables: `CELL_PAD_IN_SITES_GLOBAL_PLACEMENT` and `CELL_PAD_IN_SITES_DETAIL_PLACEMENT`

## `CELL_PAD_IN_SITES_GLOBAL_PLACEMENT`

### Description
Specifies cell padding (in site widths) on both sides during **global placement** to improve routability.

### Key Details
- **Stages**: `place`, `floorplan`
- **Default**: `0` (no padding applied by default)
- **Type**: `int` (padding must align with discrete site widths)
- **Tunable**: `1` (adjustable for specific designs)

---

## `CELL_PAD_IN_SITES_DETAIL_PLACEMENT`

### Description
Specifies cell padding (in site widths) on both sides during **detailed placement** to improve routability.

### Key Details
- **Stages**: `place`, `cts`, `grt`
- **Default**: `0` (no padding applied by default)
- **Type**: `int` (padding must align with discrete site widths)
- **Tunable**: `1` (adjustable for specific designs)

---

## Why Use `int` Instead of `float`?
- **Discrete Nature**: Site widths are the smallest unit of placement and must align with the placement grid.
- **Tool Compatibility**: Placement tools expect integer values for padding.
- **Simplicity**: Fractional values (e.g., `0.5`) are not meaningful in this context and could cause alignment issues.

oharboe: To me, the documentation in master passes this litmus test of being unambigious and adequate.

More context could be convenient, but apparently isnt required

maliberty (approved):

PR #3275, opened 2025-06-28
[Autotuner] Fix tensorboard hparam regression

### LINKED ISSUE #3273: protobuf 5.26.1 breaks hparams display in AutoTuner tensorboard
### Subject

[Stage]: Other. Please describe below.

### Describe the bug

The current AT requirements.txt versions break the HPARAMS display in tensorboard. After doing a bunch or searching, it seems like protobuf compatibility is the culprit. 

### Expected Behavior

HPARAMS display works

Some GUI-based testing before release would also be a good idea.

### Environment

```shell
current head of master
```

### To Reproduce

run AT
tensorboard -logdir path_to_AT_run --host 0.0.0.0 --port 6006
bring up localhost:6006
click on HPARAMS

If it says that there's no HPARAMS data, then there's a problem.

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

luarss:
- tensorboard < 2.16.2 used some old protobuf functions that were dropped in current protobuf, so a bump in version is needed.
fixes #3273

### DIFF
--- a/tools/AutoTuner/requirements.txt
+++ b/tools/AutoTuner/requirements.txt
@@ -5,7 +5,7 @@ optuna==3.6.0
 pandas>=2.0,<=2.2.1
 bayesian-optimization==1.4.0
 colorama==0.4.6
-tensorboard>=2.14.0,<=2.16.2
+tensorboard>=2.17.0
 protobuf>=5.26.1
 SQLAlchemy==1.4.17
 urllib3>=1.26.17


maliberty (approved):

jeffng-or:
cool. Confirmed that tensorboard 2.19.0 doesn't have an issue. Thanks!

PR #3279, opened 2025-06-30
Tclint 2

luarss:
- Targets tclint for `flow/util` scripts


maliberty:
Why use `tclint-disable-next-line line-length` instead of just breaking the line(s)?

luarss:
Cases I encountered for this were:
- Line length violations are comments -> no harm splitting, or we can preserve original 
- `file_name` cannot be neatly divided into lines without introducing newlines

maliberty:
Cell veneer will hopefully go away one day so I won't worry about it too much.

maliberty (approved):

PR #3289, opened 2025-07-04
sdc: simplify part 2

oharboe:
Less sure about this change...


maliberty:
These SDC look rather dubious but it doesn't seem harmful.  You missed one non_clock_inputs in each (the first clock)

oharboe:
I'll let this dog lie for somebody else to kick..

PR #3291, opened 2025-07-04
Make print issue variables fix

oharboe:
make print-ISSUE_VARIABLES now works. It failed because it contains newlines.


maliberty:
Please give an example of the problem this is solving.

oharboe:
> Please give an example of the problem this is solving.

`make print-ISSUE_VARIABLES` now works, previously it failed with an error

maliberty (approved):

PR #3294, opened 2025-07-07
logging: log hash alongside elapsed to help find divergent/non-deterministic results

oharboe:
Example. A failed step could end up with log and no .odb file.

```
$ make elapsed
./logs/nangate45/gcd/base
Log                        Elapsed/s Peak Memory/MB  sha1sum .odb [0:20)
1_1_yosys                          0             43                  N/A
1_1_yosys_canonicalize             0             40                  N/A
2_1_floorplan                      0             98 50eac5062cd153c214bc
2_2_floorplan_macro                0             94 50eac5062cd153c214bc
2_3_floorplan_tapcell              0             94 ac3fe102c900185d2480
2_4_floorplan_pdn                  0             96 59b27705fdbd75c33f36
3_1_place_gp_skip_io               0             95 d6b4e2de32af97e51809
3_2_place_iop                      0             95 abe5f19e242458533155
3_3_place_gp                       1            211 1a7e366da16e932d2917
3_4_place_resized                  0            114 0171f43d6d2d17c772e9
3_5_place_dp                       0            100 79e4817b19df4b20222a
4_1_cts                            2            127 b342824a6e10d783e773
5_1_grt                            5            214 ba9e7089ccedd24d7e9f
5_2_route                         11           1215 f48628e592653b8c3c3f
5_3_fillcell                       0             97 f32e153bcd24ed0862d5
6_1_fill                           0             95 f32e153bcd24ed0862d5
6_1_merge                          1            376                  N/A
6_report                           1            250                  N/A
Total                             21           1215                     
```

### DIFF
--- a/flow/test/test_genElapsedTime.py
+++ b/flow/test/test_genElapsedTime.py
@@ -30,7 +30,7 @@ class TestElapsedTime(unittest.TestCase):
         genElapsedTime.scan_logs(["--logDir", str(self.tmp_dir.name), "--noHeader"])
         # check if output is correct
         expected_output = (
-            self.tmp_dir.name + "\n1_test 5400 9440\nTotal 5400 9440\n"
+            self.tmp_dir.name + "\n1_test 5400 9440 N/A\nTotal 5400 9440\n"
         ).split()
         actual_output = mock_stdout.getvalue().split()
         self.assertEqual(actual_output, expected_output)
@@ -44,7 +44,7 @@ class TestElapsedTime(unittest.TestCase):
         # call the script with the test log file
         genElapsedTime.scan_logs(["--logDir", str(self.tmp_dir.name), "--noHeader"])
         expected_output = (
-            self.tmp_dir.name + "\n1_test 74 9440\nTotal 74 9440\n"
+            self.tmp_dir.name + "\n1_test 74 9440 N/A\nTotal 74 9440\n"
         ).split()
         actual_output = mock_stdout.getvalue().split()
         self.assertEqual(actual_output, expected_output)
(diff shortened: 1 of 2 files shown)


vvbandeira:
If `yosys` fails to run, how come the rest of the flow can still run?

oharboe:
> If `yosys` fails to run, how come the rest of the flow can still run?

Yosys didn't fail, the .odb hash just is N/A (not applicable) to yosys.

oharboe:
Also, locally, `make elapsed` can be run on after one build ran to completion and another failed, so the logs/results could contain successful builds up to, say, a failing floorplan stage and then successful logs/results after that. `make elapsed` just reports on the logs and hashes of .odb files without worrying about successful/failed builds.

vvbandeira (approved):

oharboe:
@vvbandeira Couldy you automerge? I need a docker image of this before I can try out https://github.com/The-OpenROAD-Project/OpenROAD/pull/7722

maliberty:
For the yosys step you could hash to .v as it would serve the same purpose of isolating the source of a diff.

oharboe:
> For the yosys step you could hash to .v as it would serve the same purpose of isolating the source of a diff.

Well... yosys determinism is a bit fickle, so I'd like to lave that be for now. A refinement if it turns out ot be useful.

oharboe:
Same for .gds, hash there doesn't make sense while it contains non-determinism like build dates and more.

maliberty:
yosys is deterministic afaik, its just very sensitive to input changes.  For a fixed version and a fixed input you should get the same result consistently.  The hash is only a debugging tool and it seems useful for that as much as it is for OR.

oharboe:
> yosys is deterministic afaik, its just very sensitive to input changes. For a fixed version and a fixed input you should get the same result consistently. The hash is only a debugging tool and it seems useful for that as much as it is for OR.

Agreed: a refinement beyond this PR.

oharboe:
@maliberty I'll experiment with yosys after this is in master, I suspect I'll run into some snags. This covers 80-90% of the use-case. I'll have a quickly look at canonicalizing .gds to see if a hash is possible there.

maliberty:
gds is rarely a problem and is at the end of the flow so it is a lower priority for me.

oharboe:
@maliberty I have the rest lined up, but can we merge this so I can have  docker image tomorrow morning and continue my chase?

maliberty (approved):

oharboe:
@maliberty Merge manually? There is an unrelated error/lockup here https://github.com/issues/created?issue=The-OpenROAD-Project%7COpenROAD%7C7741

PR #3295, opened 2025-07-07
synth: more consistent naming

oharboe:
I'm not keen on changing 1_1_yosys stem as I don't know what tools that rely on that name, like metrics, but canonicalization happens before 1_1_yosys, so the compromise is to create 1_0_yosys_canonicalize stem.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -286,12 +286,12 @@ do-yosys: $(DONT_USE_SC_LIB)
 
 .PHONY: do-yosys-canonicalize
 do-yosys-canonicalize: yosys-dependencies $(DONT_USE_SC_LIB)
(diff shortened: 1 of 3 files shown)


maliberty:
1_0 is inconsistent with how we name substeps in all other steps in the flow.  I find that more objectionable than have 1_1 become 1_2.  We have added/removed steps in the flow before so its possible.

oharboe:
> 1_0 is inconsistent with how we name substeps in all other steps in the flow. I find that more objectionable than have 1_1 become 1_2. We have added/removed steps in the flow before so its possible.

I'll look at that tomorrow. I want to have this in the can before I create a followup to https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3294 to do hashes for synthesis

PR #3297, opened 2025-07-07
synthesis: more consistent naming of logs and results

oharboe:

PR #3299, opened 2025-07-07
Log hashes per step

oharboe:
Review only the most recent commit, builds on https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3297

Example output:

```
$ make do-2_1_floorplan
[deleted]
Design area 650 u^2 57% utilization.
Elapsed time: 0:00.28[h:]min:sec. CPU time: user 0.29 sys 0.12 (147%). Peak memory: 101020KB.
Log                        Elapsed/s Peak Memory/MB  sha1sum .odb [0:20)
2_1_floorplan                      0             98 50eac5062cd153c214bc
```

In the bazel-orfs use-case, the .odb files are not all present in the finish step, so log as we go along in addition to the finish    step.
    
A little bit more verbose than I could have wished for, but it does the job. Refine later, possibly.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -258,7 +258,7 @@ synth-report: synth
 
 .PHONY: do-synth-report
 do-synth-report:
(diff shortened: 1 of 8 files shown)


maliberty (approved):

oharboe:
@maliberty merge manually? asap7 cva6 timeouts...

PR #3300, opened 2025-07-08
Tclint 4

### LINKED ISSUE #3268: add tclint to CI like in openroad
In OR we are using https://github.com/nmoroze/tclint and the same could be applied here.  Unfortunately there is no black for tcl.  I would be glad to have a spaces only rule.

_Originally posted by @maliberty in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3265#issuecomment-3009137791_

luarss:
- Concludes adding tclint to ORFS, resolves #3268 
- adds flow/platforms and flow/scripts
- replace `tclint-disable-next-line` with `tclint-disable-line` where appropriate for code readability

### DIFF
--- a/flow/scripts/synth_preamble.tcl
+++ b/flow/scripts/synth_preamble.tcl
@@ -141,7 +141,7 @@ proc convert_liberty_areas { } {
       continue
     }
     set area [rtlil::get_attr -string -mod $cell area]
-    if { $found_cell == "" || [$area < $found_cell_area] } {
+    if { $found_cell == "" || $area < $found_cell_area } {
       set found_cell $cell
       set found_cell_area $area
     }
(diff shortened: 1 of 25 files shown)


flow/scripts/synth_preamble.tcl
```diff
@@ -141,7 +141,7 @@ proc convert_liberty_areas { } {
       continue
     }
     set area [rtlil::get_attr -string -mod $cell area]
-    if { $found_cell == "" || [expr $area < $found_cell_area] } {
+    if { $found_cell == "" || [$area < $found_cell_area] } {
```
maliberty: No need for [] now that expr is gone

maliberty (approved):

PR #3308, opened 2025-07-09
Moved ODB hashing into its own method and made it compatible with pre…

jeffng-or:
…-Python 3.8


jeffng-or:
Tested in environment with python 3.6 which previously failed

maliberty (approved):

PR #3313, opened 2025-07-11
make: silently suppress warnings in read_liberty.tcl

oharboe:
This fixes  and `make gui_open` and also `make floorplan`

Reduce log spam: don't log command that suppresses warning

### DIFF
--- a/flow/scripts/load.tcl
+++ b/flow/scripts/load.tcl
@@ -3,12 +3,6 @@ source $::env(SCRIPTS_DIR)/util.tcl
 source $::env(SCRIPTS_DIR)/report_metrics.tcl
 
 proc load_design { design_file sdc_file } {
-  # Source platform-related Tcl command (initially for suppressing Liberty
-  # warnings
-  if { [env_var_exists_and_non_empty PLATFORM_TCL] } {
-    log_cmd source $::env(PLATFORM_TCL)
-  }
-
   # Read liberty files
   source $::env(SCRIPTS_DIR)/read_liberty.tcl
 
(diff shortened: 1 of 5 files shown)


flow/scripts/load.tcl
```diff
@@ -3,12 +3,6 @@ source $::env(SCRIPTS_DIR)/util.tcl
 source $::env(SCRIPTS_DIR)/report_metrics.tcl
 
 proc load_design { design_file sdc_file } {
-  # Source platform-related Tcl command (initially for suppressing Liberty
-  # warnings
-  if { [env_var_exists_and_non_empty PLATFORM_TCL] } {
-    log_cmd source $::env(PLATFORM_TCL)
-  }
```
maliberty, line 10: I don't think this should be moved to read_liberty.tcl.  It is currently only used for that purpose but nothing limits it to that in the future.

oharboe: On the other hand, this is how it is today and it is easy to change once the future use case is understood.

maliberty: What is the advantage of the move?  The name PLATFORM_TCL doesn't suggest it is belongs there.  On my todo is to skip read_liberty when GUI_TIMING=0

oharboe: It is in a single place.

oharboe: fixed, less repetition via helper proc

oharboe:
@maliberty unrelated errors in pr-merge

maliberty (approved):

PR #3326, opened 2025-07-16
Support zsh as shell

Faholan:
Support ZSH when sourcing the env scripts


oharboe:
Isnt the BASH_SOURCE idiom used in many places?

Faholan:
Yes, it is used in other places, but those scripts are to be executed, and the shebang specifies that it should use bash.

Those two are the only scripts that were made to be sourced

maliberty (approved):

PR #3332, opened 2025-07-18
Make CTS_BUF_DISTANCE tunable

jeffng-or:


maliberty (approved):

flow/scripts/variables.yaml
```diff
@@ -769,11 +769,15 @@ CTS_BUF_DISTANCE:
...
     - cts
+  tunable: 1
+  type: float
 CTS_BUF_LIST:
   description: |
     List of cells used to construct the clock tree. Overrides buffer inference.
   stages:
     - cts
+  tunable: 1
+  type: float
```
oharboe, line 780: @jeffng-or @maliberty This is confusing. From the description, I get a sense that this is a list of strings, but it is type float

jeffng-or: Strange, that shouldn't have been part of the PR, so I'll remove it. It's not a tunable parameter.

maliberty: It will have to be in another PR as this one got merged.

PR #3333, opened 2025-07-19
variables: PRE_GLOBAL_ROUTE_TCL now has consistent naming

oharboe:


oharboe:
@maliberty good to merge?

maliberty:
Requires coordinating with a private PDK that uses this var

oharboe:
> Requires coordinating with a private PDK that uses this var

Ah. Leave open?

maliberty (approved):

PR #3345, opened 2025-07-22
Corrected checks for falsey variables

joecrop:
The existing TCL syntax of `!$` was incorrect. I used the syntax that I see in other files to check for falsey variables.

### DIFF
--- a/flow/scripts/global_route.tcl
+++ b/flow/scripts/global_route.tcl
@@ -23,9 +23,9 @@ proc global_route_helper { } {
 
   if { $result != 0 } {
     if {
-      [!$::env(GENERATE_ARTIFACTS_ON_FAILURE) || \
+        ![env_var_exists_and_non_empty GENERATE_ARTIFACTS_ON_FAILURE] || \
         ![file exists $::global_route_congestion_report] || \
-        [file size $::global_route_congestion_report] == 0]
+        [file size $::global_route_congestion_report] == 0
     } {
       write_db $::env(RESULTS_DIR)/5_1_grt-failed.odb
       error $errMsg
@@ -46,7 +46,7 @@ proc global_route_helper { } {
     set_dont_use $::env(DONT_USE_CELLS)
   }
 
-  if { !$::env(SKIP_INCREMENTAL_REPAIR) } {
+  if { ![env_var_exists_and_non_empty SKIP_INCREMENTAL_REPAIR] } {
     if { $::env(DETAILED_METRICS) } {
       report_metrics 5 "global route pre repair design"
     }


oharboe:
@maliberty Is this an AI generated PR?

Will the flood gates now open 😬

oharboe:
What symptoms did you see?

GENERATE_ARTIFACTS_ON_FAILURE is a tcl boolean variable  with default value 0, check out variables.yam

This PR needs work

joecrop:
I get an error message when `GENERATE_ARTIFACTS_ON_FAILURE` is set to 0:
```
Error: global_route.tcl, 115 invalid command name "!0"
```
It seems to me that the old syntax of `!$::env(GENERATE_ARTIFACTS_ON_FAILURE)` resolves to `!0` which the TCL interpreter doesn't know how to handle. So I am using the syntax I see in other files.

NOTE: I am not changing the logic here and not inverting any variable checks. I am just using a different syntax for the same check.

joecrop:
Oh, I see what you are saying now @oharboe. The check should not be for if the variable exists. It needs to be for if the variable is 1 or 0. So probably something like this is more appropriate.
```
[expr {!$::env(GENERATE_ARTIFACTS_ON_FAILURE)}]
```
I don't see this syntax used anywhere else in the flow tcl scripts... but perhaps there aren't any other places where a check for a variable == 0 exists elsewhere.

oharboe:
@joecrop FYI https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3346

oharboe:
> Oh, I see what you are saying now @oharboe. The check should not be for if the variable exists. It needs to be for if the variable is 1 or 0. So probably something like this is more appropriate.
> 
> ```
> [expr {!$::env(GENERATE_ARTIFACTS_ON_FAILURE)}]
> ```
> 
> I don't see this syntax used anywhere else in the flow tcl scripts... but perhaps there aren't any other places where a check for a variable == 0 exists elsewhere.

The .tcl style is probably a result of more copy and pasting than ruminating over finer points of readable .tcl...

The wrapping `[expr ...]` bit is superfluous here, the closure of the if statement is already an expression evaluation.

I decided to separate the concern of fixing the accidental regression when touching this [unusual](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) .tcl syntax and fixing this unsual syntax. Follow up PRs to #3346 that goes through the trouble of injecting and testing failure modes of global route and these variables greatly appreciated.

I think the existence/size check on the .drc file is a vestige. Especially now that drc are stored in the .odb

oharboe:
@joecrop close? See https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3347

oharboe:
Nice find!

PR #3346, opened 2025-07-22
global_route.tcl: fix bug introduced when formatting code

oharboe:
This reverts a change from 5810f536e

### DIFF
--- a/flow/scripts/global_route.tcl
+++ b/flow/scripts/global_route.tcl
@@ -23,7 +23,7 @@ proc global_route_helper { } {
 
   if { $result != 0 } {
     if {
-      [!$::env(GENERATE_ARTIFACTS_ON_FAILURE) || \
+      [expr !$::env(GENERATE_ARTIFACTS_ON_FAILURE) || \
         ![file exists $::global_route_congestion_report] || \
         [file size $::global_route_congestion_report] == 0]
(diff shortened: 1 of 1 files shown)


oharboe:
@luarss FYI

flow/scripts/global_route.tcl
```diff
@@ -23,7 +23,7 @@ proc global_route_helper { } {
 
   if { $result != 0 } {
     if {
-      [!$::env(GENERATE_ARTIFACTS_ON_FAILURE) || \
+      [expr !$::env(GENERATE_ARTIFACTS_ON_FAILURE) || \
```
maliberty, line 26: Do we need the expr here?  I think the problem is the outer [] are not needed.

oharboe: @luarss Aha!

Now we know what happened.

Fixing warnings is dangerous  :-) Removing the "expr" fixed the warning in this unusual .tcl syntax, but introduced a bug.

```
tclfmt 0.4.2
flow/scripts/global_route.tcl:26:7: unnecessary command substitution within expression [redundant-expr]
Error: Process completed with exit code 1.
```

oharboe:
need more cleanup to be compatible with tclfmt linting

PR #3348, opened 2025-07-23
fast_route: cleanup of vestiges

oharboe:
global_place.tcl now stores FASTROUTE_TCL/set_routing_layers in .odb so the need for the brittle fast_route utility fn is gone

### DIFF
--- a/flow/scripts/global_place.tcl
+++ b/flow/scripts/global_place.tcl
@@ -18,7 +18,14 @@ if { ![env_var_exists_and_non_empty FOOTPRINT] } {
   }
 }
 
-fast_route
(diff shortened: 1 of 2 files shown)


maliberty:
@eder-matheus are all elements of fastroute.tcl stored in odb?

eder-matheus:
> @eder-matheus are all elements of fastroute.tcl stored in odb?

Yes, both routing layer adjustments and the layer range for signal and clock nets are stored in ODB.

The layer adjustments are stored on each dbTechLayer object, while the layer ranges are stored in the dbBlock.

maliberty (approved):

PR #3349, opened 2025-07-23
Tie fanout helper cleanup

oharboe:


maliberty:
Its better if your PRs are more orthogonal.  This can't be merged until my question in #3348 is answered as you included it into this PR.

oharboe:
> Its better if your PRs are more orthogonal. This can't be merged until my question in #3348 is answered as you included it into this PR.

Indeed. It is not by choice that there is a dependency, there is a conflict.

I can create two conflicting PRs?

maliberty:
What would be the conflict?

oharboe:
> What would be the conflict?

Random rebase conflict. So now we either wait, or we increase the risk by trying to mess around with rebase. Fix one problem, create another.

When spelunking and fixing code, there's a non-trivial chance of rebase conflicts like this.

maliberty:
It will wait then

oharboe:
I'll start afresh without the conflict and a bugfix, not just code cleanup.

PR #3370, opened 2025-08-01
Added support for EQUIVALENCE_CHECK_USE_LIBS

Cronus-38:
By default, equivalence checking uses the verilog simulation models. The EQUIVALENCE_CHECK_USE_LIBS variable can be used to force the tool to instead do equivalence checking using .lib files.

Along with the code change to load.tcl I've included a test script at flow/tests/test_eqy_use_lib.sh.  The comments at the top of that file explain the usage of EQUIVALENCE_CHECK_USE_LIBS.

### DIFF
--- a/flow/scripts/load.tcl
+++ b/flow/scripts/load.tcl
@@ -58,8 +58,10 @@ proc load_design { design_file sdc_file } {
 # Routines to run equivalence tests when they are enabled.
 
 proc get_verilog_cells_for_design { } {
-  set dir "$::env(PLATFORM_DIR)/work_around_yosys/"
-  set cell_files [glob $dir/*.v]
+  if {![env_var_exists_and_non_empty EQUIVALENCE_CHECK_USE_LIBS] || [env_var_equals EQUIVALENCE_CHECK_USE_LIBS 0]} {
+    set dir "$::env(PLATFORM_DIR)/work_around_yosys/"
+    set cell_files [glob $dir/*.v]
+  }  
 }
 
 proc write_eqy_verilog { filename } {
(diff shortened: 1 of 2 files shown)


povik:
@Cronus-38 Thanks for opening this. Any new variables need to be documented in `flow/scripts/variables.yaml` but I think instead of introducing the new variable we could always use `.lib` definitions for equivalence checking. I see you have provided a way to override what the `.lib` should be instead of relying on `LIB_FILES`. What is the use case for the override?

Cronus-38:
@povik  Honestly, I can’t think of a scenario in which using the .lib files instead of the verilog sim models would cause a problem.  I was just giving users the ability to exercise direct control out of an abundance of paranoia.  If you’re confident that it won’t break anything to simply use the .lib files by default in all cases then I’m 100% cool with that!

From: Martin Povišer ***@***.***>
Sent: Friday, August 1, 2025 12:56 PM
To: The-OpenROAD-Project/OpenROAD-flow-scripts ***@***.***>
Cc: Dillard, Cronus [USA] ***@***.***>; Mention ***@***.***>
Subject: [External] Re: [The-OpenROAD-Project/OpenROAD-flow-scripts] Added support for EQUIVALENCE_CHECK_USE_LIBS (PR #3370)

povik left a comment (The-OpenROAD-Project/OpenROAD-flow-scripts#3370) @Cronus-38 Thanks for opening this. Any new variables need to be documented in flow/scripts/variables. yaml but I think instead of introducing the new variable we could always

[Image removed by sender.]povik left a comment (The-OpenROAD-Project/OpenROAD-flow-scripts#3370)<https://urldefense.us/v3/__https:/github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3370*issuecomment-3145624242__;Iw!!DJ_L93zn18-Ojg!nvRfNZ_VdovjbCSt3-jiN3-1tRnG5C2hWjQH99p4QkA-cRFBiJ-sL7mYYy_DCbcQx3vUTGZYKPUrD8afp9IUk3hPQX-CbIc$>

@Cronus-38<https://urldefense.us/v3/__https:/github.com/Cronus-38__;!!DJ_L93zn18-Ojg!nvRfNZ_VdovjbCSt3-jiN3-1tRnG5C2hWjQH99p4QkA-cRFBiJ-sL7mYYy_DCbcQx3vUTGZYKPUrD8afp9IUk3hPjwWYk9s$> Thanks for opening this. Any new variables need to be documented in flow/scripts/variables.yaml but I think instead of introducing the new variable we could always use .lib definitions for equivalence checking. I see you have provided a way to override what the .lib should be instead of relying on LIB_FILES. What is the use case for the override?

maliberty:
If all parties agree it is the right thing to just switch then I don't think we need a flag.  We can always add one later if a need arises.

Please fix the DCO (see the details link)

povik:
> If you’re confident that it won’t break anything to simply use the .lib files by default in all cases then I’m 100% cool with that!

I think we can do the switch and deal with any odd cases if they arise.

povik:
@Cronus-38 Without the flag the switch to `.lib` reduces to a simple change. I've staged it in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3377 and I'll close your PR in favor of the other one.

PR #3378, opened 2025-08-04
variables: syntax error check in env_var_equals

oharboe:
@jeffng-or Thoughts?


jeffng-or:
> @jeffng-or Thoughts?

This is verifying that a variable used in env_var_equals has been documented in variables.yaml? Sounds reasonable. 

Would you want to extend this to env_var_exists_and_non_empty in case someone fat fingers the env var there?. It looks like it would cover append_env_var and env_var_or_empty as well.

oharboe:
Yes, that was the idea. I'm not going to pick up this work now, but I though it would be nice to have in place and save us some more agony.

Feel free cherry pick this & continue if you find it useful.

PR #3379, opened 2025-08-04
placement: run global placement once if IO placement is provided

oharboe:
If the computer says yes?


maliberty:
I don't see a reason to merge gpl and ppl into a single step.  Just skip that step if needed.

oharboe:
> I don't see a reason to merge gpl and ppl into a single step.  Just skip that step if needed.

Is there any reason to run global placement skip io if pins are placed?

To me this is a single concern spread across two stages that should be one stage.

maliberty:
No but they are not always placed.

oharboe:
> No but they are not always placed.

Lets talk it over in a future call, no rush.

 There is some important information I am missing for this to make sense.

PR #3380, opened 2025-08-04
io: always write out 3_2_place_iop.tcl

oharboe:
@maliberty I just noticed this while spelunking.

Does it make more sense to always write out io placement?


oharboe:
https://github.com/The-OpenROAD-Project/OpenROAD/issues/7954

oharboe:
@maliberty unrelated CI failures, OpenROAD is fixed on master, ORFS needs an upgrade of OpenROAD

oharboe:
Doesn't work when there's a FLOORPLAN_DEF, so I assume it isn't possible to figure out the pin placements in this case?

```
nangate45 aes (base)
  Last log file 3_2_place_iop.log
  Found 1 errors in the logs.
      [ERROR PPL-0113] Block is not allowed to be null.
```

PR #3387, opened 2025-08-09
Remove duplicate instructions in pre-built install docs

kcaisley:
Hi, just was reading the docs and noticed that the section for installing pre-built binaries seemed to have the same section on Klayout and Yosys dependencies duplicated. Just figured I'd help clean it up for clarity.


maliberty (approved):
Thanks!

PR #3403, opened 2025-08-14
Add global variables for repair antennas iterations

luis201420:
Creating global variables for the maximum number of repair antenna iterations in GRT and DRT.

designs/sky130hs/jpeg/rules-base.json updates:
| globalroute__antenna_diodes_count             |       87 |      141 | Failing  |
| detailedroute__antenna__violating__nets       |        1 |        0 | Tighten  |
| detailedroute__antenna_diodes_count           |       86 |      154 | Failing  |


eder-matheus (approved):

maliberty:
Please note the merge conflict

maliberty:
I'm confused why we introduced MAX_REPAIR_ANTENNAS_ITER_GRT.  It has no documentation and it appears we won't do any antenna repair if it is not set.  Also there is @jeffng-or 's problem

eder-matheus:
> I'm confused why we introduced MAX_REPAIR_ANTENNAS_ITER_GRT. It has no documentation and it appears we won't do any antenna repair if it is not set. Also there is @jeffng-or 's problem

@maliberty We first added both MAX_REPAIR_ANTENNAS_ITER_GRT and MAX_REPAIR_ANTENNAS_ITER_DRT because one design had 1 antenna violation at the end of the 5 iterations, but running a new iteration fixed it. So the idea was to allow design-specific configuration regarding how many iteration it will run to remove the violations.

The missing documentation is my fault. If you believe it's worth to keep these options, I can update the docs.

What problem Jeff is facing related to this PR? I didn't find an specific issue mentioning repair_antennas.

maliberty:
See https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/f8876928b65151076e4eb7edf5b5d0ff413ba04b#r164081818

There should be a default for the variable of 5 rather than embedding it into each platform.

eder-matheus:
> See [f887692#r164081818](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/f8876928b65151076e4eb7edf5b5d0ff413ba04b#r164081818)
> 
> There should be a default for the variable of 5 rather than embedding it into each platform.

Ok, I'll do the fixes now.

rovinski:
Why not continue running iterations until the number of violations no longer decreases? And the user can set an iteration cap if they want.

PR #3454, opened 2025-08-31
make: issue, fix regression

oharboe:
with bazel-orfs, I make issues for a specific stage at which point there's no requirement that there is a dependency on yosys AND openroad, so checking if yosys exists for a make io_placement_issue fails, for example.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -182,7 +182,7 @@ $(foreach block,$(BLOCKS),$(eval $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKN
 # Utility to print tool version information
 #-------------------------------------------------------------------------------
 .PHONY: versions.txt
-versions.txt: check-yosys check-openroad
+versions.txt:
 	mkdir -p $(OBJECTS_DIR)
 	@echo "yosys $(shell $(YOSYS_EXE) -V 2>&1)" > $(OBJECTS_DIR)/$@
 	@echo "openroad $(shell $(OPENROAD_EXE) -version 2>&1)" >> $(OBJECTS_DIR)/$@


flow/Makefile
```diff
@@ -182,7 +182,7 @@ $(foreach block,$(BLOCKS),$(eval $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKN
 # Utility to print tool version information
 #-------------------------------------------------------------------------------
 .PHONY: versions.txt
-versions.txt: check-yosys check-openroad
+versions.txt:
 	mkdir -p $(OBJECTS_DIR)
 	@echo "yosys $(shell $(YOSYS_EXE) -V 2>&1)" > $(OBJECTS_DIR)/$@
 	@echo "openroad $(shell $(OPENROAD_EXE) -version 2>&1)" >> $(OBJECTS_DIR)/$@
```
maliberty, line 188: How can you write out the version of a tool if the tool doesn't exist?

oharboe: Don't have to.

For the make issue in question, when the tool is not needed, a note is added in versions.txt that an unused tool is not available..

maliberty: They had to have been actually set when you ran the design in order to produce the issue.  Why wouldn't we want to capture that info in the issue?

oharboe: It isnt immediately available to be captured + it is evidently never used: so far no one has asked me for yosys info absent in an OR bug.

oharboe: I can silence this with a dummy yosys cmd, no one would ever notice :-)

maliberty: You could add `yosys -V` to synth.sh so it goes in the log file.  OR already prints it version.  Something similar for klayout and we can just remove this target.

oharboe: Yes, keep it simple...


So remove versions.txt when all tools log version?

oharboe:
moot https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3458

PR #3469, opened 2025-09-03
remove read_macro_placement and use MPL commands instead

AcKoucher:


eder-matheus (approved):
I agree with this change, but I'll let @maliberty give his review too.

PR #3472, opened 2025-09-04
allow macro preplacement and make ngt45/bp our first test case with partial mpl

AcKoucher:
Requires [#8099](https://github.com/The-OpenROAD-Project/OpenROAD/pull/8099) and #3481 

Modify scripts in order to allow macro pre-placement and make ngt45/bp our first test case for both fixed and unfixed macros.


flow/scripts/macro_place_util.tcl
```diff
@@ -36,40 +36,40 @@ if { [find_macros] != "" } {
...
-      set all_args $::env(RTLMP_ARGS)
-    }
+  set all_args $additional_rtlmp_args
 
-    log_cmd rtl_macro_placer {*}$all_args
+  if { [env_var_exists_and_non_empty RTLMP_ARGS] } {
+    set all_args $::env(RTLMP_ARGS)
   }
 
+  log_cmd rtl_macro_placer {*}$all_args
```
maliberty, line 71: If all the macros are pre-placed, will rtl_macro_placer just return immediately?

AcKoucher: If the all the macros are pre-placed it will return from C++, not tcl:

The call order is something like:

```
rtl_macro_placer

HierRTLMP::run()

ClusteringEngine::run()

    Metrics* ClusteringEngine::computeModuleMetrics(odb::dbModule* module) // Grab data from modules

    ClusteringEngine::init() // Log data

    // If no unplaced macros are found, we return at this point.
```

AcKoucher:
If the changes here are ok I can update the OR submodule before we merge it so that the CI doesn't break.

maliberty:
go ahead and update

maliberty (approved):

flow/designs/nangate45/black_parrot/config.mk
```diff
@@ -31,6 +31,8 @@ export IO_CONSTRAINTS = $(DESIGN_HOME)/$(PLATFORM)/$(DESIGN_NAME)/io.tcl
 
 export PLACE_DENSITY_LB_ADDON = 0.05
 
+export MACRO_PLACEMENT_TCL = $(DESIGN_HOME)/$(PLATFORM)/$(DESIGN_NICKNAME)/macro_placement.tcl
```
maliberty:
```suggestion
export MACRO_PLACEMENT_TCL = $(DESIGN_HOME)/$(PLATFORM)/$(DESIGN_NAME)/macro_placement.tcl
```

maliberty (approved):

PR #3499, opened 2025-09-17
Document floorplanning methods and flow variables

kcaisley:
Hello,

I am working on building a top-level layout with a pad ring, and noticed that the four alternatives for specifying top level footprints in `floorplan.tcl` weren't highlighted in the docs. Also important, `FOOTPRINT_TCL` was not mentioned in the flow variables .md doc. It's used in several of the design examples, so I figured it should be added.

Thanks!


maliberty (approved):

maliberty:
Thanks!

PR #3562, opened 2025-10-09
ihp-sg13g2: Sync PDK

dnltz:
- Sync the PDK with the latest version from the PDK.
- Include the sg13g2.map file to the update script.
- Change the metal routing direction to match changes made in the PDK.


flow/platforms/ihp-sg13g2/config.mk
```diff
@@ -86,8 +86,8 @@ export MATCH_CELL_FOOTPRINT = 1
 export PLACE_SITE = CoreSite
 
 # IO Placer pin layers
-export IO_PLACER_H ?= Metal2
-export IO_PLACER_V ?= Metal3
+export IO_PLACER_V ?= Metal2
+export IO_PLACER_H ?= Metal3
```
dnltz, line 90: These changes are important to match the changed routing direction.

maliberty:
There are lot of design failures.  Please say more about the layer direction changes.  I suspect the pdn setup will also need adjusting.

dnltz:
> There are lot of design failures. Please say more about the layer direction changes. I suspect the pdn setup will also need adjusting.

@maliberty Yes, I'm working on this issue right now.

I opened this PR to run all designs in the CI environment (making sure my OpenROAD set up is not broken) and in case I need to open an issue, I can link this PR.

Do you have a "Do not merge" label?

and sorry for misusing the PR flow :)

maliberty:
Usually people just put that in the title temporarily, eg [DO NOT MERGE]

flow/designs/ihp-sg13g2/aes/rules-base.json
```diff
@@ -93,7 +93,7 @@
         "compare": "<="
     },
     "detailedroute__route__drc_errors": {
-        "value": 0,
+        "value": 4,
```
dnltz: @maliberty I had to move the PDN from M5+TM1 to TM1+TM2. Since IHP changed the routing direction, M5 had the same orientation as M1 and they can't be connected.

Some designs now have `route DRC errors`. I assume they need to be fixed, right?

maliberty: I think so as it routed cleanly before.  What do they look like?

dnltz: image

The pin placer is completely ignoring those vias between Metal1 and the power ring. I will try to move the power ring a  little bit away from the edge.

dnltz:
@maliberty can you please remove the `do not merge` label and review this PR?

I had to update some designs because IHP updated the standard cells.

dnltz:
@sergeiandreyev @KrzysztofHerman FYI bumping the PDK here to update the platform config to match with changed metal routing direction.

maliberty (approved):

maliberty:
Thanks!

PR #3592, opened 2025-10-16
add target to run floorplan to place in a single openroad run

eder-matheus:


maliberty (approved):

oharboe:
Curious....

Why the copy and paste instead of re-use of .tcl?

What is the motivation for this feature?

eder-matheus:
> What is the motivation for this feature?

This is for an internal project and will probably not be used in the main flow. I didn't reuse the existing Tcl to avoid adding extra checks when loading the design.

Of course it has room for improvement, but the way it is now is enough for our needs.

oharboe:
Interesting I have been pondering if the current scripts can be refactored in some way so that it is a user option to decide how many openroad invocations and how many interim .odb/.sdc files to write out.

maliberty:
The old single flow that you removed could be reinstated.  In some PDKs the time to reload the Liberty + odb makes it worth considering doing that.

oharboe:
> The old single flow that you removed could be reinstated.  In some PDKs the time to reload the Liberty + odb makes it worth considering doing that.

It would be neat, but it needs to come with a refactoring for CI testability and maintainability to my mind.

oharboe:
How about something like this?

It uses automated duplicatin/editing of the script. Not super-pretty, but all these stage scripts have a very similar pattern so they lend themselves to this approach.

1. Fix up the .tcl code below to load the .tcl stage files and strip the load and the save .odb 
2. Now create a single large script using these procs
3. Load the starting .odb/.v/.sdc
4. Use tcl eval to run stages
5. Save the .odb/.v/.sdc as desired

```tcl
proc strip_lines_by_patterns {lines start end patterns} {
    set matched_lines {}
    set to_remove {}
    for {set i $start} {$i <= $end} {incr i} {
        set line [string trim [lindex $lines $i]]
        set matched 0
        foreach pattern $patterns {
            if {[regexp $pattern $line]} {
                set matched 1
                lappend matched_lines $line
                lappend to_remove $i
                break
            }
        }
        if {!$matched} {
            puts "Error: Expected line not found: $line"
            exit 1
        }
    }
    # Remove matched lines by index
    set stripped_lines $lines
    # Remove from highest to lowest index to avoid shifting
    foreach idx [lsort -integer -decreasing $to_remove] {
        set stripped_lines [lreplace $stripped_lines $idx $idx]
    }
    return [list $stripped_lines $matched_lines]
}

# Strip load from and return the stripped script along with removed lines to write the .odb and .sdc
proc strip_stage{script} {
    set f [open $script r]
    set lines [split [read $f] "\n"]
    close $f

    set patterns {
        {^source\s+\$::env\(SCRIPTS_DIR\)/load\.tcl\s*$}
        {^erase_non_stage_variables\s.*$}
        {^load_design\s.*$}
    }

    set result [strip_lines_by_patterns $lines 1 5 $patterns]
    set lines [lindex $result 0]
    set removed_lines [lindex $result 1]

    # Remove lines at end and keep them
    set end_patterns {
        {^write_db\s+\$::env\(RESULTS_DIR\)/2_1_floorplan\.odb\s*$}
        {^write_sdc\s+-no_timestamp\s+\$::env\(RESULTS_DIR\)/2_1_floorplan\.sdc\s*$}
    }

    set result [strip_lines_by_patterns $lines [expr {[llength $lines] - 6}] [expr {[llength $lines] - 1}] $end_patterns]

    return [list [join [lindex $result 0] "\n"] $removed_lines [lindex $result 1]]
}

**Comment by maliberty** [comment]
The old approach seems simpler than this.

**Comment by oharboe** [comment]
> The old approach seems simpler than this.

I recall some code there, but I looked at pdn.tcl history and couldn't find it...

pdn.tcl is tiny, so it is a nice use-case to rewrite to some new form that is usable from both the current and a single invocation flow so that we can get an idea of what dual use-case articulation of these scripts should look like...

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/b14dd6d766b8269dc459546ad29206ff2f494ad9/flow/scripts/pdn.tcl#L1-L21

**Comment by maliberty** [comment]
image

**Comment by oharboe** [comment]
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1740 can be implemented with much less iffy code... We should be able to articulate each of these scripts so that they can be invoked in both use-cases without iffy code.

**Comment by maliberty** [comment]
Please explain what you have in mind

**Comment by oharboe** [comment]
> Please explain what you have in mind

Look at [pdn.tcl](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1740/files#diff-3ce00b500c0c2ff2e6fdc5c981d9e6fae76e64a09f82ffec8d9267678a5f5ed8) in the old implementation, write_db/verilog/sdc are conditionally written out.

So why not override the behavior of some existing procedures for the multi-stage use-case?

- load_design - loads a design if one isn't already loaded, no change to header of pdn.tcl
- write_db/verilog/sdc - remembers the arguments of the last invocation, but does not save anything. We can write out a final version using this information with a new write_outputs tcl proc.
- erase_non_stage_variables - stubbed out

So a multistage script would look like:

```tcl

source  $::env(SCRIPTS_DIR)/multi-stage.tcl

# Any number of scripts invoked here
source $::env(SCRIPTS_DIR)/floorplan.tcl
source $::env(SCRIPTS_DIR)/....tcl
source $::env(SCRIPTS_DIR)/pdn.tcl

# Writes the .sdc/.odb/.v file names recorded in write_db/verilog/sdc using the current state
write_outputs

```

maliberty:
I'll want the option to write per stage outputs even if running a single run so I can restart at any point of my choosing later.   The looks remarkably like the old all.tcl so I'm not seeing the difference.

The conditional load is a nice trick though it is still an if.

oharboe:
I prefer bazel at that point

maliberty:
Bazel has the same problem of restarting at every stage and having to reload the data.

oharboe:
> Bazel has the same problem of restarting at every stage and having to reload the data.

I find that do- handles that use case for me the few times it matters.

oharboe:
I think your use case is supportable with scripts that are identical for both use cases with injected iffy, rather than spelled out iffy code.

My use case is bazel, my stake in this is to have maintainable ORFS scripts over the supported use cases.

PR #3602, opened 2025-10-19
bazel: add bazel build hack tip

oharboe:


docs/user/BuildLocally.md
```diff
@@ -12,6 +12,21 @@ cd OpenROAD-flow-scripts
...
 
+## Using Bazel to build OpenROAD and run the ORFS flow
+
+Long story short: OpenROAD will eventually switch to using Bazel for downloading dependencies and building OpenROAD for all the reasons that the DependencyInstaller.sh and cmake are hard to support and brittle across platforms.
+
+Currently the simplest way to build OpenROAD and run ORFS is to run one test, which will download all OpenROAD dependencies and build OpenROAD in the exec configuration:
+
+``` shell
+cd tools/OpenROAD
+bazelisk test src/drt/...
```
maliberty, line 23: why test drt?

docs/user/BuildLocally.md
```diff
@@ -12,6 +12,21 @@ cd OpenROAD-flow-scripts
...
+cd ../flow
```
maliberty: this won't work if you are in tools/OpenROAD

docs/user/BuildLocally.md
oharboe: the whole exec vs  target configuration business, the target config fails to build sometimes + it jells w a local modify + test cycle

oharboe: there is no explicit way in bazel to build an exec version of a binary, it is built because the exec binary is used by another rule

docs/user/BuildLocally.md
oharboe: fixed

oharboe:
@maliberty Good to go? 

pr-merge failure is an unrelated CI error.

maliberty (approved):

PR #3655, opened 2025-11-13
Add flow variable BUFFER_PORTS_DONT_USE_CELLS

chrisgelinek-rl:
This PR makes the suggestion in the following comment usable from a design's `config.mk`:
https://github.com/The-OpenROAD-Project/OpenROAD/discussions/7757#discussioncomment-13713371

For example, if the cell library has some delay cells that shouldn't be used for port buffering, add a line

    export BUFFER_PORTS_DONT_USE_CELLS = DLY_CELL1 DLY_CELL2

to your `config.mk` and these cells will be excluded during the `buffer_ports` call within *flow/scripts/global_place.tcl*.

### DIFF
--- a/docs/user/FlowVariables.md
+++ b/docs/user/FlowVariables.md
@@ -102,6 +102,7 @@ configuration file.
 | <a name="ADDITIONAL_LIBS"></a>ADDITIONAL_LIBS| Hardened macro library files listed here. The library information is immutable and used throughout all stages. Not stored in the .odb file.| |
 | <a name="BALANCE_ROWS"></a>BALANCE_ROWS| Balance rows during placement.| 0|
 | <a name="BLOCKS"></a>BLOCKS| Blocks used as hard macros in a hierarchical flow. Do note that you have to specify block-specific inputs file in the directory mentioned by Makefile.| |
+| <a name="BUFFER_PORTS_DONT_USE_CELLS"></a>BUFFER_PORTS_DONT_USE_CELLS| List of cells that should not be used for buffering of input/output ports during floorplanning.| |
 | <a name="CAP_MARGIN"></a>CAP_MARGIN| Specifies a capacitance margin when fixing max capacitance violations. This option allows you to overfix.| |
 | <a name="CDL_FILES"></a>CDL_FILES| Insert additional Circuit Description Language (`.cdl`) netlist files.| |
 | <a name="CELL_PAD_IN_SITES_DETAIL_PLACEMENT"></a>CELL_PAD_IN_SITES_DETAIL_PLACEMENT| Cell padding on both sides in site widths to ease routability in detail placement.| 0|
@@ -369,6 +370,7 @@ configuration file.
 ## place variables
 
 - [BALANCE_ROWS](#BALANCE_ROWS)
+- [BUFFER_PORTS_DONT_USE_CELLS](#BUFFER_PORTS_DONT_USE_CELLS)
 - [CELL_PAD_IN_SITES_DETAIL_PLACEMENT](#CELL_PAD_IN_SITES_DETAIL_PLACEMENT)
 - [CELL_PAD_IN_SITES_GLOBAL_PLACEMENT](#CELL_PAD_IN_SITES_GLOBAL_PLACEMENT)
 - [CLUSTER_FLOPS](#CLUSTER_FLOPS)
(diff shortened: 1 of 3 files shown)


povik:
In the linked discussion @jhkim-pii points out

> buffer_ports -buffer_cell <buf_cell> has been implemented in the PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/7845 . You can use it now.

I prefer using the `-buffer_cell` option to amending the dont use list.

maliberty:
I think the same.  It seems more work to list dont use cells than just to pick one.

Note that we try to avoid delay cells via foot prints if they are available.

chrisgelinek-rl:
Thanks for your comments. Some thoughts:

 1. There doesn't appear to be an ORFS variable to set the `-buffer_cell` option in the *global_place.tcl* file.
 2. If there was, wouldn't we limit ourselves to having only a single type of buffer for all outputs, even if the loads of some outputs were different, such as when they're specified using `set_load`, or am I missing something?

maliberty:
The idea is to add an option for the buffer cell to ORFS rather than add a dont-use list otion.

buffer_ports will use a single cell for all ports.  Later repair_design/timing may adjust its size.

chrisgelinek-rl:
>  Later repair_design/timing may adjust its size.

Ah, I see - and I agree with you, then it makes more sense to change the strategy to use the `-buffer_cell` option.

Before I go ahead and close this PR in favour of one using `-buffer_cell`, I feel there are quite a few possible combinations of options to the `buffer_ports` call in *global_place.tcl*.

How about a new variable `BUFFER_PORTS_COMMAND` defaulting to just `buffer_ports` but allowing designs to use more specific and interesting ones, such as:
 1. `export BUFFER_PORTS_COMMAND = buffer_ports -buffer_cell BUFX1` (our main use case at the moment),
 2. `export BUFFER_PORTS_COMMAND = buffer_ports -verbose` (if one's interested in what it is doing)
 3. `export BUFFER_PORTS_COMMAND = buffer_ports -inputs -buffer_cell BUFX1 ; buffer_ports -outputs -buffer_cell INVX1` (in case we want different cells for input/output)

What do you think?

maliberty:
We already have

variables.yaml:432:PLACE_PINS_ARGS:
variables.yaml:463:GLOBAL_PLACEMENT_ARGS:
variables.yaml:495:CTS_ARGS:
variables.yaml:637:DETAILED_ROUTE_ARGS:
variables.yaml:758:SYNTH_ARGS:
variables.yaml:884:GLOBAL_ROUTE_ARGS:
variables.yaml:1016:RTLMP_ARGS:

which lets you add additional arguments to the command.  BUFFER_PORTS_ARGS would match the pattern.

chrisgelinek-rl:
I'm aware of those. The only one use case I'd be hesitant about would be `-input` and `-output`, unless they can be combined in a single `buffer_ports` call such as

    buffer_ports -inputs -buffer_cell INVX1 -outputs -buffer_cell BUFX1

...would that work?

Honestly I'm not even sure if specifying inverters as buffer cells makes sense, is `buffer_ports` aware of the logic sense at all?

Alternatively, maybe for more custom/complex use cases (such as `-inputs`) there should be another variable `GLOBAL_PLACE_TCL` similar to the following ones

```
150:PLATFORM_TCL:
281:TAPCELL_TCL:
292:MACRO_PLACEMENT_TCL:
312:PDN_TCL:
577:SET_RC_TCL:
780:POST_CTS_TCL:
785:FASTROUTE_TCL:
```

maliberty:
Its **buffer** ports - you should use a buffer.

chrisgelinek-rl:
What about the `GLOBAL_PLACE_TCL` variable vs. specifying both `-inputs` and `-outputs` options in a single `buffer_ports` call?

maliberty:
Why do you need different input and output buffers?  Even with your original dont-use idea you would get a single buffer type.

maliberty:
I don't know how GLOBAL_PLACE_TCL  got into the discussion of buffer_ports.

chrisgelinek-rl:
If the `-inputs` and `-outputs` options are only used to *avoid* inserting buffers for the respective other, you're right, we don't need GLOBAL_PLACE_TCL.

maliberty:
In theory you could do what you suggest but I've never even seen someone use those flags at all.

PR #3680, opened 2025-11-27
Add `gpl` min/max phi coefficients

### LINKED ISSUE #3385: Make global placement min/max phi coef tunable
### Description

Create enh request to track discussion from staff meeting.

### Suggested Solution

Steps:
1. Global_placement’s -min_phi_coef and -max_phi_coef are not exposed as flow variables, so we’d need to update global_place.tcl to expose them. Would we want to have the flow variables have a name that’s more user-friendly (e.g. MIN_PLACE_STEP_COEFF, MAX_PLACE_STEP_COEFF”
2. Add a check in global_place.tcl that ensures that MIN < MAX (covers if calling via ORFS)
3. Document the flow variables and mark them as tunable
4. Update AutoTuner _is_valid_config() to automatically flag if MIN > MAX (covers if AT chooses an invalid config and avoid running anything for an invalid config)

### Additional Context

_No response_

luarss:
Fixes #3385

### DIFF
--- a/flow/scripts/global_place.tcl
+++ b/flow/scripts/global_place.tcl
@@ -33,6 +33,25 @@ if { $::env(GPL_TIMING_DRIVEN) } {
   }
 }
 
+# Parameters for phi coefficients in global placement
+# Validate phi coefficient constraints before using them
+if { [info exists ::env(MIN_PLACE_STEP_COEF)] && [info exists ::env(MAX_PLACE_STEP_COEF)] } {
+  set min_phi $::env(MIN_PLACE_STEP_COEF)
+  set max_phi $::env(MAX_PLACE_STEP_COEF)
+
+  if { $min_phi > $max_phi } {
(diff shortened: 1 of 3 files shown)


flow/scripts/global_place.tcl
```diff
@@ -33,6 +33,30 @@ if { $::env(GPL_TIMING_DRIVEN) } {
...
+      "MIN_PLACE_STEP_COEF ($min_phi) cannot be greater than \
+MAX_PLACE_STEP_COEF ($max_phi)"
+  }
+}
+
+if { [info exists ::env(MIN_PLACE_STEP_COEF)] } {
+  lappend global_placement_args -min_phi_coef $::env(MIN_PLACE_STEP_COEF)
+}
+
+if { [info exists ::env(MAX_PLACE_STEP_COEF)] } {
```
oharboe: with default values, no need for exists checks; simplify code

vvbandeira (approved):

PR #3690, opened 2025-12-05
[DNM] platforms: ihpsg13g2: Add IO cells to DONT_USE_CELLS

dnltz:
Yosys started to add the IHP IO cells as standard cells and global placement fails with:
  [ERROR RSZ-2001] failed bnet construction for _078639_/p2c

Add 'sg13g2_IOPad*' to DONT_USE_CELLS to make sure Yosys won't use those cells.

### DIFF
--- a/flow/platforms/ihp-sg13g2/config.mk
+++ b/flow/platforms/ihp-sg13g2/config.mk
@@ -41,7 +41,8 @@ export DONT_USE_CELLS += \
 sg13g2_lgcp_1 \
 sg13g2_sighold \
 sg13g2_slgcp_1 \
-sg13g2_dfrbp_2
+sg13g2_dfrbp_2 \
+sg13g2_IOPad*
 
 
 # Define fill cells


dnltz:
@maliberty It fells a little bit hacky, but for some reason Yoysy is using IO cells with the latest IHP PDK.

rovinski:
IHP should probably mark I/O cells as dont_use in the liberty file itself. I/O cells shouldn't be used unless they're explicitly instantiated in the RTL.

maliberty:
@rovinski is correct, I see this in other PDKs.

dnltz:
Thanks @rovinski

I can see the `dont_use` property in the `sg13g2_io_dummy.lib` file, but it's missing in the typical, slow and fast files. I opened a pull-request to fix this in the PDK: https://github.com/IHP-GmbH/IHP-Open-PDK/pull/755


Will convert this PR to a draft and close as soon.

dnltz:
Fixed in the PDK. Sorry for bothering you and thanks for the hint!

PR #3728, opened 2025-12-23
LEC check

nanocoh:


nanocoh:
Hi Matt,

I saw this mail but i dont see any comment in the PR.
But yes, tbb is needed and added in etc/DependencyInstaller.sh
<https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3728#discussion_r2646292383>
for
the multithreading implementation for high speed.

maliberty:
I installed the dependencies but build.sh fails with:
```
-- Installing: /usr/local/lib/libyosys_liberty.a
CMake Error at thirdparty/naja/thirdparty/yosys-liberty/src/cmake_install.cmake:46 (file):
  file INSTALL cannot copy file
  "/workspace/w7/tools/kepler-formal/build/thirdparty/naja/thirdparty/yosys-liberty/src/libyosys_liberty.a"
  to "/usr/local/lib/libyosys_liberty.a": Permission denied.
  thirdparty/naja/thirdparty/yosys-liberty/cmake_install.cmake:47 (include)
  thirdparty/naja/thirdparty/cmake_install.cmake:47 (include)
  thirdparty/naja/cmake_install.cmake:47 (include)
  thirdparty/cmake_install.cmake:47 (include)
  cmake_install.cmake:52 (include)
```
You should not be installing into /usr/local

maliberty:
Please see INSTALL_PATH in build_openroad.sh.

nanocoh:
Hi @maliberty ,

"I installed the dependencies but build.sh fails with:

-- Installing: /usr/local/lib/libyosys_liberty.a
CMake Error at thirdparty/naja/thirdparty/yosys-liberty/src/cmake_install.cmake:46 (file):
  file INSTALL cannot copy file
  "/workspace/w7/tools/kepler-formal/build/thirdparty/naja/thirdparty/yosys-liberty/src/libyosys_liberty.a"
  to "/usr/local/lib/libyosys_liberty.a": Permission denied.
  thirdparty/naja/thirdparty/yosys-liberty/cmake_install.cmake:47 (include)
  thirdparty/naja/thirdparty/cmake_install.cmake:47 (include)
  thirdparty/naja/cmake_install.cmake:47 (include)
  thirdparty/cmake_install.cmake:47 (include)
  cmake_install.cmake:52 (include)
You should not be installing into /usr/local
Please see INSTALL_PATH in build_openroad.sh."

Can you check again with the last commit? I have used INSTALL_PATH for the installation as you instructed.

maliberty:
It installs but now I see:
```
kepler-formal 
kepler-formal: error while loading shared libraries: libnaja_python.so: cannot open shared object file: No such file or directory
```
env.sh doesn't setup the path to find the .so.   Fwiw it is typically easier if you just statically link the binary and side step this.

maliberty:
I manually fixed the above and ran `make LEC=1` but it fails
```
Error: cts.tcl, 83 0
child process exited abnormally
```
That's because in run_lec_test you do the equivalent of:
```
% grep -c "Found difference" logs/nangate45/gcd/base/4_lec_check.log 
0
% echo $?
1
```
gemini suggests
```
  try {
    set count [exec grep -c "Found difference" $::env(LOG_DIR)/4_lec_check.log]
  } trap CHILDSTATUS {results options} {
    # This block executes if grep returns a non-zero exit code
    set count 0
  }
```
instead

maliberty:
I don't know why the eqy code is in load.tcl but I think it is time to split out a lec.tcl for lec & eqy.

nanocoh:
For this comment: 

> I manually fixed the above and ran `make LEC=1` but it fails
> 
> ```
> Error: cts.tcl, 83 0
> child process exited abnormally
> ```
> 
> That's because in run_lec_test you do the equivalent of:
> 
> ```
> % grep -c "Found difference" logs/nangate45/gcd/base/4_lec_check.log 
> 0
> % echo $?
> 1
> ```
> 
> gemini suggests
> 
> ```
>   try {
>     set count [exec grep -c "Found difference" $::env(LOG_DIR)/4_lec_check.log]
>   } trap CHILDSTATUS {results options} {
>     # This block executes if grep returns a non-zero exit code
>     set count 0
>   }
> ```
> 
> instead

I done it intentionally in order to fail the compile in case let_check is failing. I followed the same process as the eqy test that s already there. do you prefer for it to pass in case the lec test find changes?

nanocoh:
> I don't know why the eqy code is in load.tcl but I think it is time to split out a lec.tcl for lec & eqy.

You are completely right. I done it as I saw the eqy test there but better to separate. done with latest commits in this pr.

This change will also allow to use it in the general case.

nanocoh:
> It installs but now I see:
> 
> ```
> kepler-formal 
> kepler-formal: error while loading shared libraries: libnaja_python.so: cannot open shared object file: No such file or directory
> ```
> 
> env.sh doesn't setup the path to find the .so. Fwiw it is typically easier if you just statically link the binary and side step this.

Are you sure you are using the update bin under install or the one that might have been previously installed in /usr/lib before the fixes?

maliberty:
| Are you sure you are using the update bin under install or the one that might have been previously installed in /usr/lib before the fixes?

I don't have write to /usr/lib.   It is the current one:
```
% ldd /workspace/w7/tools/install/kepler-formal/bin/kepler-formal |grep naja
	libnaja_python.so => /workspace/w7/tools/install/kepler-formal/lib/libnaja_python.so (0x00007f8643e74000)
```

maliberty:
lec is not failing when errors=0 so why should the flow fail?

maliberty:
You need to run tclfmt on flow/scripts/lec_check.tcl

nanocoh:
> You need to run tclfmt on flow/scripts/lec_check.tcl

done

maliberty (approved):

povik:
I'm happy to see this landing. We need to do something about the options as both "EQUIVALENCE_CHECK" and "LEC_CHECK" now enable a LEC check using different tools.

maliberty:
I plan to migrate over in another pr

PR #3740, opened 2025-12-30
flow: faster, leaner single openroad invocation flows

oharboe:
no longer save out all .odb/.sdc files automatically, the user can save out as many or as few as they need.


maliberty:
> no longer save out all .odb/.sdc files automatically

I don't see a change in the default flow.  I suppose you mean it in the context of the SINGLE_OPENROAD_INVOCATION_FLOW.   

SINGLE_OPENROAD_INVOCATION_FLOW doesn't actually control the invocation of OR so it feels odd.  Further you might want a single invocation flow to write out an .odb at each step even if it doesn't exit, so this feels orthogonal.

oharboe:
> > no longer save out all .odb/.sdc files automatically
> 
> I don't see a change in the default flow. I suppose you mean it in the context of the SINGLE_OPENROAD_INVOCATION_FLOW.
> 
> SINGLE_OPENROAD_INVOCATION_FLOW doesn't actually control the invocation of OR so it feels odd. Further you might want a single invocation flow to write out an .odb at each step even if it doesn't exit, so this feels orthogonal.

Fixed.

flow/scripts/floorplan_to_place.tcl
```diff
@@ -1,5 +1,6 @@
 # Enable keeping variables between stages
 set ::env(KEEP_VARS) 1
+set ::env(WRITE_ODB_AND_SDC_EACH_STAGE) 0
```
maliberty: This will prevent the user from overriding it.  I think this should only be set if currently unset.

oharboe: Consistent with other variables, I have added default values in variables.yaml as a perference, so if I check if these values are set, they always are. Hence, I set them to explicitly here, which I think is the intent of the script.

Do you have any more information about the use-case of this script, I'm guessing here.

maliberty: The intent of the script is to perform all operations in a single invocation.  Whether to write odb at each stage or not is a separate decision.  @eder-matheus does this matter for you?

eder-matheus: In the scenario I'm using it, I'm interested on the ODB files of each stage. However, I'm not against having it as an option, and the default behavior would be not writing the ODBs of each stage.

oharboe:
@eder-matheus @maliberty Should be all good now. Anything else?

maliberty (approved):

PR #3765, opened 2026-01-07
synth: fix operator swapping check that hierarchical is enabled

oharboe:
Separation of concerns: require OPENROAD_HIERARCHICAL to be set, but don't set it implicitly.

This fixes suprising potential bug w.r.t. e.g SYNTH_ variables not being set outside of the synthesis stage


oharboe:
@vvbandeira @maliberty unrelated autotuner failure in CI:

image

oharboe:
@maliberty Should be good to go, passes now after some false positive CI hickups.

maliberty:
We had an API key expire which is now fixed.

maliberty:
@precisionmoon any concern with the change?  It seems reasonable to me.

precisionmoon:
> @precisionmoon any concern with the change? It seems reasonable to me.
This looks fine.

maliberty:
@precisionmoon I see a bunch of "Error: synth_odb.tcl, 4 SYNTH_WRAPPED_OPERATORS or SWAP_ARITH_OPERATORS require OPENROAD_HIERARCHICAL to be set." in pr-merge so I think it doesn't work well with your recent PR.

oharboe:
@maliberty good to go, merged w master and fixed errors

maliberty (approved):

PR #3766, opened 2026-01-07
slang: add a SYNTH_BLACKBOXES syntax check

oharboe:
Non-existand verilog modules should be ignored to be compartible with Yosys.

### DIFF
--- a/flow/designs/asap7/uart/config.mk
+++ b/flow/designs/asap7/uart/config.mk
@@ -5,6 +5,9 @@ export DESIGN_NAME            = uart
 
 export VERILOG_FILES          = $(sort $(wildcard $(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/*.v))
 export SDC_FILE               = $(DESIGN_HOME)/$(PLATFORM)/$(DESIGN_NICKNAME)/constraint.sdc
+# Exercise syntax on command line for a dummy module and
+# check that non-existant modules are silently ignored.
+export SYNTH_BLACKBOXES = dummy
 
 export PLACE_DENSITY          = 0.70
 export DIE_AREA               = 0 0 17 17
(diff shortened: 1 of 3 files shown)


oharboe:
@povik I got the following locally:

```
11. Executing SLANG frontend.
read_slang: unknown command line argument '--blackboxed-module'
ERROR: Bad command
```

povik:
@oharboe perhaps your yosys-slang submodule is stale (`--blackboxed-module` was added in recent update https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3759)

oharboe:
@povik @maliberty Merge?

It demonstrates a) that the right version of yosys supporting the required feature is there b) that it is fine to list unknown verilog modules as blackboxes. The latter is a feature that is especially useful when doing parallel abc in bazel.

oharboe:
> @oharboe perhaps your yosys-slang submodule is stale (`--blackboxed-module` was added in recent update #3759)

Ah, I had to pin yosys because newer versions were broken. To fix this I unpinnned yosys and switched to slang.

povik:
@oharboe do you still want to see this getting merged now that we know old tools were the cause?

oharboe:
yes, this is a  smoketest of supported use case

flow/designs/asap7/uart/config.mk
```diff
@@ -5,6 +5,9 @@ export DESIGN_NAME            = uart
 
 export VERILOG_FILES          = $(sort $(wildcard $(DESIGN_HOME)/src/$(DESIGN_NICKNAME)/*.v))
 export SDC_FILE               = $(DESIGN_HOME)/$(PLATFORM)/$(DESIGN_NICKNAME)/constraint.sdc
+# Exercise syntax on command line for a dummy module and
```
povik: I think "Smoke test for the option; shouldn't do anything as the named module does not exist" would be clearer as a comment. I want to make sure people understand it's not a load-bearing setting

oharboe: fixed

povik (approved):

povik:
@maliberty please merge

maliberty (approved):

PR #3769, opened 2026-01-08
Docs: improve installation clarity for new users

NishDs-06:
As a new user, I found the installation options slightly confusing. This PR improves grammar and adds a brief recommendation to help first-time contributors choose an installation method more easily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,13 @@ timeline
 There are multiple ways to install and develop OpenROAD and ORFS. However, the best option depends on your use case, experience level, and personal preference.
 
 
- > **Recommendation for new users:**  
- > If you are new to OpenROAD-flow-scripts, we recommend starting with the Docker-based installation to avoid dependency issues.
+> **Recommendation for new users:**  
+> If you are new to OpenROAD-flow-scripts, Docker can be a reliable way to get started since it avoids most dependency and environment issues.  
+>  
+> On supported platforms, using the pre-built binaries can be an even simpler option as it avoids building from source.  
+>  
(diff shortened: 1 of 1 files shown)


maliberty:
Thanks for contributing.  As a new user do you docker easier, which still requires a build, or using the pre-built binaries (https://openroad-flow-scripts.readthedocs.io/en/latest/user/BuildWithPrebuilt.html#install-openroad)?  Of course you have to be on a supported platform for that.

NishDs-06:
Thanks for pointing that out.
As a new user, I personally felt Docker was easier to get started with since it avoids most dependency and environment issues, even though it still involves building inside the container.
I haven’t tried the pre-built binaries yet, but I see that they can be simpler on supported platforms. My intention with the note was mainly to help first-time users choose a more reliable starting point when platform compatibility isn’t clear.
I’m happy to update the wording to explicitly mention pre-built binaries as an alternative if you think that would be better.

maliberty:
I think it would be good to mention it briefly.

Note that Bazel also avoids having to install dependencies (aside from bazelisk itself; similar to how you would need to install docker).

NishDs-06:
Thanks for the suggestion!
I’ve updated the note to briefly mention pre-built binaries on supported platforms and clarified that the Bazel-based flow also avoids manual dependency installation. Please let me know if this looks good.

maliberty:
Hmmm... I don't see any text about pre-built binaries?

NishDs-06:
Sorry about that — that was my mistake.
I hadn’t committed the updated wording when I replied earlier. I’ve now committed and pushed the change that briefly mentions pre-built binaries and clarifies the Bazel-based flow.

Please let me know if this looks correct now.

maliberty (approved):

PR #3771, opened 2026-01-08
synth: SWAP_ARITH_OPERATORS smoketest

oharboe:


maliberty:
asap7/riscv32i is a better smoke test as uart doesn't show an improvement currently.

oharboe:
@maliberty I will leave you to it.

PR #3778, opened 2026-01-10
fix: improve robustness of flow.sh by quoting paths

NishDs-06:
This PR improves the robustness of flow.sh by properly quoting variables and
paths, preventing failures when directories contain spaces. No functional
behavior is changed.

### DIFF
--- a/flow/scripts/flow.sh
+++ b/flow/scripts/flow.sh
@@ -1,11 +1,21 @@
 #!/usr/bin/env bash
-set -u -eo pipefail
-mkdir -p $RESULTS_DIR $LOG_DIR $REPORTS_DIR $OBJECTS_DIR
-echo Running $2.tcl, stage $1
-(trap 'mv $LOG_DIR/$1.tmp.log $LOG_DIR/$1.log' EXIT; \
- $OPENROAD_EXE $OPENROAD_ARGS -exit $SCRIPTS_DIR/noop.tcl 2>&1 >$LOG_DIR/$1.tmp.log; \
- eval "$TIME_CMD $OPENROAD_CMD -no_splash $SCRIPTS_DIR/$2.tcl -metrics $LOG_DIR/$1.json" 2>&1 | \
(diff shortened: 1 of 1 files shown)


maliberty:
Thanks for contributing.  You need to sign your commits (`git commit -s`) to pass [DCO](https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin).  See the details link by DCO for more info on amending your changes.

NishDs-06:
Thanks for the heads-up! I’ve amended my commit.

maliberty (approved):

flow/scripts/flow.sh
```diff
@@ -1,11 +1,21 @@
...
+    >"$LOG_DIR/$1.tmp.log" 2>&1
+
+  eval "$TIME_CMD $OPENROAD_CMD -no_splash \"$SCRIPTS_DIR/$2.tcl\" -metrics \"$LOG_DIR/$1.json\"" \
+    2>&1 | tee -a "$(realpath "$LOG_DIR/$1.tmp.log")"
+)
+
 # Log the hash for this step. The summary "make elapsed" in "make finish",
 # will not have all the .odb files for the bazel-orfs use-case.
-$PYTHON_EXE $UTILS_DIR/genElapsedTime.py --match $1 -d $LOG_DIR | tee -a $(realpath $LOG_DIR/$1.log)
+"$PYTHON_EXE" "$UTILS_DIR/genElapsedTime.py" --match "$1" -d "$LOG_DIR" \
```
oharboe, line 20: Not handling spaces in file names is endemic in ORFS, indeed in EDA tools. The problem is languages(tcl + make + bash) and, Linux and EDA legacy.

I'm unconvinced that it is a good idea to try to fix this.

There's no test in ORFS that space in filenames work.

maliberty: I agree that there are likely to be more issues but it didn't seem harmful by itself.  I think it better to avoid having spacing in paths as a methodology.

PR #3791, opened 2026-01-14
docs: avoid double negation

oharboe:


maliberty:
I agree with Gemini on this one

oharboe:
> I agree with Gemini on this one

I went over this with Gemini... :-) 

"defined" is misleading, because we can have defined, but empty variables.

Alternatives?

oharboe:
@maliberty Better?

maliberty (approved):

PR #3794, opened 2026-01-15
floorplan: dry run of PDN to check for errors

oharboe:
@gadfort @maliberty Thoughts?

If PDN was blazingly fast, I would remove the variable.

Enabled by default as this is to help novices(such as me).


oharboe:
doesn't work: macros have to be placed before PDN can run.

PR #3795, opened 2026-01-15
variables: CORE_MARGIN docs

oharboe:


flow/scripts/variables.yaml
```diff
@@ -927,6 +927,9 @@ CORE_MARGIN:
     values, one for each margin. The order of the four values are:
     `{bottom top left right}`.
 
+    If this value is too high, then area is wasted, too low and PDN
```
maliberty, line 930: pdn will only fail if you are using a ring afaik (@gadfort )

gadfort: yes

PR #3796, opened 2026-01-15
variables: document MACRO_PLACE_HALO and MACRO_BLOCKAGE_HALO a bit more

oharboe:


flow/scripts/variables.yaml
```diff
@@ -452,12 +452,26 @@ MACRO_PLACE_HALO:
   description: >
     Horizontal/vertical halo around macros (microns). Used by automatic macro
     placement.
+
+    There is no provision for setting a halo per macro, this is a global
+    setting. With an advanced flow, beyond the scope of ORFS, a user
+    could generate a mock .lef with halos that vary per macro,
+    generate a macro placement .tcl and use `MACRO_PLACEMENT_TCL` to
+    place macros. A .lef halo excludes all instances, including standard
```
maliberty, line 460: Halos exist in def not lef.  I'm not sure why we want to document a non-feature (also we may add it and forget to update this).

PR #3797, opened 2026-01-15
synth: improve canonicalization

oharboe:
Improves canonincalization

Option to prune:

1. attribute \src statements with filename and line number.
2. slang fixes: wires with embedded filenames and line numbers

```
attribute \src "bazel-out/k8-opt/bin/foo.sv:149872.1-149904.10"
attribute \top 1
[deleted]
  wire width 55 $ternary$bazel-out/k8-opt/bin/hardware/tile/foo.sv:149903$4199_Y
```


oharboe:
@maliberty @povik I believe this is a dead end, .rtlil is not designed for canonicalization w.r.t. hashes.

povik:
@oharboe I think once you switch to slang (I know issues are pending there) it will solve "wires with embedded filenames and line numbers".

"attribute \src statements with filename and line number." should be solvable with `setattr -unset src *` or similar. I don't think we should strip the source information unconditionally as it will be useful to some users.

oharboe:
Great! Eagerly awaiting to switch to slang then. As you say there's a couple of things to fix for slang...

povik (approved):

maliberty (approved):
Please advise when you switch and we can remove this

oharboe:
@maliberty @vvbandeira Can someone restart the builds, CI outages...

PR #3798, opened 2026-01-15
[FOR DISCUSSION] synth: use .v for canonicalization

oharboe:
.rtlil is optimized for debugging and has lots of paths and line numbers embedded into wires, etc.

.v is yields more stable hashes

@povik Thoughts?


povik:
.v is not meant for checkpointing with Yosys. I would need to think through the consequences, it might break some of the synthesis knobs.

oharboe:
I'll close this once it finishes building, I just want to see if there are failures.

oharboe:
lots of build failures... Switching to slang is next...

PR #3803, opened 2026-01-16
testing: fast default local smoke testing of future slang and hierachical

oharboe:


oharboe:
@maliberty Could you update the rules.json? I don't have that ability locally any longer as DependencyInstaller.sh no longer works.

maliberty:
> @maliberty Could you update the rules.json? I don't have that ability locally any longer as DependencyInstaller.sh no longer works.

In what way does it no longer work?

maliberty:
> @maliberty Could you update the rules.json? I don't have that ability locally any longer as DependencyInstaller.sh no longer works.

I don't believe I have push access.

oharboe:
just cherry pick and use your preferred PR repository?

PR #3804, opened 2026-01-16
Local hacking with bazel

oharboe:


oharboe:
@vvbandeira @maliberty Unrelated CI outage:

```
08:49:53      1225* |       5 |     181 |       34 |      9 |    44 |    +0.0% |   -0.094 |      -53.3 |    990 | swerv.dec.instbuff.genblk5_ib2ff.genblock_dff.dout[8]$_DFF_PN0_/D
08:50:30  jenkins-worker-central1-c4d-4uknrw seems to be removed or offline ; will wait for 5 min 0 sec for it to come back online
08:55:41  jenkins-worker-central1-c4d-4uknrw has been removed or offline for 5 min 0 sec; assuming it is not coming back, and terminating shell step
```

maliberty:
What is the purpose of this change?

oharboe:
> What is the purpose of this change?

I am no longer able to run dependencyinstaller so I build only w bazel.

This allows me to hack ORFS locally more easily.

maliberty:
I often have both and don't want the default changed.  Why can't you run dependencyinstaller ?

oharboe:
> I often have both and don't want the default changed. Why can't you run dependencyinstaller ?

Doesn't work, fails at some point, fickle, other fish to fry....

oharboe:
regarding closing: this PR is supposed to make things easier.

I'm wondering if a better solution would be to create a "bazelisk run :install" that builds and copies openroad into the canonical location, tools/install/OpenROAD/bin.

Thoughts?

maliberty:
sounds ok

maliberty:
I was thinking about this PR and I think what is needed is a way for the user to express a priority preference in that case that both exist.  Perhaps another var?

oharboe:
I prefer same as cmake: canonical location, no options and an install command.

PR #3805, opened 2026-01-16
synth: avoid stdcells in .rtlil

oharboe:
Easier to browse through small .rtlil files now, saves disk space.

Canonicalization improvements are slight as stdcells basically never change.


povik (changes requested):
We can't remove the step from synth_preamble because RTL elaboration requires knowledge of the libraries to interpret macros and standard cell instantiations.

In theory we could delete the modules coming from .lib just before writing out `1_1_yosys_canonicalize.rtlil` with `delete A:liberty_cell`

PR #3806, opened 2026-01-16
synth: clean out unneeded liberty cells from .rtlil

oharboe:
Tested locally:

Before:

```
$ grep liberty results/nangate45/gcd/base/1_1_yosys_canonicalize.rtlil  | wc -l
815
```

After:

```
$ make do-yosys-canonicalize 
$ grep liberty results/nangate45/gcd/base/1_1_yosys_canonicalize.rtlil  | wc -l
0
$
```


oharboe:
@povik Need some help here, I managed to delete the liberty cells, but that caused other problems.

PR #3825, opened 2026-01-21
synth: add SYNTH_ODB_TCL

oharboe:
My thinking is that I would run:

```
remove_buffers
# maybe with some max fanout?
repair_design
```

This gets rid of buffers added for in-buffer-out and replace that with in-out in modules that would have no buffer if synthesis was flattened, yet we get timing after synthesis that is accurate enough that can look for problems in the design.

Thoughts?

Hmmm... we tried swap operators and saw enormous clock period increases immediately after synthesis. Would this fix that too maybe?

If swapping operators was done on some net with massive fanout...


maliberty:
Are you describing creating a pure feedthrough?   If so that is a bad idea as you will not get accurate timing.

oharboe:
> Are you describing creating a pure feedthrough? If so that is a bad idea as you will not get accurate timing.

This is a feedthrough that is going through a module that wouldn't exist if I flattened, so removing buffers with fanout of 1 that are not connected to top level pins that are(I assume) an artifact of hierarchical synthesis.

maliberty:
I'm not clear why hierarchical synthesis would add a buffer for that case.

oharboe:
> I'm not clear why hierarchical synthesis would add a buffer for that case.

Well, it is "hierarchical synthesis at home" or rather parallel hierarchical synthesis. I have a list of modules that I synthesize in parallel and I list all modules except the current module as BLACKBOX, then stitch together all the netlists before feeding them into the ORFS flow.

I assume Yosys adds a buffer on pins at the top level, except when I stitch them together as above, a module isn't necessarily top level anymore.

So maybe what is needed is run `opt_clean -purge` after loading in netlists with `SYNTH_NETLIST_FILES`, except in synth_preamble.tcl, the files are simply copied.

oharboe:
Maybe what I am looking for is a way not to run the two lines below for non-top level synthesis?

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/fcbb70dad1ddc444cc20fc6d57d773fe756333c8/flow/scripts/synth.tcl#L221-L222

I doesn't appear to be a way to do this, except for top level in2out pins.

oharboe:
Closing PR for now. Need to understand the problem better to understand what is needed here.

maliberty:
OR buffers ports, I'm not aware of yosys doing that.

PR #3840, opened 2026-01-26
Show all physical instances other than filler cells

gudeh:
Set to `visible false` only filler cells, this way we can see tap cells and end caps at the end of placement on the resulting image.


gudeh:
Example with asap7/aes-block

Current image:
image

New resulting image:
image

maliberty:
The images look the same??

gudeh:
> The images look the same??

Look closely, they are different! The second one has end or tap cells. For example the top right corner

maliberty (approved):

PR #3845, opened 2026-01-28
gui: add GUI_WNS that prunes to N WNS paths for large designs

oharboe:
If the concept is sound, I'll start testing to see if it helps scaling for large projects.

    make gui_synth

image

vs.

    make GUI_WNS=1 gui_synth

image


oharboe:
@maliberty Thoughts?

maliberty:
You have to time the entire design to get WNS so I'm unclear how this speeds things up.

oharboe:
> You have to time the entire design to get WNS so I'm unclear how this speeds things up.

The list of endpoints to keep can be written to a file during a build and available as an artifact.

maliberty:
Does this PR do that or is that future work?

oharboe:
> Does this PR do that or is that future work?

Future work, I just wanted your thoughts if that is a meaningful path?

maliberty:
You will have to at least up the fanin cone of all the endpoints.  Depending on the design / num pts that could still be a large fraction of the design.  My guess is this won't be a game changer and would need to see some evidence to go further.

oharboe:
maybe pursue something like this later

Issue #3645, opened 2025-11-10
Improve user experience and documentation for large designs failing to pass synthesis


oharboe:
### Description

When trying out large designs, Yosys can easily get stuck on some futile optimizations due to a missing inferred memory or pathological sizes in some module.

Example below:

Untar [slang-stuck.tar.gz](https://github.com/user-attachments/files/23470727/slang-stuck.tar.gz)

Run the below:

    make --file=~/OpenROAD-flow-scripts/flow/Makefile SYNTH_HDL_FRONTEND=slang SYNTH_MEMORY_MAX_BITS=1024 DESIGN_CONFIG=~/OpenROAD-flow-scripts/flow/designs/asap7/minimal/config.mk DESIGN_NAME=CoralNPUChiselSubsystem  SDC_CONSTRAINTS=~/OpenROAD-flow-scripts/flow/designs/asap7/mock-alu/constraints.sdc VERILOG_FILES=./home/oyvind/coralnpu/bazel-bin/fpga/ip/coralnpu_chisel_subsystem_default/CoralNPUChiselSubsystem.sv SDC_CONSTRAINTS=~/OpenROAD-flow-scripts/flow/designs/asap7/mock-alu/constraints.sdc do-yosys-canonicalize do-yosys
    

Fails quickly, as expected, memories are inferred:

```
6.15.10. Executing MEMORY_COLLECT pass (generating $mem cells).
6.16. Executing OPT_CLEAN pass (remove unused cells and wires).
ERROR: TCL interpreter returned an error: Source files actually used in the design:
 home/oyvind/coralnpu/bazel-bin/fpga/ip/coralnpu_chisel_subsystem_default/CoralNPUChiselSubsystem.sv
Memories found in the design:
 Rows | Width |   Bits | Module               | Instances                                                                       
---------------------------------------------------------------------------------------------------------------------------------
[...]

PR #3855, opened 2026-02-02
mem_dump: add summary of inferred memory usage

NishDs-06:
This PR improves the usability of mem_dump.py by adding a small summary section
after the inferred memory table.

The summary highlights:
- total number of inferred memories
- largest single inferred memory instance
- total inferred memory bits across all instances
- status relative to SYNTH_MEMORY_MAX_BITS (when provided)

This makes it easier for users to understand why synthesis may be slow or fail
early due to large inferred memories, and complements the existing fail-fast
and memory mocking mechanisms discussed in #3645.

### DIFF
--- a/flow/scripts/mem_dump.py
+++ b/flow/scripts/mem_dump.py
@@ -4,8 +4,6 @@ import sys
 
 
 def find_top_modules(data):
-    # There can be some cruft in the modules list so that
-    # we have multiple top level candidates.
     top_module = []
     instantiations = set(
         [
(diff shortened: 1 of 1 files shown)


oharboe (commented):
Probably better that @povik comments on this. I just hacked together something that read in the .json file and that has done the job for a year or two...

These days, I think the best way to do this is to query the AST in Yosys directly?

That's what @povik did when updating synth.tcl to handle mocking of large memories which ties into the motivation for this report/error: avoid synthesis getting stuck silently forever on unintentional massive flip flop memories.

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/36661735d8061bc5bdf7a2cb57fbfa2a9bc0a671/flow/scripts/synth.tcl#L92-L117

maliberty (approved):

maliberty:
I think suggesting a different approach entirely is beyond the scope of this PR

oharboe:
Thanks! I tried to clean up this code, but failed.

PR #3857, opened 2026-02-02
gui: add gui/open_yosys targets

oharboe:
Useful when viewing a netlist with blackboxed modules where there's no .odb file in parallel synthesi

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -389,6 +389,8 @@ endef
 # Custom target to go from synthesis to placement in a single OpenROAD run
 $(eval $(call do-step,1_3_floorplan_to_place, $(RESULTS_DIR)/1_synth.odb $(RESULTS_DIR)/1_synth.sdc,floorplan_to_place))
 
+$(eval $(call OPEN_GUI_SHORTCUT,yosys,1_2_yosys.v))
+
 .PHONY: floorplan_to_place
 floorplan_to_place: $(RESULTS_DIR)/1_synth.odb $(RESULTS_DIR)/1_synth.sdc
 	$(UNSET_AND_MAKE) do-1_3_floorplan_to_place
@@ -777,6 +779,7 @@ $(eval $(call OPEN_GUI_SHORTCUT,final,6_final.odb))
 
 $(foreach file,$(RESULTS_DEF),$(eval $(call OPEN_GUI,$(file),DEF_FILE)))
 $(foreach file,$(RESULTS_ODB),$(eval $(call OPEN_GUI,$(file),ODB_FILE)))
+$(foreach file,$(RESULTS_V),$(eval $(call OPEN_GUI,$(file),V_FILE)))
 
 # Write a def for the corresponding odb
 $(foreach file,$(RESULTS_ODB),$(file).def): %.def:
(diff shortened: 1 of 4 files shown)


oharboe:
@maliberty Recently write_db will no longer write out .odb files with blackboxed submodules, I believe an intentional sanity check, so this covers the view timing for netlist use-case.

oharboe:
@maliberty unrelated mock-cpu (mesochronous-fifo) QoR failures:

    14:33:49  [ERROR] finish__timing__setup__tns fail test: -2362.01 >= -2230

maliberty:
Please explain the write_db issue.  You probably need to merge master for the metrics issue.

oharboe:
> Please explain the write_db issue. You probably need to merge master for the metrics issue.

If I use SYNTH_BLACKBOXES for a module, write_db no  longer will save a .odb file, but fails.

I thought this was an intentional sanity check?

oharboe:
did a merge, pr head passed, pr merge failed.

@maliberty so master is a bit flaky now?

maliberty:
In what way does write_db  fail?  I'm not sure what sanity check you are describing.

oharboe:
It says there is no master for the blackboxed module, as I recall.

oharboe:
Example:

```
$ make DESIGN_CONFIG=designs/asap7/aes/config.mk SYNTH_BLACKBOXES=aes_rcon
[deleted]
[WARNING STA-0198] ./results/asap7/aes/base/1_2_yosys.v line 102385, module aes_rcon not found. Creating black box for \u0.r0 .
[ERROR ORD-2013] instance u0.r0 LEF master aes_rcon not found.
```

oharboe:
@maliberty Merge? I think this is a nice addition, even if https://github.com/The-OpenROAD-Project/OpenROAD/issues/9413 is/should be fixed.

maliberty:
That error isn't related to write_db.  It is a sanity check as you can't do physical design without a LEF to give a physical description.  Do you not have a LEF or have you not loaded it?

oharboe:
> That error isn't related to write_db. It is a sanity check as you can't do physical design without a LEF to give a physical description. Do you not have a LEF or have you not loaded it?

Neither: I'm doing hierarchical synthesis.

# Hierarchical Synthesis with Bazel-ORFS

## The Concept

Hierarchical synthesis breaks a large design into parallelizable chunks to save time and memory. Instead of a single "flat" synthesis run, you treat sub-modules as independent entities.

### 1. Parallel Synthesis

Define a list of modules to be synthesized simultaneously. Bazel handles the scheduling, ensuring each module is mapped to gates in its own isolated process.

### 2. Blackboxing and synthesizing each module separately as a top module

Each module in that list is synthesized as a top module.

When synthesizing each **Top** module, use `SYNTH_BLACKBOXES` to list all other modules except the top module. blackboxing modules that don't exist in the design gives a warning in yosys but is harmless.

* **Effect:** The synthesizer treats sub-modules as empty shells with defined ports.
* **Benefit:** The Top-level run finishes in seconds because it only synthesizes the "glue logic" connecting the blocks.

### 3. Netlist Assembly

Since the outputs are structural Verilog, the final step is a simple file concatenation.

```bash
cat sub_module_A.v sub_module_B.v top_level.v > final_netlist.v
```

oharboe:
@maliberty Where to from here?

maliberty (approved):

PR #3873, opened 2026-02-07
Add missing ASAP7 OA Verilog cell models

taktoa:
For some reason the asap7sc7p5t_OA_RVT_TT_201020.v file was the only one missing from the set of Verilog standard cell models in `flow/platforms/asap7/verilog/stdcell`. ORFS does generate netlists that contain these cells, so I'm not sure how simulation could have been working previously.

### DIFF
--- a/flow/platforms/asap7/verilog/stdcell/asap7sc7p5t_OA_RVT_TT_201020.v
+++ b/flow/platforms/asap7/verilog/stdcell/asap7sc7p5t_OA_RVT_TT_201020.v
@@ -1,0 +1,5785 @@
+// BSD 3-Clause License
+// 
+// Copyright 2020 Lawrence T. Clark, Vinay Vashishtha, or Arizona State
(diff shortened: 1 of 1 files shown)


taktoa:
I see now that there is no simulation flow in ORFS currently (for some reason I thought there was a power simulation flow that made use of these models). I would recommend either completing the set here, or deleting them entirely since they seem to be unused.

oharboe:
ORFS asap7 files are used via docker image in bazel flows here https://github.com/The-OpenROAD-Project/OpenROAD/tree/master/test/orfs

maliberty (approved):

maliberty:
The commits needs to be signed to pass DCO.  Click on DCO for more info on how to fix that.  In general `git commit -s` signs commits for the future.

taktoa:
Fixed, sorry about that.

Issue #3867, opened 2026-02-05
ihp-sg13g2: Missing connection between PWR/GND ring and IO cells


dnltz:
### Subject

[Stage]: Floorplaning.

### Describe the bug

I'm building the i2c-gpio-expander in this repository. See https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/designs/ihp-sg13g2/i2c-gpio-expander/pdn.tcl

After updating OpenROAD from (ORFS commit) 252042c9c294386a7a8594dc3465b542c0184c52 to cc3eba7744b76ebe4c864d20a735e07bd7d357c6, the power/ground ring around the core area is not connected to the IO cells anymore.

Before this update, the ITerms in the IO cell were set to `VSS` and `VDD` and matched the core voltage grid. Now, they're called `IO_CORNER_NORTH_WEST_INST.vdd_RING` and `...vss_RING`.

<img width="566" height="406" alt="Image" src="https://github.com/user-attachments/assets/f0b9f951-9aef-49c2-9acb-77416b865bb8" />

Screenshot of the missing connection:

<img width="917" height="861" alt="Image" src="https://github.com/user-attachments/assets/101014dd-0fc6-4558-b831-dbec5a59a527" />

Is there any command to merge those two nets into one big?

How can I overwrite nets in the IO cells?

### Expected Behavior

.

### Environment

```shell
I don't have these information anymore, because I build OpenROAD in a multi-stage container. This is the ORFS commit is used to built OpenROAD: cc3eba7744b76ebe4c864d20a735e07bd7d357c6

OS: Ubuntu 24.04
```

### To Reproduce

[floorplan_i2c-gpio-expander_ihp-sg13g2_base_2026-02-05_18-44.tar.gz](https://github.com/user-attachments/files/25107945/floorplan_i2c-gpio-expander_ihp-sg13g2_base_2026-02-05_18-44.tar.gz)

### Relevant log output

```shell

```

dnltz:
There are also these warnings in the logs, but I have seen them for 2 years now.
```
==========================================================================
Floorplan check_setup
--------------------------------------------------------------------------
...
[WARNING PAD-0033] Could not find a block terminal associated with net: "IO_CORNER_NORTH_WEST_INST.vdd_RING", creating now.
[...]

PR #3874, opened 2026-02-07
designs: ihp-sg13g2: i2c-gpio-expander: Fix padring power/ground nets

dnltz:
Place `add_global_connection` for power and ground nets in the padring before the padring got created.

Also remove `add_global_connection` from the ihp-sg13g2 default `pdn.tcl` file, because it should be defined in the `FOOTPRINT_TCL` file.

See #3867 for reference.

### DIFF
--- a/flow/designs/ihp-sg13g2/i2c-gpio-expander/pad.tcl
+++ b/flow/designs/ihp-sg13g2/i2c-gpio-expander/pad.tcl
@@ -28,6 +28,14 @@ proc calc_vertical_pad_location { index total IO_LENGTH IO_WIDTH BONDPAD_SIZE SE
   }]
 }
 
+# padframe core power pins
+add_global_connection -net {VDD} -pin_pattern {^vdd$} -power
+add_global_connection -net {VSS} -pin_pattern {^vss$} -ground
+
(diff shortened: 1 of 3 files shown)


maliberty:
Why move the pad related connect out of the platform?  It seems like it would be relevant to any design with pads, no?

dnltz:
> Why move the pad related connect out of the platform? It seems like it would be relevant to any design with pads, no?

The `pdn.tcl` file is the wrong location and too late. Those nets should be created before the padring got placed.

I would like to make a proposal soon to move some of common code from https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/designs/ihp-sg13g2/i2c-gpio-expander/pad.tcl into a platform-specific file. For example, the `calc_horizontal_pad_Location` function can be re-used. This common file should also properly define the padring nets.

dnltz:
I wrote down my idea here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3875

dnltz:
@maliberty can we continue with this PR?

maliberty (approved):

Issue #3826, opened 2026-01-22
[ERROR] asap7-ibex write_verilog produces syntax error verilog with OPENROAD_HIERARCHICAL enabled


liW-J:
### Subject

[Stage]: Other. Please describe below.

### Describe the bug

when running the default **asap7-ibex** flow with ORFS, the generated `6_final.v` cannot be correctly parsed by OpenROAD due to invalid Verilog syntax. This issue seems to be related to the recently introduced default enablement of `OPENROAD_HIERARCHICAL` for some designs.

```
openroad> read_verilog ./results/asap7/ibex/base/6_final.v
[ERROR STA-0164] ./results/asap7/ibex/base/6_final.v line 72893, syntax error
STA-0164
while evaluating read_verilog ./results/asap7/ibex/base/6_final.v
```

### Expected Behavior

The verilog generated by `write_verilog` should be syntactically valid and readable by OpenROAD.

I tested **nangate45-aes**, with OPENROAD_HIERARCHICAL also default enabled. Its generated 6_final.v can be successfully parsed using `read_verilog`:
```
openroad> read_verilog ./results/nangate45/aes/base/6_final.v
openroad> 
```

For **asap7-ibex**, we tried using yosys to flatten the verilog output during the **synthesis stage**, and the  final verilog written by OpenROAD can be read correctly.

### Environment

```shell
ORFS git commit: e19b7a5176e45bae645ba2d5b46f4457e8913b1c
OpenROAD git commit: 36764173fae6bb1eb4b35f0c2138f42b3cdfadd8
kernel: Linux 3.10.0-1160.el7.x86_64
os: Ubuntu 24.04.3 LTS (Noble Numbat)
cmake version 3.28.3
-- The CXX compiler identification is GNU 13.3.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-27957-g36764173fa
-- System name: Linux
-- Compiler: GNU 13.3.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 20
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- LTO/IPO is disabled
-- The C compiler identification is GNU 13.3.0
-- Detecting C compiler ABI info
[...]

PR #3880, opened 2026-02-10
enable lec check

nanocoh:


maliberty:
I see some errors that look like tool issues on your side, eg ng45/swerv

[2026-02-10T22:12:14.527Z] [2026-02-10 22:12:13.459] [info] [KeplerFormal.cpp:263] Found top design: swerv
[2026-02-10T22:12:14.527Z] [2026-02-10 22:12:13.681] [error] [KeplerFormal.cpp:339] Workflow failed: Anonymous instance in SNLPath constructor.
[2026-02-10T22:12:14.527Z] child process exited abnormally

nanocoh:
> I see some errors that look like tool issues on your side, eg ng45/swerv
> 
> [2026-02-10T22:12:14.527Z] [2026-02-10 22:12:13.459] [info] [KeplerFormal.cpp:263] Found top design: swerv [2026-02-10T22:12:14.527Z] [2026-02-10 22:12:13.681] [error] [KeplerFormal.cpp:339] Workflow failed: Anonymous instance in SNLPath constructor. [2026-02-10T22:12:14.527Z] child process exited abnormally

A lot of issues are regarding instances that need to be cleaned before dumping the verilog through "write_verilog -remove_cells", will need to edit it one by one but there might be also tool issues indeed among the tests. 

Working on cleaning it all one by one and will update the pr right after.

nanocoh:
We filtered the majority of config issues.

Some issues remain on the tool that we are solving.

There is also one issue in the verilog parsing level for which @xtofalex have some comments:

Parser error: syntax error, unexpected ESCAPED_IDENTIFIER_TK, expecting '(	
That can be found on ibex asap7 for example.

maliberty:
There is a pending fix for a Verilog escaping issue.  Perhaps that is what you are seeing.  What does that line look like?

nanocoh:
> There is a pending fix for a Verilog escaping issue. Perhaps that is what you are seeing. What does that line look like?

It very might well be the case here. Will let @xtofalex comment on what he saw.

nanocoh:
In case it is a pending fix, I can disable it for now in the related test cases.

xtofalex:
> There is a pending fix for a Verilog escaping issue. Perhaps that is what you are seeing. What does that line look like?

We saw an escaping issue coming from the OpenSTA Verilog dumper when reading results/ihp-sg13g2/ibex/base/4_before_rsz_lec.v

```console
%Error: 4_before_rsz_lec.v:70061:7: syntax error, unexpected IDENTIFIER-for-type, expecting ')'
70061 |     .Y\[32:1\]({net462,
      |       ^~~~~~~~~~~~~~~~~
        ... See the manual at https://verilator.org/verilator_doc.html?v=5.044 for more assistance.
%Error: Exiting due to 1 error(s)
zsh: exit 1     verilator --lint-only -Wall 4_before_rsz_lec.v
```

The line should be: ".\Y[32:1] ({net462,".

From quick analysis, the problem comes from:
https://github.com/The-OpenROAD-Project/OpenSTA/blame/e872c55bfe1c005cb0dae8cd281f986dd39cfd34/verilog/VerilogWriter.cc#L391

where in "void VerilogWriter::writeInstBusPin"  :
```c++
fprintf(stream_, ".%s({", network_->name(port));
```
should be replace with something like in "void VerilogWriter::writeInstPin" :
```c++
 string net_vname = netVerilogName(net_name);
```

@maliberty, If this is not related to the pending fix you are referring to, I can open an issue in OpenSTA.

povik:
@xtofalex looks to be related to the pending fix. See https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3826

xtofalex:
> @xtofalex looks to be related to the pending fix. See #3826

Yes thanks !

flow/scripts/cts.tcl
```diff
@@ -70,7 +70,7 @@ if { !$::env(SKIP_CTS_REPAIR_TIMING) } {
   if { $::env(EQUIVALENCE_CHECK) } {
     run_equivalence_test
   }
-  if { [env_var_exists_and_non_empty LEC_CHECK] } {
+  if { ![env_var_exists_and_non_empty DISABLE_LEC_CHECK] } {
```
maliberty: The variable should be defined and documented in variables.yaml (e.g. EQUIVALENCE_CHECK).  It is better to have positive names to avoid having double negatives like DISABLE_LEC_CHECK=false.  Let's use LEC_CHECK with the default=1.

flow/scripts/lec_check.tcl
```diff
@@ -1,6 +1,6 @@
 proc write_lec_verilog { filename } {
-  if { [env_var_exists_and_non_empty REMOVE_CELLS_FOR_EQY] } {
-    write_verilog -remove_cells $::env(REMOVE_CELLS_FOR_EQY) $::env(RESULTS_DIR)/$filename
+  if { [env_var_exists_and_non_empty REMOVE_CELLS_FOR_LEC] } {
```
maliberty, line 2: REMOVE_CELLS_FOR_LEC in variables.yaml

flow/designs/asap7/aes-block/config.mk
```diff
@@ -30,3 +30,4 @@ export PWR_NETS_VOLTAGES      =
 export MACRO_PLACE_HALO        ?= 3 3
 
 export ROUTING_LAYER_ADJUSTMENT = 0.3
+export REMOVE_CELLS_FOR_LEC     = TAPCELL*
```
maliberty: It would be better to put this in the platform rather than every design's config.mk.  Use ?= so it can be overridden if needed.

maliberty (changes requested):
I'm curious about the ones with DISABLE_LEC_CHECK=1 - do you plan to look at those later?

nanocoh:
> I'm curious about the ones with DISABLE_LEC_CHECK=1 - do you plan to look at those later?

We disabled all designs with the pending fix for Verilog escaping. We will of-course turn them on and verify them in another pr when the fix is integrated.

flow/designs/asap7/aes-block/config.mk
nanocoh: Done

flow/scripts/lec_check.tcl
nanocoh: Done

flow/scripts/cts.tcl
nanocoh: Done

maliberty (approved):

PR #3886, opened 2026-02-12
util/docker_shell: Don't hard-code DISPLAY for fetching auth info

frej:
The flow/util/docker_shell script looks up the X authorization information (magic cookie) for display :0 and exports it to the container, but sets DISPLAY to match DISPLAY of the host shell. There are no guarantees that the authorization information for :0 is the same as for $DISPLAY, if $DISPLAY != :0.

This patch changes flow/util/docker_shell to always extract the authorization information for $DISPLAY.


maliberty (approved):
Thanks for contributing!

PR #3888, opened 2026-02-13
Fix asap7 mbff lib files

LeeJaKang:
This PR replaces the ASAP7 [*_FAKE.lib] MBFF libraries with regenerated versions.

1. Why the previous lib values were incorrect
Our analysis of the existing FAKE MBFF libs found that some values deviated from the intended baseline.
The key issue was evidence that values were effectively copied from single-bit SDFH cells, instead of using the intended baseline single-bit flip-flop (i.e., DFFHQN).
As a result, leakage/power values were inconsistent with the intended corner/VT context, reducing MBFF data reliability.

2. Regeneration method
Instead of preserving legacy values, we regenerated the FAKE libs using the Python-script-based flow from the MBFFClustering repository.
Reference: https://github.com/ABKGroup/MBFFClustering

During regeneration, MBFF values were rebuilt using the corresponding SBFF with matched corner/VT, and existing FAKE files were replaced with the regenerated outputs.

3. Corner expansion (FF/SS added)
Previously, the FAKE MBFF set was effectively TT-focused.
This update adds FF/SS coverage as well.
Now the following combinations are covered:

VT: LVT / RVT / SLVT
Corner: TT / FF / SS

4. DFHV removal
DFHV-family FAKE libs were removed because they were considered unnecessary for this cleaned MBFF set.
Therefore, this PR keeps the FAKE MBFF set focused on the required MBFF cells only.


maliberty:
I have concerns about the correctness still.  For example the one-bit DFFHQNx1_ASAP7_75t_R has a leakage of 1.50698e-10 (average of all conditional leakage() statements assuming equal likelihood).   

DFFHQNV2Xx1_ASAP7_75t_R is a two-bit MBFF with area that is exactly twice the above.  The leakage however is 4.135266e-10 which is more than double the single-bit leakage above.  I expect it to be at most equal and probably less to find some saving by clustering.

maliberty:
My bad I mixed up TT and FF corners.

maliberty (approved):

PR #3904, opened 2026-02-22
variables: add ASAP7_USE_VT

oharboe:
Use-case, check effect on timing by using different voltage thresholds, in which case bazel-orfs wants to know that ASAP_USE_VT is an ORFS variable:

    bazelisk run --define "ASAP7_USE_VT=LVT SLVT" Foo_synth $(pwd)/synth gui_synth

### DIFF
--- a/flow/scripts/variables.yaml
+++ b/flow/scripts/variables.yaml
@@ -1306,3 +1306,10 @@ REMOVE_CELLS_FOR_LEC:
     String patterns directly passed to write_verilog -remove_cells <> for
     lec checks.
   type: string
+ASAP7_USE_VT:
+  description: >
+    A space separated list of VT options to use with the ASAP7 standard cell library:
+    RVT, LVT, SLVT.
+  stages:
+    - All stages
+  default: RVT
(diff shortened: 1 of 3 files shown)


flow/scripts/variables.yaml
```diff
@@ -1305,4 +1305,11 @@ REMOVE_CELLS_FOR_LEC:
   description: >
     String patterns directly passed to write_verilog -remove_cells <> for
     lec checks.
-  type: string
\ No newline at end of file
+  type: string
+ASAP7_USE_VT:
```
maliberty, line 1309: This variable isn't useful outside of asap7 so I'm not sure why to promote it to all PDKs.

oharboe: It is useful for bazel-orfs to know that it is an ORFS variable.

oharboe: And we get some docs generated....

oharboe:
Documented use-case in PR synopsis.

maliberty (approved):
Don't love it but not worth fighting.  Ideally we would have a more general Vt configuration strategy.

oharboe:
> Don't love it but not worth fighting. Ideally we would have a more general Vt configuration strategy.

@jeffng-or Thank you for adding this support! This was very helpful to put some questions we had to bed.

PR #3908, opened 2026-02-23
LEC_CHECK: default 0 for out of tree designs

### LINKED ISSUE #3907: Make LEC_CHECK 0 default for non-ORFS test-suite
### Description

designs using ORFS, but not part of ORFS test-suite, probably don't want to run LEC_CHECK=1.

The trick is that LEC_CHECK=1 is a reasonable default for the ORFS test scripts themselves.

### Suggested Solution

Make LEC_CHECK=0 by default, pass in `make LEC_CHECK=1` in CI.

Some designs explicitly disable LEC_CHECK in config.mk and they override `make LEC_CHECK=1`.

### Additional Context

_No response_

oharboe:
```shell
$ cd OpenROAD-flow-scripts/flow
$ make print-ORFS_TEST_SUITE_DESIGN print-LEC_CHECK
ORFS_TEST_SUITE_DESIGN: 1
LEC_CHECK: 1
$ make DESIGN_CONFIG=../config.mk print-ORFS_TEST_SUITE_DESIGN print-LEC_CHECK
ORFS_TEST_SUITE_DESIGN: 0
LEC_CHECK: 0
```

fixes #3907

### DIFF
--- a/flow/scripts/variables.yaml
+++ b/flow/scripts/variables.yaml
@@ -1297,8 +1297,10 @@ WRITE_ODB_AND_SDC_EACH_STAGE:
   default: 1
 LEC_CHECK:
   description: >
-    Perform a formal equivalence check between before and after netlists.
-  default: 1
+    This is primarily an ORFS CI regression testing feature and is default 0
+    for DESIGN_CONFIG pointing designs outside of the ORFS tree.
+
+    Requires `KEPLER_FORMAL_EXE` to point to a built Kepler executable.
   stages:
     - cts
 REMOVE_CELLS_FOR_LEC:
--- a/docs/user/FlowVariables.md
+++ b/docs/user/FlowVariables.md
@@ -161,7 +161,7 @@ configuration file.
 | <a name="KLAYOUT_TECH_FILE"></a>KLAYOUT_TECH_FILE| A mapping from LEF/DEF to GDS using the KLayout tool.| |
 | <a name="LATCH_MAP_FILE"></a>LATCH_MAP_FILE| Optional mapping file supplied to Yosys to map latches| |
 | <a name="LAYER_PARASITICS_FILE"></a>LAYER_PARASITICS_FILE| Path to per layer parasitics file. Defaults to $(PLATFORM_DIR)/setRC.tcl.| |
-| <a name="LEC_CHECK"></a>LEC_CHECK| Perform a formal equivalence check between before and after netlists.| 1|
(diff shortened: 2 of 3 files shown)


flow/scripts/variables.yaml
```diff
@@ -1297,8 +1297,10 @@ WRITE_ODB_AND_SDC_EACH_STAGE:
   default: 1
 LEC_CHECK:
   description: >
-    Perform a formal equivalence check between before and after netlists.
-  default: 1
+    This is primarily an ORFS CI regression testing feature and is default 0
```
maliberty, line 1300: I don't agree with this description.  Anyone doing production work will likely want LEC checking.  I can see not wanting to spend the time while prototyping.  

The in-tree / out-of-tree distinction is a bit odd.  I think it works for your particular use case but isn't that general.   Perhaps it would be better to have an option project setting file you could put it there.

oharboe: I think there are two use-cases, or concerns, that should be clearly separated: development of ORFS/OpenROAD and usage of OpenROAD/ORFS.

For the usage use-case(where I'm not an OpenROAD developer, but a user), I think LEC should be default off for out of tree designs: minimum dependencies, fast lean, optional(for architectural exploration I don't care about correctness of throwaway results). I trust OpenROAD and only need LEC as a last crossing of t's and dotting's of i's. I don't want to bog down day to day CI wasted compute or extra dependencies.

For the OpenROAD/ORFS development use-case, I have no strong opinions, I think OpenROAD/ORFS developers are the best to judge and I think that they should be free to do so without worrying about the user-experience.

For the user's use-case, how about a more explicit invocation like `make lec-floorplan/place/cts/grt/drt/final` that checks previous stage against current stage?

`make lec-final` should probably do a LEC against drt as well as synthesis netlist following POLA.

I think the user use-case can be kicked down the road until we have more milage on LEC and we know that it works and is scalable for production, so this PR is good to go from that perspective.

oharboe:
@maliberty @vvbandeira pr-merge worked yesterday, failed today.

Some unrelated CI error on pr-merge? The error doesn't give me much to go on.

image

oharboe:
@maliberty What should a user do if LEC check fails?

My understanding is:

1. report a github issue to be a good citizen
2. go about his business and continue to work on the project
3. ignore it unless the LEC failure is on `make lec-final` for the signoff

No?

PR #3911, opened 2026-02-23
Support Kali GNU/Linux and generic Debian derivatives in DependencyInstaller

### LINKED ISSUE #3910: Add support for Kali GNU/Linux (and generic Debian derivatives)
## Summary

Building OpenROAD-flow-scripts on Kali GNU/Linux fails at multiple stages due to OS detection not recognizing Kali (or generic Debian derivatives). Since Kali is Debian-based, it should be routed through the existing Debian/Ubuntu installation paths.

## Issues Encountered

1. **OS detection failure** — Both `etc/DependencyInstaller.sh` and `tools/OpenROAD/etc/DependencyInstaller.sh` only match `"Ubuntu"`, `"Debian GNU/Linux rodete"`, and `"Debian GNU/Linux"` exactly. Kali reports its OS as `"Kali GNU/Linux"`, causing an "unsupported system" error.

2. **`libtcl` virtual package** — On Kali (and some Debian versions), `libtcl` is a virtual package with no install candidate. The script needs to fall back to `libtcl8.6`.

3. **Hardcoded `libpython3.8`** — The Debian install path defaults to `libpython3.8`, which doesn't exist on newer Debian-based distros (Kali has Python 3.13+). The version should be auto-detected.

4. **or-tools download URL** — Kali's `VERSION_ID` is `2025.4`, which doesn't correspond to any or-tools release. Non-standard Debian version IDs need to be normalized to a compatible Debian release (e.g., `12`).

5. **Abseil compiled without `-fPIC`** — On GCC 15 (Kali/Debian Testing), PIE is the default. Abseil static libraries need `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` to link into PIE executables.

## Environment

- **OS**: Kali GNU/Linux 2025.4 (Rolling, based on Debian Testing)
- **GCC**: 15.2.0
- **glibc**: 2.42
- **Python**: 3.13

## Expected Behavior

`setup.sh` and `build_openroad.sh` should complete successfully on Kali and other Debian derivatives.

Subhankar-hub:
## Summary

- Add `"Kali GNU/Linux"` and wildcard `"Debian GNU/Linux"*` to the OS case match in `etc/DependencyInstaller.sh` so that Debian-based distributions (including Kali) are routed through the existing Ubuntu/Debian package installation path

## Context

This is the companion PR to The-OpenROAD-Project/OpenROAD#9522, which contains the bulk of the changes in the OpenROAD submodule's `DependencyInstaller.sh`.

## Test plan

- [x] Tested `sudo ./setup.sh` on Kali GNU/Linux 2025.4
- [x] Verified full OpenROAD build succeeds
- [x] Verified `openroad -version` runs correctly
- [ ] Verify no regressions on Ubuntu 22.04/24.04
- [ ] Verify no regressions on Debian 12

Fixes #3910

Made with [Cursor](https://cursor.com)

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -459,7 +459,7 @@ case "${os}" in
             _installPipCommon
         fi
         ;;
-    "Ubuntu" | "Debian GNU/Linux rodete" )
+    "Ubuntu" | "Debian GNU/Linux"* | "Kali GNU/Linux" )
         version=$(awk -F= '/^VERSION_ID/{print $2}' /etc/os-release | sed 's/"//g')
         if [[ -z ${version} ]]; then
             version=$(awk -F= '/^VERSION_CODENAME/{print $2}' /etc/os-release | sed 's/"//g')


vvbandeira:
@Subhankar-hub you need to fix the DCO error.

PR #3919, opened 2026-02-24
flows: platforms: ihp-sg13g2: Suppress DRT-0349

dnltz:
OpenROAD is printing some warnings about missing CUTCLASS with LEF58_ENCLOSURE. LEF58_ENCLOSURE is not defined in the IHP Open PDK nor is this feature required for 130nm.

Suppress this warning to remove them from the logs.

[WARNING DRT-0349] LEF58_ENCLOSURE with no CUTCLASS is not supported. Skipping for layer Cont [WARNING DRT-0349] LEF58_ENCLOSURE with no CUTCLASS is not supported. Skipping for layer Via1 ...

### DIFF
--- a/flow/platforms/ihp-sg13g2/config.mk
+++ b/flow/platforms/ihp-sg13g2/config.mk
@@ -31,6 +31,8 @@ export LIB_FILES ?= $(TYP_LIB_FILES)
 export GDS_FILES ?= $(PLATFORM_DIR)/gds/sg13g2_stdcell.gds \
                     $(ADDITIONAL_GDS)
 
+export PLATFORM_TCL ?= $(PLATFORM_DIR)/suppress_message.tcl
+
 # Dont use cells to ease congestion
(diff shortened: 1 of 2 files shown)


dnltz:
@maliberty do you seen any issue with disabling this warning?

maliberty:
@osamahammad21 I don't think we care about DRC rules on the contact layer.  Should we suppress this at the source rather than disabling it in ORFS?

dnltz:
@maliberty @osamahammad21 BTW this warning comes from this line: https://github.com/The-OpenROAD-Project/OpenROAD/blob/d7b40432ab60dffa02d475c784865afb52f2da5f/src/drt/src/io/io.cpp#L2079

maybe we should add a check for LEF58_ENCLOSURE exists to `!rule->isCutClassValid()`?

dnltz:
Closing this PR. Should be fixed in OpenROAD (hopefully) soon.

PR #3936, opened 2026-02-27
ihp-sg13g2: Generic pad.tcl script

dnltz:
Provide a generic pad.tcl script to create a padfram for the IHP SG13G2 platform. The design specific config.mk has to define all pins for the north, south, east and west bank.

Additionally, switch the I2C GPIO Expander to this script.


flow/platforms/ihp-sg13g2/pad.tcl
```diff
@@ -0,0 +1,41 @@
+set IO_OFFSET [expr { $::env(IO_BONDPAD_SIZE) + $::env(IO_SEALRING_OFFSET) }]
```
maliberty, line 1: On new files please add the usual licensing header (see other .tcl files)

dnltz: I can do that. However, I was unable to find a licensing header in any tcl file. Can you please point me to one?

maliberty: Interestingly I don't see much use either.  Let's leave that for a separate PR then to address it more globally.

maliberty:
[ERROR] synth__design__instance__area__stdcell fail test: 311335.1808 <= 286097.29
needs updating ( `make update_ok` )

dnltz:
> [ERROR] synth__design__instance__area__stdcell fail test: 311335.1808 <= 286097.29 needs updating ( `make update_ok` )

Thanks. Just updated this PR and removed DNM.

flow/platforms/ihp-sg13g2/pad.tcl
dnltz: I'm using `REUSE` [1] in my projects. It allows to lint for SPDX license headers in files and throws an error if a) the header is missing or b) it's the wrong license identifier. Blocking GPLv3 in a BSD project, for example.

It's running in a GitHub action to ensure all license headers are OK: https://github.com/aesc-silicon/i2c-gpio-expander/blob/main/.github/workflows/license-check.yaml

We could start rolling out this check for .tcl files only at the beginning.

1: https://pypi.org/project/reuse/

maliberty: That's interesting but needs a bit more study.  We are using SPDX-License-Identifier but not SPDX-FileCopyrightText so that might need global updating as well.

maliberty (approved):

PR #3937, opened 2026-02-27
flow: scripts: variables: Fix KEPLER_FORMAL_EXE export

dnltz:
Fix KEPLER_FORMAL_EXE export overriding user assignments 'export VAR := $(VAR)' re-assigns the variable with simple expansion, causing the ?= default to always win over user-provided values. Change to plain 'export' to only mark it for export without reassigning.

### DIFF
--- a/flow/scripts/variables.mk
+++ b/flow/scripts/variables.mk
@@ -118,7 +118,7 @@ KLAYOUT_DIR = $(abspath $(FLOW_HOME)/../tools/install/klayout/)
 KLAYOUT_BIN_FROM_DIR = $(KLAYOUT_DIR)/klayout
 
 KEPLER_FORMAL_EXE ?= $(abspath $(FLOW_HOME)/../tools/install/kepler-formal/bin/kepler-formal)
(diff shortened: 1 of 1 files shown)


maliberty (approved):

Kimplul:
Sorry to necrobump, but in what context does the default win over user-provided values? I can't replicate, `make KEPLER_FORMAL_EXE=whatever ` picks up `whatever` as I would expect it to.

maliberty:
We use `?=` for default assignment and it should follow the normal rules of `make`.

Kimplul:
Yes, if `KEPLER_FORMAL_EXE` is not defined as an environment variable or passed via the command line to `make`, then the default value is used, otherwise the user-provided value is used, as expected. I'm more confused about how the latter re-definition with `:=` would affect whether the default or user-provided value is used and what problem is solved by removing it in this case. `?=` followed by `:=` is a pattern to make the default value immediately expanded instead of it being deferred, which is semantically meaningful but not in a way that is described in the commit message, AFAICT.

PR #3940, opened 2026-03-01
[AutoTuner] Replace failed metric string "-" with ERROR_METRIC

### LINKED ISSUE #3482: Change AutoTuner failed metrics from "-" to ERROR_METRIC
### Description

Ran into this while building out the RTL designer demo.

Tensorboard seems to get "confused" when the trials are a mix of successful and failed runs. The resulting behavior is that important metrics like effective_clk_period and die_size don't appear in the HPARAMs results. The previous workaround of touching all the files to change the order in which tensorboard reads the files didn't work for me. So, I ended up pruning out all the variant directories for the failed runs. Also, including tensorflow in the dependency list (suggested in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3309) didn't work all the time either.

Based on this, I'm guessing that tensorboard gets confused about the data type for the metric, so it throws the metric away (even for the successful runs). Some of the trials have a number for the metric and some trials have a "-", so it doesn't know what to do with them.

### Suggested Solution

My proposal is to change the AutoTuner evaluate() methods to return ERROR_METRIC instead of "-" for any metric. That way, the metric data type is the same, regardless of success or failure.

The downstream impact is that the HPARAMs Parallel Coordinates scales can be off, but this can be addressed by limiting the Max value for the metrics column.

### Additional Context

_No response_

harsh-kumar-patwa:
Fixes #3482

When AutoTuner trials fail, sub-metrics (`effective_clk_period`, `num_drc`, `die_area`) were returned as the string `"-"`. This caused TensorBoard HPARAMS to silently drop entire metric columns because of mixed string/numeric data types across trials.

### Changes

- `utils.py` — `calculate_score()`: return `ERROR_METRIC` instead of `"-"` for all sub-metrics on error/not-found
- `distributed.py` — `AutoTunerBase.step()`: return `ERROR_METRIC` instead of `"-"` for invalid config, and add missing `die_area` key to match the success path
- `distributed.py` — `PPAImprov.evaluate()`: return `ERROR_METRIC` instead of `"-"` for all sub-metrics on error/not-found

### Why this works

`ERROR_METRIC` (`9e99`) is already defined in `utils.py` and used for the primary score on failure. The `TensorBoardLogger.log_sweep_metrics()` method uses `isinstance(value, (int, float))` checks — with `"-"` (string), failed metrics were silently skipped; with `ERROR_METRIC` (float), they are now properly logged.

### Trade-off

As noted in #3482, `9e99` will distort TensorBoard Parallel Coordinates axis scales for failed runs. Users can adjust the Max slider per axis to filter these out.

### Testing

- Unit tested `calculate_score()` for ERR, N/A, and success cases — all return consistent numeric types
- Verified `isinstance` compatibility with `TensorBoardLogger` 
- Black formatting passes

@jeffng-or @luarss

### DIFF
--- a/tools/AutoTuner/src/autotuner/distributed.py
+++ b/tools/AutoTuner/src/autotuner/distributed.py
@@ -147,7 +147,12 @@ class AutoTunerBase(tune.Trainable):
 
         # if not a valid config, then don't run and pass back an error
         if not self.is_valid_config:
-            return {METRIC: ERROR_METRIC, "effective_clk_period": "-", "num_drc": "-"}
+            return {
+                METRIC: ERROR_METRIC,
+                "effective_clk_period": ERROR_METRIC,
+                "num_drc": ERROR_METRIC,
+                "die_area": ERROR_METRIC,
+            }
         self._variant = f"{self.variant}-{self.step_}"
         metrics_file = openroad(
             args=args,
@@ -242,7 +247,7 @@ class PPAImprov(AutoTunerBase):
         error = "ERR" in metrics.values() or "ERR" in reference.values()
         not_found = "N/A" in metrics.values() or "N/A" in reference.values()
         if error or not_found:
-            return (ERROR_METRIC, "-", "-", "-")
+            return (ERROR_METRIC, ERROR_METRIC, ERROR_METRIC, ERROR_METRIC)
         ppa = self.get_ppa(metrics)
         gamma = ppa / 10
         score = ppa * (self.step_ / 100) ** (-1) + (gamma * metrics["num_drc"])
--- a/tools/AutoTuner/src/autotuner/utils.py
+++ b/tools/AutoTuner/src/autotuner/utils.py
@@ -79,7 +79,7 @@ def calculate_score(metrics, step=1):
     not_found = "N/A" in metrics.values()
 
     if error or not_found:
-        return (ERROR_METRIC, "-", "-", "-")
+        return (ERROR_METRIC, ERROR_METRIC, ERROR_METRIC, ERROR_METRIC)
 
     effective_clk_period = metrics["clk_period"] - metrics["worst_slack"]
     num_drc = metrics["num_drc"]


jeffng-or:
Hi Harsh,

Great, thanks! I defer to Jack on the review.

harsh-kumar-patwa:
> Hi Harsh,
> 
> Great, thanks! I defer to Jack on the review.

Hi Jeff,

Thanks for the update! Sounds good. I’ll wait to hear from Jack regarding the review.
Appreciate it.

luarss (commented):
## Review

The core fix is correct and addresses the root cause. Summary of observations:

### What's good

- **Fix is sound.** `TensorBoardLogger.log_sweep_metrics()` gates scalar writes behind `isinstance(value, (int, float))` — passing `"-"` caused silent skips, `ERROR_METRIC = 9e99` (float) will now log consistently.
- **Bonus bug fix.** The invalid-config path in `AutoTunerBase.step()` previously returned `{METRIC: ERROR_METRIC, "effective_clk_period": "-", "num_drc": "-"}` — missing `die_area`. The success path always includes it. Adding `die_area: ERROR_METRIC` fixes the key mismatch in Ray Tune's result dict.
- **Consistent with existing sentinel pattern.** `ERROR_METRIC` was already used for the primary `score` on failure; using it for sub-metrics is idiomatic.

### Nits

1. **Stale type annotations** in `tensorboard_logger.py:30–32`: `Union[float, str]` for `effective_clk_period`, `num_drc`, `die_area` — since callers no longer pass strings, these can be narrowed to `float`/`int`.

2. **No tests committed.** The PR mentions local unit testing of `calculate_score()`, but nothing lands in `tools/AutoTuner/test/`. A small test file verifying that ERR/N/A inputs return a 4-tuple of `ERROR_METRIC` floats would prevent regressions.

### Trade-off

The `9e99` axis distortion in TensorBoard Parallel Coordinates is a known and acceptable trade-off. Silent data loss (current behavior) is worse than visible outliers users can filter with the axis slider.

luarss (approved):
LGTM

maliberty (approved):

PR #3948, opened 2026-03-03
ensure final odb contains parasitics information and log extract command

AcKoucher:
Currently, `6_final.odb` does not contain the parasitics extracted by RCX.

I'm creating this as a draft, because the changes here also suppress the routing obstructions data from the final odb and I'm not sure whether that is okay or not.

### DIFF
--- a/flow/scripts/final_report.tcl
+++ b/flow/scripts/final_report.tcl
@@ -8,8 +8,6 @@ set_propagated_clock [all_clocks]
 # Ensure all OR created (rsz/cts) instances are connected
 global_connect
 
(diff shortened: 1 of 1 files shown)


maliberty:
How much does this increase the file size on a bigger design?

AcKoucher:
> How much does this increase the file size on a bigger design?

Considerably. I tested with a very large private design and it went from 4.14 GB (final) to 5.12 GB (final with parasitics).

AcKoucher:
We can also call `extract_parasitics` in write_net_rc_script.tcl rather than having to rely on the data of `6_final.odb`.

maliberty:
It runs quickly so I think that is a better option

PR #3967, opened 2026-03-08
[slang] Automatically blackbox macros from ADDITIONAL_LIBS

### LINKED ISSUE #3849: [slang] Automatically blackbox macros from ADDITIONAL_LIBS
### Description

I have to manually list macros from ADDITIONAL_LIBS in SYNTH_BLACKBOXES for slang.

### Suggested Solution

Should be automatic as without slang.

### Additional Context

_No response_

oharboe:
@povik I tested Claude to see if it could fix this problem. Is this helpful?

When using the slang HDL frontend, macros defined in ADDITIONAL_LIBS liberty files were not automatically blackboxed. If the same module had a competing Verilog definition in the source files, slang would elaborate the Verilog version instead of using the liberty view.

Fix by extracting cell names from ADDITIONAL_LIBS and passing them as --blackboxed-module arguments to slang, making the behavior consistent with the builtin Verilog frontend.

Tested:

  asap7/uart with a fake ADDITIONAL_LIBS containing uart_tx:
    Confirmed slang blackboxes uart_tx from the liberty file
    instead of elaborating it from the Verilog source.

  asap7/cva6 (which uses slang and ADDITIONAL_LIBS with fakeram
    macros that also have .sv definitions in VERILOG_FILES):
    make DESIGN_CONFIG=designs/asap7/cva6/config.mk do-yosys-canonicalize
    Passes in ~5s with no errors.

  asap7/uart without ADDITIONAL_LIBS:
    Normal synthesis still passes (no regression).

Fixes #3849

### DIFF
--- a/flow/scripts/synth_preamble.tcl
+++ b/flow/scripts/synth_preamble.tcl
@@ -63,6 +63,16 @@ proc read_design_sources { } {
       lappend slang_args -G "$key=$value"
     }
 
+    # Automatically blackbox macros from ADDITIONAL_LIBS so that
+    # any competing Verilog definitions in the source files are
+    # ignored in favor of the liberty view, consistent with the
+    # behavior of the builtin Verilog frontend.
+    if { [env_var_exists_and_non_empty ADDITIONAL_LIBS] } {
+      foreach m [get_liberty_cell_names $::env(ADDITIONAL_LIBS)] {
+        lappend slang_args --blackboxed-module "$m"
+      }
+    }
+
     # Apply module blackboxing based on module names as they appear
     # in the input, that is before any module name mangling done
     # by elaboration and synthesis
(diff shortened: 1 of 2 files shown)


povik:
@oharboe I have checked the CI logs to understand why your branch is failing on nangate45/mempool_group. It's because on master the design is impacted by #3849 and fakerams are not instantiated coming out of synthesis.

placement master
image

placement your branch
image

povik (changes requested):
We can ask yosys for the cell names which is more robust than the regex. I have the change locally but seem unable to push it to your branch due to permissions. Please can you fetch it from https://github.com/povik/OpenROAD-flow-scripts/tree/claude-fixing

oharboe:
@povik I'm closing this PR and letting you run with it.

PR #3977, opened 2026-03-10
Add -bazel flag to skip OpenROAD build dependencies

oharboe:
Add -bazel option to DependencyInstaller.sh and setup.sh that skips OpenROAD-specific build dependencies (boost, swig, spdlog, eigen, or-tools, etc.) which Bazel manages, while still installing flow-level dependencies and building Yosys/eqy/sby.

This supports the new `bazelisk run //:setup-orfs` target in OpenROAD which provides a single entry point for developer environment setup.

needs https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3976

### DIFF
--- a/docs/user/BuildLocally.md
+++ b/docs/user/BuildLocally.md
@@ -14,9 +14,47 @@ sudo ./setup.sh
 
 ## Using Bazel to build OpenROAD and run the ORFS flow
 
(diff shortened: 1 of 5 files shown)


maliberty:
I don't understand why you are moving the compilation of yosys (etc) into the dependency installer.  They are already covered in build_openroad.sh.  Your version is incomplete (eg no verific or slang support).  Why reimplement this?

PR #4000, opened 2026-03-15
Fix false GDS dependency in generate_abstract target

oharboe:
generate_abstract.tcl reads ODB and SDC files to produce LEF/LIB abstracts — it does not use GDS at all. The dependency on 6_final.gds forced unnecessary GDS generation (requiring KLayout) before abstract generation could run.


oharboe:
waiting for https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/4001

PR #4001, opened 2026-03-15
Klayout dependency

oharboe:
## Make KLayout an optional dependency

KLayout is only needed for GDS generation, DRC, and LVS. Most ORFS
use cases — education, architectural exploration, CI testing, OpenROAD
development — never need GDS output. This PR decouples KLayout so that
`make finish` and `do-finish` / `do-final` complete without it.

### What changes

- `finish` / `do-finish` no longer depend on GDS generation
- New `make gds` / `do-gds` target for explicit GDS generation
  (follows existing pattern: place/do-place, finish/do-finish, gds/do-gds)
- `check-klayout` guard on all KLayout-dependent targets with a
  helpful error message when KLayout is missing
- `do-klayout` / `do-klayout_wrap` sed-based XML manipulation replaced
  with `util/generate_klayout_tech.py` (stdlib, no KLayout dependency)
- `def2stream.py` refactored into a testable `merge_gds()` function
  with guarded `import pya`
- 34 new unit tests using `unittest.mock` — run without KLayout
- CI path triggers widened to `flow/util/*.py` and `flow/test/test_*.py`

### Backwards compatibility

- `make finish gds` produces GDS exactly as before
- `do-gds` is unchanged — bazel-orfs can call it from a new
  `orfs_gds()` rule
- All existing scripts that run `make finish` will complete without
  KLayout; they simply no longer attempt GDS generation

### bazel-orfs

`do-final` no longer requires KLayout in the toolchain. A separate
`orfs_gds()` Bazel rule calling `do-gds` makes KLayout an optional
dependency configured in `MODULE.bazel`.

### Test plan

- [ ] `cd flow/test && python -m unittest test_generate_klayout_tech test_def2stream test_convertDrc` (34 tests, no KLayout)
- [ ] `make finish DESIGN_CONFIG=designs/nangate45/gcd/config.mk` completes without KLayout
- [ ] `make gds DESIGN_CONFIG=designs/nangate45/gcd/config.mk` fails with helpful message when KLayout is missing
- [ ] `make finish gds DESIGN_CONFIG=designs/nangate45/gcd/config.mk` produces GDS when KLayout is installed

### DIFF
--- a/flow/docs/KLayoutOptionalDependency.md
+++ b/flow/docs/KLayoutOptionalDependency.md
@@ -11,8 +11,6 @@ without KLayout installed.
 |---|---|---|
 | `make finish` | No | Complete flow through timing reports |
 | `make gds` | Yes | Generate GDS/OAS from finished design |
-| `make do-finish` / `make do-final` | No | Same as `finish`, for bazel-orfs |
-| `make do-gds` | Yes | Same as `gds`, for bazel-orfs |
 | `make drc` | Yes | Run DRC checks (requires GDS) |
 | `make lvs` | Yes | Run LVS checks (requires GDS) |
 | `make gallery` | Yes | Generate layout screenshots |
@@ -26,19 +24,6 @@ missing and a KLayout-dependent target is invoked:
 Error: KLayout not found. Install KLayout or set KLAYOUT_CMD.
 Hint: 'make finish' works without KLayout. Only GDS/DRC/LVS need it.
 ```
-
-## Use Cases
-
-| Use Case | Needs KLayout? |
-|---|---|
-| Education / learning digital design | No |
-| Architectural exploration (PPA analysis) | No |
-| CI testing of ORFS / OpenROAD | No |
-| Local ORFS and OpenROAD development | No |
-| bazel-orfs default `orfs_flow()` | No |
-| GDS generation for tapeout | Yes |
-| DRC / LVS verification | Yes |
-| Layout visualization in KLayout GUI | Yes |
 
 ## bazel-orfs Integration
 
--- a/flow/docs/KLayoutOptionalDependency.md
+++ b/flow/docs/KLayoutOptionalDependency.md
@@ -9,7 +9,7 @@ without KLayout installed.
 
 | Target | Requires KLayout | Description |
 |---|---|---|
-| `make finish` | No | Complete flow through timing reports |
+| `make finish` | Yes | Complete flow including GDS generation |
 | `make gds` | Yes | Generate GDS/OAS from finished design |
 | `make drc` | Yes | Run DRC checks (requires GDS) |
 | `make lvs` | Yes | Run LVS checks (requires GDS) |
@@ -22,7 +22,7 @@ missing and a KLayout-dependent target is invoked:
 
 ```
 Error: KLayout not found. Install KLayout or set KLAYOUT_CMD.
-Hint: 'make finish' works without KLayout. Only GDS/DRC/LVS need it.
+Hint: KLayout is needed for GDS/DRC/LVS targets.
 ```
 
 ## bazel-orfs Integration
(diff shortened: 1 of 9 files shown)


oharboe:
@hzeller @maliberty TL;DR merge this and only users and developers who need klayout have to install it.

flow/docs/KLayoutOptionalDependency.md
```diff
@@ -0,0 +1,89 @@
...
+without KLayout installed.
+
+## Makefile Targets
+
+| Target | Requires KLayout | Description |
+|---|---|---|
+| `make finish` | No | Complete flow through timing reports |
+| `make gds` | Yes | Generate GDS/OAS from finished design |
+| `make do-finish` / `make do-final` | No | Same as `finish`, for bazel-orfs |
+| `make do-gds` | Yes | Same as `gds`, for bazel-orfs |
```
maliberty: The do targets are just for b-orfs and don't need to be documented for users

flow/docs/KLayoutOptionalDependency.md
```diff
@@ -0,0 +1,89 @@
...
+| Use Case | Needs KLayout? |
+|---|---|
+| Education / learning digital design | No |
+| Architectural exploration (PPA analysis) | No |
+| CI testing of ORFS / OpenROAD | No |
+| Local ORFS and OpenROAD development | No |
+| bazel-orfs default `orfs_flow()` | No |
+| GDS generation for tapeout | Yes |
+| DRC / LVS verification | Yes |
+| Layout visualization in KLayout GUI | Yes |
```
maliberty: Redundant to the paragraph in # KLayout as an Optional Dependency

maliberty (changes requested):
I would rather not change the default behavior.  Perhaps a new target like `signoff` could be used for your purpose.

oharboe:
> I would rather not change the default behavior. Perhaps a new target like `signoff` could be used for your purpose.

Well spotted, thanks!

Fixed.

maliberty:
hmm... I don't see a new target

oharboe:
> hmm... I don't see a new target

bazel-orfs already has do-gds broken out, so I am good.

flow/docs/KLayoutOptionalDependency.md
```diff
@@ -0,0 +1,74 @@
...
+Error: KLayout not found. Install KLayout or set KLAYOUT_CMD.
+Hint: KLayout is needed for GDS/DRC/LVS targets.
+```
+
+## bazel-orfs Integration
+
+bazel-orfs uses the `do-` prefixed targets which bypass Make's dependency
+management. The relevant targets follow the standard pattern:
+
+- `do-finish` / `do-final` — runs the finish stage without GDS
```
maliberty: How can do-finish not include GDS while finish does?

flow/docs/KLayoutOptionalDependency.md
```diff
@@ -0,0 +1,74 @@
...
+completed, it would eliminate the KLayout dependency for GDS generation
+in many cases. However, KLayout support will live alongside `write_gds`
+until there is confidence that `write_gds` is robust and fully covers
+all current use cases.
+
+Even then, non-public or future PDKs may have requirements that benefit
+from KLayout's mature GDS handling. Whether KLayout support stays in
+ORFS indefinitely remains to be seen. Once `write_gds` works robustly
+for all of today's use cases, extending it to cover future ones may
+turn out to be straightforward.
```
maliberty: Unnecessary.  Let's document it when it becomes relevant as nobody is working on it actively.

flow/docs/KLayoutOptionalDependency.md
maliberty: I see do-finish isn't equivalent to finish already.

oharboe: Yes: do-finish no longer includes do-gds now, whereas "finish" is left alone. The do- form is composable, the non-do versions are compatible. So that's the compromise.

maliberty (approved):

PR #4008, opened 2026-03-16
Claude procedure to hunt down non-determinism in pr-head

oharboe:
Found and reported this issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/9781

If you have a pr-head with non-determinism, run:

    /diagnose-pr-head <github-pr-url>


oharboe:
@vvbandeira Just FYI on how I dig into non determinism

PR #4019, opened 2026-03-20
Use yosys stat -json for metrics extraction

### LINKED ISSUE #3485: Use yosys stat -json for metrics extraction
### Description

Yosys has support to generate JSON stats, which are better for scripting and don't break when the log output format changes.

### Suggested Solution

Use stats -json

### Additional Context

_No response_

rohithsiddi:
# Use yosys stat -json for metrics extraction

Fixes #3485

## Description

Yosys `stat -json` produces structured JSON output that is more reliable for scripting than the plain-text `stat` output. This replaces the text-based `synth_stat.txt` report and the fragile regex patterns used to parse it with a proper JSON file parsed in Python.

## Changes

- `flow/scripts/synth.tcl`: replace `stat` with `stat -json`, writing output to `synth_stat.json`
- `flow/util/genMetrics.py`: parse `synth_stat.json` with `json.load()` instead of using regex on `synth_stat.txt`; skip non-JSON prefix (e.g. "Found gzip magic in file...") that yosys emits before JSON output when reading `.lib.gz` liberty files
- `docs/tutorials/FlowTutorial.md`: update references from `synth_stat.txt` to `synth_stat.json`

## Testing

Tested locally against `nangate45/gcd`, `asap7/gcd`, and `gf180/aes`. asap7 and gf180 use `.lib.gz` liberty files which cause yosys to emit log messages before the JSON output; the parser handles this correctly.

| Metric | nangate45/gcd | asap7/gcd | gf180/aes |
|--------|---------------|-----------|-----------|
| `synth__design__instance__count__stdcell` | 425 | 307 | 15336 |
| `synth__design__instance__area__stdcell` | 562.324 | 35.663 | 499940.179 |

AI tools were used to assist with code generation. All technical decisions and implementation details are my own.

### DIFF
--- a/flow/scripts/synth.tcl
+++ b/flow/scripts/synth.tcl
@@ -226,7 +226,7 @@ if { $::env(SYNTH_INSBUF) } {
 # Reports
 tee -o $::env(REPORTS_DIR)/synth_check.txt check
 
-tee -o $::env(REPORTS_DIR)/synth_stat.txt stat {*}$lib_args
+tee -o $::env(REPORTS_DIR)/synth_stat.json stat -json {*}$lib_args
 
 # check the design is composed exclusively of target cells, and
 # check for other problems
--- a/flow/util/genMetrics.py
+++ b/flow/util/genMetrics.py
@@ -234,7 +234,10 @@ def extract_metrics(
     synth_stat_file = rptPath + "/synth_stat.json"
     try:
         with open(synth_stat_file) as f:
-            synth_stat = json.load(f)
+            content = f.read()
+        # yosys may emit log messages (e.g. "Found gzip magic in file...")
+        # before the JSON when reading compressed liberty files. Skip them.
+        synth_stat = json.loads(content[content.index("{") :])
         modules = synth_stat.get("modules", {})
         # Prefer lookup by design name; yosys may prefix module names with a
         # backslash in JSON output, so try both forms. Fall back to the last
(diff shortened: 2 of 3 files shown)


flow/scripts/synth.tcl
```diff
@@ -226,7 +226,7 @@ if { $::env(SYNTH_INSBUF) } {
 # Reports
 tee -o $::env(REPORTS_DIR)/synth_check.txt check
 
-tee -o $::env(REPORTS_DIR)/synth_stat.txt stat {*}$lib_args
+tee -o $::env(REPORTS_DIR)/synth_stat.json stat -json {*}$lib_args
```
vvbandeira, line 229: We should tee to a log file and use the `-o` option to output a valid JSON:
https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/index_backends.html#json-write-design-in-json-format

flow/util/genMetrics.py
```diff
@@ -231,20 +231,50 @@ def extract_metrics(
...
-        metrics_dict,
-        "^\\s+(\\d+)\\s+[-0-9.]+\\s+cells$",
-        rptPath + "/synth_stat.txt",
-    )
+    synth_stat_file = rptPath + "/synth_stat.json"
+    try:
+        with open(synth_stat_file) as f:
+            content = f.read()
+        # yosys may emit log messages (e.g. "Found gzip magic in file...")
+        # before the JSON when reading compressed liberty files. Skip them.
```
vvbandeira, line 239: If the above is implemented we likely will not need this post-processing.

vvbandeira (changes requested):

flow/scripts/synth.tcl
rohithsiddi: The json backend (`write_json -o`) outputs the design netlist rather than synthesis statistics, it doesn't include aggregated cell count or area values that we need for metrics. `stat -json` provides these but doesn't support `-o`; its docs recommend using `tee` for file output. Our current approach uses `tee -o`to capture `stat -json` and skips any log prefix (e.g. "Found gzip magic") in `genMetrics.py` before parsing. Would this be acceptable?

Issue #3969, opened 2026-03-09
The model used for estimating `set_layer_rc` and `set_wire_rc` values should be improved


AcKoucher:
### Context

Today we use a single RC value per layer for estimating parasitics. These values are computed based on a series of scripts that compile RCX information for a certain technology into a rough estimation. Here’s what happens today:

- Run many designs from one PDK and record for each net:
  - cap & res;
  - wire length per layer;
- Formulate a linear regression model.

### General Problem

The linear regression model, as it is formulated today, gives us an average over different contexts, but also “hides” differences in topology and layer assignment which may lead to inconsistencies in the parasitics estimated at global routing.

### Suggested Solution

 - [ ] **Improve the Current Regression Model**: Instead of relying on the total RC values of each individual net, formulate the linear regression model from the segments of each net so that the increased granularity allows a more precise approximation. For that, use the .spef file data that can be kept inside ODB.

 - [ ] **Make the Model Context-Aware**: Instead of having a single RC value per layer, having a table based on density, so that, when EST is working based on GRT guides, it can choose the RC values based on the density context.

PR #4021, opened 2026-03-20
fetch segments rc also from nets which are not top-level

AcKoucher:
Related to #3969.

asap7/swerv_wrapper is using `-hier`. Without the changes here we end up not considering the vast majority of the segments for this design.


eder-matheus (approved):

eder-matheus:
@AcKoucher check the CI failures.

Issue #4022, opened 2026-03-20
macOS fail across different stages of dependency installation and build locally


Sahil7741:
### Subject

[Build] for compilation-related issues.

### Describe the bug

Running `sudo ./setup.sh` on macOS (Apple Silicon) results in multiple failures across different stages of dependency installation and build locally.

Script includes partial macOS support, overall setup process does not complete successfully due to a combination following

- Linux specific command (eg - `nproc`)

- Inconsistent privilege expectations, sudo vs non root on macOS 

- Build failures in dependencies (eg - `lemon-graph` via CMake)

- Toolchain incompatibilities with modern macOS/Homebrew environments

### Expected Behavior

`sudo ./setup.sh` should complete successfully on macOS.

### Environment

```shell
sahil@Sahils-MacBook-Air:~/Developer/OpenROAD-flow-scripts/tools/OpenROAD $ ./etc/Env.sh

[WARNING] Your current OpenROAD version is outdated.
It is recommened to pull the latest changes.
If problem persists, file a github issue with the re-producible test case.
kernel: Darwin 25.3.0
os:             macOS           26.3.1
cmake version 4.3.0
CMake Deprecation Warning at CMakeLists.txt:18 (cmake_policy):
  The OLD behavior for policy CMP0144 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.


CMake Error at /opt/homebrew/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:290 (message):
  Could NOT find GTest (missing: GTEST_LIBRARY GTEST_INCLUDE_DIR
  GTEST_MAIN_LIBRARY)
Call Stack (most recent call first):
  /opt/homebrew/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:654 (_FPHSA_FAILURE_MESSAGE)
  /opt/homebrew/share/cmake/Modules/FindGTest.cmake:421 (find_package_handle_standard_args)
  src/CMakeLists.txt:154 (find_package)


-- The CXX compiler identification is AppleClang 17.0.0.17000604
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
[...]

PR #4023, opened 2026-03-20
Improve macOS handling in setup and dependency installer scripts

Sahil7741:
This PR improves macOS behavior in the ORFS setup path:

- Allow `sudo` on Linux and disallow on macOS
- Fix submodule update to run with correct user per OS
- Use `sysctl` for CPU count on macOS instead of `nproc`
- Minor portability and error message improvements

Addresses #4022

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -5,18 +5,12 @@ set -euo pipefail
 # allow this script to be invoked from any folder
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-# macOS detection
-IS_DARWIN=false
-if [[ "$(uname)" == "Darwin" ]]; then
-  IS_DARWIN=true
-fi
-
-if $IS_DARWIN && [[ $EUID -eq 0 ]]; then
+if [[ "$OSTYPE" == "darwin"* ]] && [[ $EUID -eq 0 ]]; then
   echo "Do NOT run this script with sudo on macOS"
   exit 1
 fi
 
-if ! $IS_DARWIN && [[ $EUID -ne 0 ]]; then
+if [[ "$OSTYPE" != "darwin"* ]] && [[ $EUID -ne 0 ]]; then
   echo "This script must be run with sudo on Linux"
   exit 1
 fi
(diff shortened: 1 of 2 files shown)


setup.sh
```diff
@@ -5,8 +5,19 @@ set -euo pipefail
...
 
-if [ $EUID -ne 0 ]; then
-  echo "This script must be run with sudo"
+# macOS detection
+IS_DARWIN=false
+if [[ "$(uname)" == "Darwin" ]]; then
+  IS_DARWIN=true
+fi
+
+if $IS_DARWIN && [[ $EUID -eq 0 ]]; then
```
luarss: why not sudo on macos?

setup.sh
```diff
@@ -5,8 +5,19 @@ set -euo pipefail
 # allow this script to be invoked from any folder
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-if [ $EUID -ne 0 ]; then
-  echo "This script must be run with sudo"
+# macOS detection
+IS_DARWIN=false
+if [[ "$(uname)" == "Darwin" ]]; then
```
luarss: maybe we can follow existing coding style, it should be okay to repeat `if [[ "$OSTYPE" == "darwin"* ]]; then` multiple times

setup.sh
Sahil7741: Actually `setup.sh` invokes `DependencyInstaller.sh` and dependencies are installed via  Homebrew which do not need sudo. Additionally `DependencyInstaller.sh` is build in such a way that on mac its recommend to not use sudo. If we run it using sudo then script fails and exits giving below error.
Look Line 410 of `DependencyInstaller.sh`
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/8ad0ff7795bab4e6d94bcf0295fe1b451ea637d1/etc/DependencyInstaller.sh#L410

setup.sh
Sahil7741: Thanks, I did the changes please check

maliberty (approved):

PR #4024, opened 2026-03-21
setup: skip sudo requirement on macOS for Homebrew compatibility

Divinesoumyadip:
On macOS, Homebrew explicitly recommends against running with sudo.The existing DependencyInstaller.sh already blocks sudo on macOS ,but `setup.sh `was requiring sudo unconditionally, causing a conflict.

This fix skips the sudo requirement on macOS only.


maliberty:
conflict to resolve

vvbandeira:
Fixed on another PR.

PR #4029, opened 2026-03-22
setup: fix unbound SUDO_USER variable when running as root in Docker

Divinesoumyadip:
When `setup.sh` is run directly as root , which is standard in Docker containers and CI/CD pipelines so,`$SUDO_USER` is never set. Because the script uses `set -euo pipefail`, the `-u` flag causes the script to crash immediately on any unbound variable reference.

Two lines were affected:

**Line 28:**` bash sudo -u $SUDO_USER git submodule update --init --recursive`

**Line 42:**` sudo -u $SUDO_USER "$DIR/etc/DependencyInstaller.sh" -common -prefix="$DIR/dependencies"`

When `$SUDO_USER` is empty, `sudo -u `consumes` git `(or the script path) as the username and crashes with:
```
sudo: unknown user: git 
sudo: error initializing audit plugin sudoers_audit
```
To resolve this, the script now checks if `$SUDO_USER` is set before attempting to use it. If the variable is set (indicating the script was invoked via sudo), it drops privileges to the original user as intended. However, if it is empty (indicating the script is running natively as root), it simply runs the command directly without the `sudo -u` flag. The macOS paths were already handled correctly by #4023, so this PR completes the fix for the remaining Linux and Docker environments.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -25,7 +25,11 @@ if grep -q "^-" "$tmpfile"; then
   if [[ "$OSTYPE" == "darwin"* ]]; then
     git submodule update --init --recursive
   else
-    sudo -u $SUDO_USER git submodule update --init --recursive
+    if [[ -n "${SUDO_USER:-}" ]]; then
+      sudo -u "$SUDO_USER" git submodule update --init --recursive
+    else
+      git submodule update --init --recursive
+    fi
   fi
 elif grep -q "^+" "$tmpfile"; then
   # Make it easy for users who are not hacking ORFS to do the right thing,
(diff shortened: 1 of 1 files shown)


Divinesoumyadip:
@maliberty @luarss could you approve the workflows to run CI? Happy to address any review feedback.

setup.sh
```diff
@@ -25,7 +25,11 @@ if grep -q "^-" "$tmpfile"; then
   if [[ "$OSTYPE" == "darwin"* ]]; then
     git submodule update --init --recursive
   else
-    sudo -u $SUDO_USER git submodule update --init --recursive
+    if [[ -n "${SUDO_USER:-}" ]]; then
```
luarss: this nested loop causes lots of repeated code. is a helper function possible here?

Divinesoumyadip: > this nested loop causes lots of repeated code. is a helper function possible here?

Done! Refactored to use a `run_as_user()` helper function that handles both cases drops privileges to the original user when invoked via sudo, and runs directly when already root. Removes the repeated pattern across both submodule and dependency installer calls.

PR #4030, opened 2026-03-22
setup: fix macOS thread count using hw.logicalcpu instead of hw.ncpu

### LINKED ISSUE #4028: [Build] setup: nproc command not found crashes DependencyInstaller.sh on macOS
### Subject

### Describe the bug

> `etc/DependencyInstaller.sh` has `set -euo pipefail` at line 3. Line 14 unconditionally calls `nproc`:
> ```bash
> numThreads=$(nproc)
> ```
> `nproc` is a GNU coreutils command — it does not exist on macOS. The script already has a macOS-specific check at line 6 for path resolution, but the thread count detection at line 14 has no equivalent macOS fallback. On macOS, the script crashes immediately.

### Expected Behavior

> The script should detect macOS and use `sysctl -n hw.logicalcpu` instead of `nproc` — exactly the same pattern already used in `Build.sh` in the OpenROAD submodule.

### Environment

```shell
> OS: macOS
> File: `etc/DependencyInstaller.sh`
> Line: 14
```

### To Reproduce

> Run `./setup.sh` on a fresh macOS machine after a clean clone.

### Relevant log output

```shell
> 
> etc/DependencyInstaller.sh: line 14: nproc: command not found
>
```

### Additional Context

> **Proposed Fix:**
> Replace line 14 with a platform-aware thread count detection:
> ```bash
> if [[ "$OSTYPE" == "darwin"* ]]; then
>   numThreads=$(sysctl -n hw.logicalcpu)
> else
>   numThreads=$(nproc)
> fi
> ```
> 
> *Note: `hw.logicalcpu` is used rather than `hw.ncpu`. `hw.logicalcpu` returns all logical cores including hyperthreaded cores — the correct equivalent of `nproc` on Linux. `hw.ncpu` returns only physical cores and underestimates available parallelism on Apple Silicon and Intel Macs with hyperthreading.*

Divinesoumyadip:
Fixes #4028 

Currently, `etc/DependencyInstaller.sh` uses `sysctl -n hw.ncpu` to determine the thread count on macOS. However, `hw.ncpu` only returns the number of *physical* cores. This ends up underestimating the available parallelism, particularly on Intel Macs that utilize hyperthreading. 

This PR replaces `hw.ncpu` with `hw.logicalcpu`. `hw.logicalcpu` correctly returns all logical cores (including hyperthreaded ones), making it the exact macOS equivalent to the `nproc` command on Linux.

- On an Intel Mac with 4 physical cores and hyperthreading, `hw.ncpu` returns `4`, while `hw.logicalcpu` correctly returns `8`.
- On Apple Silicon (e.g., an M2 with 10 cores), both commands return `10`.

This swap ensures we are fully utilizing the available CPU resources to speed up dependency installation on macOS.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -12,7 +12,7 @@ fi
 # package versions
 klayoutVersion=0.30.3
 if [[ "$OSTYPE" == "darwin"* ]]; then
-    numThreads=$(sysctl -n hw.ncpu)
+    numThreads=$(sysctl -n hw.logicalcpu)
 else
     numThreads=$(nproc)
 fi


Divinesoumyadip:
@maliberty @luarss could you approve the workflows to run CI? Happy to address any review feedback.

luarss (approved):

maliberty (approved):

Issue #4041, opened 2026-03-23
[Flow] Division-by-zero fatal error in report_metrics.tcl for 0-delay paths


Vi-shub:
### Description
While checking more flow scripts, I found there is a potential Fatal TCL Error (crash) inside [flow/scripts/report_metrics.tcl](cci:7://file:///c:/Users/smsha/Desktop/OpenROAD-flow-scripts/flow/scripts/report_metrics.tcl:0:0-0:0).
At line 227, the script calculates the slack division ratio using this line:
```tcl
report_puts "[format "%4f" [expr $path_slack / $path_delay * 100]]"
```

if a design (or an unrouted/ideal testblock) happens to have a critical path delay of exactly 0.0, the TCL interpreter throws a fatal division-by-zero error and completely crashes the metrics reporting workflow.
This causes the entire flow to fail unexpectedly during metric extraction on specific test cases or custom sub-blocks that lack significant path delay.

### Suggested Solution

Simple fix we can put by wrapping the division in a simple protective 
```
tcl
if { $path_delay != 0.0 } {
      report_puts "[format "%4f" [expr $path_slack / $path_delay * 100]]"
    } else {
      report_puts "N/A (0 delay)"
    }

```

### Additional Context

Hi @luarss @maliberty this is small fix, can you please assign it to me. Thankyou for you time :)

maliberty:
Even an unrouted design won't have zero delay.  It won't hurt to fix it but I doubt it ever comes up.

Vi-shub:
Thanks for reviewing and merging the PR! Agreed that a true 0-delay path is highly unlikely in a real routed design, but I noticed it threw a fatal interpreter error during some isolated unit testing with dummy nets, so I figured it was worth hardening the script just in case. Closing this issue since the patch is merged. Thanks again :)

PR #4042, opened 2026-03-23
[Flow] Fix potential division-by-zero crash in report_metrics.tcl

Vi-shub:
- Added a protective check before calculating path_slack / path_delay percentage.
- Prevents TCL fatal errors on designs with 0-delay critical paths.

Hi @luarss and @maliberty this is quick fix for the 0 division error can you review and merge it. Thanks for your time.


Vi-shub:
Closes #4041

maliberty (approved):

PR #4043, opened 2026-03-24
Add files of ASAP7 based IO-cells

jooyeon38:


maliberty:
I would like to have IOs but we can't accept proprietary tool output due to copyright issues.

PR #4045, opened 2026-03-24
fix: APB_UART RX FIFO threshold uses wrong register

ashnaaseth2325-oss:
## 1. SUMMARY

This PR fixes an incorrect register reference in the APB UART module where the RX FIFO threshold logic was comparing against the TX threshold register. This caused RX threshold interrupts to trigger at incorrect levels.

The issue affects `flow/designs/src/chameleon/IPs/APB_UART.v` (around line 145) and impacts correct interrupt generation based on RX FIFO occupancy.

---

## 2. FIX

```verilog
// Before (incorrect threshold comparison)
wire rx_more_threshold = (rx_level > TXFIFOTR);

// After (correct threshold comparison)
wire rx_more_threshold = (rx_level > RXFIFOTR);
```

---

## 3. VERIFICATION

Configure different threshold values (e.g., `TXFIFOTR = 4`, `RXFIFOTR = 8`) and enable the RX threshold interrupt. Fill the RX FIFO gradually and observe interrupt behavior.

Before the fix, the interrupt triggers based on `TXFIFOTR`; after the fix, it correctly triggers only when `rx_level > RXFIFOTR`.

### DIFF
--- a/flow/designs/src/chameleon/IPs/APB_UART.v
+++ b/flow/designs/src/chameleon/IPs/APB_UART.v
@@ -142,7 +142,7 @@ module APB_UART(
                                                         32'hDEADDEAD;
     
     wire tx_less_threshold = (tx_level < TXFIFOTR);
-    wire rx_more_threshold = (rx_level > TXFIFOTR);
+    wire rx_more_threshold = (rx_level > RXFIFOTR);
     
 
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hi @eder-matheus @maliberty,
Please review this PR whenever you get time.
Thanks!

maliberty (approved):

PR #4057, opened 2026-03-26
fix: prevent file descriptor leak in is_git_repo() by using context manager

ashnaaseth2325-oss:
## 1. SUMMARY

This PR fixes a file descriptor leak in `is_git_repo()` in `flow/util/genMetrics.py`. It ensures `os.devnull` is properly closed after use.

---

### Before

```python
return call(cmd, stderr=STDOUT, stdout=open(os.devnull, "w")) == 0
```

### After

```python
with open(os.devnull, "w") as devnull:
    return call(cmd, stderr=STDOUT, stdout=devnull) == 0
```

---

## 3. VERIFICATION

**Steps:**

* Call `is_git_repo()` repeatedly
* Check open FDs (`lsof -p <pid>`)

**Expected:**

* No FD growth over time
* Same behavior as before
* No warnings raised

### DIFF
--- a/flow/util/genMetrics.py
+++ b/flow/util/genMetrics.py
@@ -175,10 +175,11 @@ def read_sdc(file_name):
 
 def is_git_repo(folder=None):
     cmd = ["git", "branch"]
-    if folder is not None:
-        return call(cmd, stderr=STDOUT, stdout=open(os.devnull, "w"), cwd=folder) == 0
-    else:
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hi @maliberty @eder-matheus,
Kindly check this PR.
Thanks!

maliberty (approved):

PR #2735, opened 2025-01-29
asap7/ethmac_lvt: study placement densities and grt

oharboe:
```
bazel run //designs/asap7/ethmac_lvt:plot
```

Builds for ca. 1 hour if you have *lots* of CPUs... and then shows the graph below.

Here we can clearly see that 0.60 is at the brink of grt routing collapse and that we should expect asap/ethmac_lvt to fall over with random changes in initial conditions, such as in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2694

image

PR #4074, opened 2026-03-30
fix: validate plot_congestion.py CLI arguments

mithlohar144:
This PR fixes a small CLI robustness issue in plot_congestion.py.

Problem:

The script accessed positional arguments directly and crashed with IndexError when invoked with too few arguments.
The script also accepted malformed trailing args (odd count) that should be rejected.

### DIFF
--- a/flow/util/plot_congestion.py
+++ b/flow/util/plot_congestion.py
@@ -4,6 +4,7 @@ import sys
 import re
 import os
 
+
 def usage(script):
     print(
         f"Usage: {script} <sweep_name> <output_png> <congestion_file...> <sweep_value...>",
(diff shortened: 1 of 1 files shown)


maliberty (approved):
@oharboe is this file even in use?

maliberty:
Needs to be formatted with `black`

oharboe:
> @oharboe is this file even in use?

Not by me.

maliberty:
@oharboe you created this and a number of other scripts in #2735.  Can we remove all that?

oharboe:
> @oharboe you created this and a number of other scripts in #2735. Can we remove all that?

I have cleanup PRs that I will line up, so perhaps you prefer to wait?

PR #4078, opened 2026-03-30
add vhdl frontend support via ghdl

fredowski:
I added the ghdl frontend to also support vhdl files. I also added a "counter" example for sky130hd and ihp-sg13g2 with VHDL code.

I think this was requested also here: https://github.com/The-OpenROAD-Project/OpenLane/issues/292 for openlane.


maliberty:
I believe ghdl is a plugin and not available by default.  To add this then you need to update the dependency installer to ensure it is available.  I expect it will be a somewhat painful process though maybe AI can help.

fredowski:
Yes, you need the yosys-ghdl plugin for this. I make this a draft.

fredowski:
I did not add the tooling due to other work I have to do. I will reopen this once I have the tooling ready.

PR #4080, opened 2026-03-30
fix: use DESIGN_DIR instead of custom dir variables in chameleon/microwatt

oharboe:
DESIGN_DIR is already defined in the Makefile as $(dir $(DESIGN_CONFIG)). These two designs defined redundant custom variables (chameleon_DIR, microwatt_DIR) that were identical to DESIGN_DIR. Use the standard variable to simplify config and remove non-standard patterns.

### DIFF
--- a/flow/Makefile
+++ b/flow/Makefile
@@ -88,7 +88,7 @@ DESIGN_CONFIG ?= ./designs/nangate45/gcd/config.mk
 export DESIGN_CONFIG
 include $(DESIGN_CONFIG)
 
-export DESIGN_DIR ?= $(dir $(DESIGN_CONFIG))
+export DESIGN_DIR ?= $(patsubst %/,%,$(dir $(DESIGN_CONFIG)))
 
 # default value "base" for FLOW_VARIANT and "." for WORK_HOME are duplicated
 # from variables.yaml and variables.mk because we need it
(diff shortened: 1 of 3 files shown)


maliberty (approved):

oharboe:
Just FYI:

I'm trying to spin of a number of these cleanup issues to whittle down the amount of patches I have to carry in https://github.com/The-OpenROAD-Project/bazel-orfs/tree/main/orfs/patches

As things settle down and I see the exact shape of changes to ORFS and they are stable, I'll start filing various other patches from that list.

oharboe:
@maliberty Why on earth would this cause???

```
21:24:39  [ERROR] globalroute__antenna_diodes_count fail test: 4276 <= 2635
```

maliberty:
I would expect this to be a no-op so something isn't right.  Do the wildcard matches yield the same results?

oharboe:
> I would expect this to be a no-op so something isn't right. Do the wildcard matches yield the same results?

I can't tell these two apart, I haven't tried to investigate. I would have dismissed it as unstable master, but it is from the same design that I changed.

oharboe:
@maliberty I had my guy dig into it.

maliberty (approved):

oharboe:
@maliberty If this works, then doesn't this merit further investigation? Why would diode thingy fail due to an extra '/'?

maliberty:
Locally I ran this change and did not reproduce the problem.  I ran the prior commit and got the same results.  I'm guessing the 'fix' from your guy is just luck rather than anything real and there is a real non-determinism.

oharboe:
> Locally I ran this change and did not reproduce the problem. I ran the prior commit and got the same results. I'm guessing the 'fix' from your guy is just luck rather than anything real and there is a real non-determinism.

Can we deterministically reproduce the non-determinism with this one?

If so, gold.

PR #4091, opened 2026-04-01
fix: override PDN_TCL for riscv32i-mock-sram to fix PDN-0232/0233

oharboe:
The platform default BLOCKS_grid_strategy.tcl defines an ElementGrid macro grid with only an M5-M6 connection rule but no stripes, producing empty macro PDN grids. Use BLOCK_grid_strategy.tcl instead, which has M4-M5 connections matching the block's MAX_ROUTING_LAYER=M4 constraint.

This is the same pattern used by aes-block (via block.mk).


oharboe:
@maliberty If this needs regeneration of the rules.json, could you do it?

maliberty (approved):

PR #4093, opened 2026-04-01
fix: use SDC_FILE_EXTRA for mock-cpu util.tcl sourcing

oharboe:
The constraint.sdc used a hardcoded $::env(DESIGN_HOME) path and io.tcl tried two relative paths with file-exists guards. Both are fragile and break in sandboxed build environments.

Use SDC_FILE_EXTRA to declare the dependency in config.mk and source $::env(SDC_FILE_EXTRA) for a portable path in both files.


oharboe:
Also this will be explicitly supported in bazel-orfs to set up the dependency.

oharboe:
@maliberty Aha! microwatt is flaky, didn't change it in this PR.

    05:49:19  [ERROR] globalroute__antenna_diodes_count fail test: 3224 <= 2635

maliberty:
mock-cpu passed so I assume the other failures are unrelated.

maliberty (approved):

PR #4094, opened 2026-04-01
[RFQ] build: add bazel-orfs beta test for design builds

oharboe:
# bazel-orfs Beta Test

This is an early integration of [bazel-orfs](https://github.com/The-OpenROAD-Project/bazel-orfs)
into OpenROAD-flow-scripts. It lets you build ORFS designs with Bazel
using the same `config.mk` files you already have.

**Status**: beta -- 46 designs across 6 public PDK platforms build and
pass QoR tests. Platforms without public PDK files (gf12, gf55,
rapidus2hp) are not wired up.

## Quick Start

```bash
# Install bazelisk (one-time) -- see https://github.com/bazelbuild/bazelisk
# Then, from the ORFS root:

# Synthesize gcd on asap7 (downloads Docker image on first run)
bazelisk build //flow/designs/asap7/gcd:gcd_synth

# Run through floorplan
bazelisk build //flow/designs/asap7/gcd:gcd_floorplan

# Full flow through final
bazelisk build //flow/designs/asap7/gcd:gcd_final

# Run the test (includes QoR regression check)
bazelisk test //flow/designs/asap7/gcd:gcd_test

# List all targets for a design
bazelisk query //flow/designs/asap7/gcd:all
```

## Available Targets Per Design

Each enabled design gets these targets automatically from its `config.mk`:

| Target suffix | What it does |
|---|---|
| `_synth` | Yosys synthesis |
| `_floorplan` | Floorplan + I/O placement |
| `_place` | Global + detailed placement |
| `_cts` | Clock tree synthesis |
| `_grt` | Global routing |
| `_route` | Detailed routing |
| `_final` | Final optimization + fill |
| `_generate_abstract` | LEF/LIB abstract for hierarchical use |
| `_generate_metadata` | Flow metadata (logs, reports) |
| `_test` | Full flow + QoR regression check |

Each stage depends on the previous one, so building `_final` runs the
entire flow.

## Working Designs (46 passing tests)

All designs below have `orfs_design()` enabled and pass
`bazelisk test` with QoR regression checks.

### asap7 (16 designs)

| Design | DESIGN_NAME | Notes |
|--------|-------------|-------|
| `gcd` | gcd | smoketest, ~1 min |
| `gcd-ccs` | gcd | CCS lib model |
| `aes` | aes_cipher_top | medium |
| `aes-block` | aes_cipher_top | block variant |
| `aes-mbff` | aes_cipher_top | MBFF |
| `aes_lvt` | aes_cipher_top | LVT cells |
[...]

PR #4100, opened 2026-04-02
Yosys CSA tree - demo

hexratcc:
This PR aims to evaluate QoR of https://github.com/YosysHQ/yosys/pull/5753, which introduces a new CSA pass, potentially improving critical path length. Note that the CSA pass runs before the wrapped operator flow (`SWAP_ARITH_OPERATORS`). The CSA pass consumes `$alu`/`$macc` chains, which may reduce the cells available for the wrapped operator flow, so hopefully this doesn't break things. 

cc @maliberty @QuantamHD

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "tools/yosys"]
 	path = tools/yosys
-	url = ../../The-OpenROAD-Project/yosys.git
+	url = https://github.com/YosysHQ/yosys.git
 [submodule "tools/OpenROAD"]
 	path = tools/OpenROAD
 	url = ../OpenROAD.git
--- a/flake.lock
+++ b/flake.lock
@@ -149,18 +149,18 @@
         "nixpkgs": "nixpkgs_3"
       },
       "locked": {
-        "lastModified": 1775061651,
-        "narHash": "sha256-3a+Q9hkR4Ny07xdt2DEd3DM1n1pNy7iJlm7UT5pNbbk=",
-        "ref": "nella/carry-save-adders",
-        "rev": "0909e462c3efc2ecd9395e1d69485f9ad4027857",
-        "revCount": 16821,
+        "lastModified": 1775135859,
+        "narHash": "sha256-Feg5TeX5cUd+I0uwciytR3Fv4onpCivC54VDOyG1gPU=",
+        "ref": "nella/csa-orfs-eval",
+        "rev": "a41f55eba71a000a8aba26bb1b209659ffeb159a",
+        "revCount": 16822,
         "submodules": true,
         "type": "git",
         "url": "https://github.com/YosysHQ/yosys"
       },
       "original": {
-        "ref": "nella/carry-save-adders",
-        "rev": "0909e462c3efc2ecd9395e1d69485f9ad4027857",
+        "ref": "nella/csa-orfs-eval",
+        "rev": "a41f55eba71a000a8aba26bb1b209659ffeb159a",
         "submodules": true,
         "type": "git",
         "url": "https://github.com/YosysHQ/yosys"
--- a/flake.lock
+++ b/flake.lock
@@ -54,16 +54,16 @@
     },
     "nixpkgs_2": {
       "locked": {
-        "lastModified": 1724316499,
-        "narHash": "sha256-Qb9MhKBUTCfWg/wqqaxt89Xfi6qTD3XpTzQ9eXi3JmE=",
+        "lastModified": 1771419570,
+        "narHash": "sha256-bxAlQgre3pcQcaRUm/8A0v/X8d2nhfraWSFqVmMcBcU=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "797f7dc49e0bc7fab4b57c021cdf68f595e47841",
+        "rev": "6d41bc27aaf7b6a3ba6b169db3bd5d6159cfaa47",
         "type": "github"
       },
       "original": {
         "owner": "nixos",
-        "ref": "nixos-24.05",
+        "ref": "nixos-25.11",
         "repo": "nixpkgs",
         "type": "github"
       }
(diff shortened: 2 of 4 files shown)


maliberty:
Please summarize the benefit you see

widlarizer:
CSA_TREE isn't getting run in CI for some reason. Probably the yosys is not the one we set in the submodule, I can't rule it out since it reports `Yosys 0.63+post (git sha1 UNKNOWN-dirty`. I'm seeing in the logs that a docker image `foss-fpga-tools-ext-openroad` gets used. Does that mean we no longer can just bump the submodule to whatever experimental yosys version we'd like to try out then?

maliberty:
Submodules point at a specific commit which I don't see updated here.  You just changed repo path and are still getting the same commit

Goubermouche:
> Submodules point at a specific commit which I don't see updated here. You just changed repo path and are still getting the same commit

@maliberty Apologies, but I'm not sure what you mean by that - as far as I can see, the submodule points to the relevant commit ([Enable CSA for ORFS eval.](https://github.com/YosysHQ/yosys/commit/a41f55eba71a000a8aba26bb1b209659ffeb159a)).

image

maliberty:
@vvbandeira suggestions?

oharboe:
If you are particularly adventurous, you can try #4094 with upcoming changes to https://github.com/The-OpenROAD-Project/bazel-orfs to run the full ORFS test suite against a patched Yosys.

Here is what that looks like, step by step:

**Step 1: Install bazelisk**

[Bazelisk](https://github.com/bazelbuild/bazelisk) is a small launcher that auto-downloads the correct version of Bazel for the project. Think of it as a wrapper — you run `bazelisk` instead of `bazel` and it handles the rest.

```bash
# Download the binary
curl -L -o /usr/local/bin/bazelisk \
  https://github.com/bazelbuild/bazelisk/releases/latest/download/bazelisk-linux-amd64
chmod +x /usr/local/bin/bazelisk
```

**Step 2: Checkout #4094**

PR #4094 adds Bazel build definitions for 46 designs across 6 PDK platforms. Without it, the ORFS repo has no Bazel support.

```bash
git fetch origin pull/4094/head:bazel-orfs-beta
git checkout bazel-orfs-beta
```

**Step 3: Bump Yosys with patches in `MODULE.bazel`**

`MODULE.bazel` is a file at the root of the repo that pins the exact version of every external dependency (Yosys, OpenROAD, etc.). Think of it as a lockfile — it guarantees everyone uses the same tool versions. No Docker image can silently override it.

Open `MODULE.bazel`, find the Yosys entry, and change it to point to your patched Yosys (e.g. the one with the CSA tree pass from YosysHQ/yosys#5753). Save the file.

**Step 4: Run the full test suite**

```bash
bazelisk test ... --keep_going
```

For those used to Make, here is what each part means:

- `bazelisk` — launches Bazel (auto-downloading it if needed).
- `test` — builds all targets and runs their tests (like `make test`, but it also builds any missing dependencies first).
- `...` — Bazel syntax for "everything, recursively." You type the three dots literally. This is equivalent to running make in every subdirectory at once.
- `--keep_going` — do not stop at the first failure (like `make -k`). You get the full picture of what passes and what breaks.

Behind the scenes, Bazel will:

1. Download the exact Yosys version you pinned in Step 3.
2. Run synthesis, floorplan, placement, CTS, routing, and final optimization for every design/platform combination.
3. Compare QoR metrics against baselines.
4. Report pass/fail per design.

Bazel parallelizes across all available cores automatically. On a 48-core machine, the full suite takes roughly 3–4 hours.

**Step 5: Compare before/after QoR numbers**

Each ORFS run produces a `metadata.json` file (in `reports/`) containing every metric from the flow — area, timing, wirelength, DRC errors, etc. The keys look like `finish__timing__setup__ws`, `detailedroute__route__wirelength`, and so on.

To get a before/after comparison, run the flow twice — once with the original Yosys and once with your patched version — and save the `metadata.json` from each run. Then use this Python script to plot the differences:

```python
#!/usr/bin/env python3
"""Compare before/after QoR metrics from two ORFS metadata.json files."""

import json
import sys
import matplotlib.pyplot as plt
import numpy as np

def load_metrics(path):
    with open(path) as f:
        data = json.load(f)
    # Keep only numeric metrics (skip strings, lists, etc.)
    return {k: v for k, v in data.items() if isinstance(v, (int, float))}

def compare(before_path, after_path, filter_prefix=None):
    before = load_metrics(before_path)
    after = load_metrics(after_path)

    # Find common keys
    common = sorted(set(before) & set(after))
    if filter_prefix:
        common = [k for k in common if k.startswith(filter_prefix)]

    labels = []
    pct_changes = []

    for key in common:
        b, a = before[key], after[key]
        if b == 0 and a == 0:
            continue
        if b == 0:
            pct = float('inf')
        else:
            pct = ((a - b) / abs(b)) * 100
        labels.append(key)
        pct_changes.append(pct)

    return labels, pct_changes

def plot(labels, pct_changes, title="QoR: Before vs After"):
    # Color: green for improvement-direction, red for regression-direction
    # For timing (ws, tns): higher (less negative) is better → positive % = green
    # For area/wirelength: lower is better → negative % = green
    colors = []
    for label, pct in zip(labels, pct_changes):
        is_lower_better = any(x in label for x in ["area", "wirelength", "drc", "violations", "diodes"])
        if is_lower_better:
            colors.append("green" if pct <= 0 else "red")
        else:
            colors.append("green" if pct >= 0 else "red")

    fig, ax = plt.subplots(figsize=(12, max(4, len(labels) * 0.35)))
    y_pos = np.arange(len(labels))
    ax.barh(y_pos, pct_changes, color=colors, edgecolor="black", linewidth=0.5)
    ax.set_yticks(y_pos)
    ax.set_yticklabels(labels, fontsize=8)
    ax.set_xlabel("% change (after vs before)")
    ax.set_title(title)
    ax.axvline(x=0, color="black", linewidth=0.8)
    plt.tight_layout()
    plt.savefig("qor_comparison.png", dpi=150)
    print("Saved qor_comparison.png")
    plt.show()

if __name__ == "__main__":
    if len(sys.argv) < 3:
        print("Usage: python compare_qor.py <before.json> <after.json> [metric_prefix]")
        print("  metric_prefix: optional filter, e.g. 'finish__' or 'detailedroute__'")
        sys.exit(1)

    prefix = sys.argv[3] if len(sys.argv) > 3 else None
    labels, pct_changes = compare(sys.argv[1], sys.argv[2], prefix)
    plot(labels, pct_changes)
```

Usage example — compare just the final-stage metrics for gcd on nangate45:

```bash
# Run once with original Yosys, save the metadata
cp flow/reports/nangate45/gcd/base/metadata.json before.json

# (bump Yosys in MODULE.bazel, re-run)
cp flow/reports/nangate45/gcd/base/metadata.json after.json

# Plot all metrics
python3 compare_qor.py before.json after.json

# Or filter to just finish-stage metrics
python3 compare_qor.py before.json after.json finish__
```

This produces a horizontal bar chart where green bars are improvements and red bars are regressions, sized by percentage change.

**Why this helps with the problem in this PR:**

The confusion here was about which Yosys binary CI was actually running — the submodule commit with the CSA patch, or the pre-built one from the Docker image. With bazel-orfs, the Yosys version is explicitly written in `MODULE.bazel`. There is no ambiguity, no Docker override, and you can test all 46 designs with a single command.

I'll probably add a `bazelisk run //:plot-merge-base` or something that does this.

widlarizer:
Ran ORFS locally, comparing with and without -csa in SYNTH_ARGS. Ibex, jpeg and gcd didn't have csa_tree kick in, so no difference there. `csa_tree` did optimize something inside ethmac but it was only triggering on a few of the `$alu` cells that can be found in the design, and the results were overall a wash. Carry save adders are a design-sensitive optimization. Within multipliers we already have wallace trees built by the existing `booth` command, which is integrated in the wrapped operator flow. We'll probably look for specific designs to show off `csa_tree` (a FIR filter? systolic array?) rather than just synthesizing ORFS test designs

PR #4108, opened 2026-04-04
fix: separate EQUIVALENCE_CHECK from RUN_EQY to avoid requiring eqy

oharboe:
EQUIVALENCE_CHECK now only controls writing equivalence check files. The new RUN_EQY variable (default 0) gates actually executing the eqy tool, so builds don't fail when eqy is not installed.

### DIFF
--- a/flow/scripts/cts.tcl
+++ b/flow/scripts/cts.tcl
@@ -68,7 +68,7 @@ if { !$::env(SKIP_CTS_REPAIR_TIMING) } {
 
   repair_timing_helper
 
-  if { $::env(EQUIVALENCE_CHECK) } {
+  if { $::env(EQUIVALENCE_CHECK) && $::env(RUN_EQY) } {
     run_equivalence_test
   }
   if { $::env(LEC_CHECK) } {
--- a/docs/user/FlowVariables.md
+++ b/docs/user/FlowVariables.md
@@ -253,7 +253,7 @@ configuration file.
 | <a name="RTLMP_RPT_DIR"></a>RTLMP_RPT_DIR| Path to the directory where reports are saved.| |
 | <a name="RTLMP_WIRELENGTH_WT"></a>RTLMP_WIRELENGTH_WT| Weight for half-perimiter wirelength.| 100.0|
 | <a name="RULES_JSON"></a>RULES_JSON| json files with the metrics baseline regression rules. In the ORFS Makefile, this defaults to $DESIGN_DIR/rules-base.json, but ORFS does not mandate the users source directory layout and this can be placed elsewhere when the user sets up an ORFS config.mk or from bazel-orfs.| |
-| <a name="RUN_EQY"></a>RUN_EQY| Actually run the eqy equivalence checking tool. Requires EQUIVALENCE_CHECK to be enabled and eqy to be installed.| 0|
+| <a name="RUN_EQY"></a>RUN_EQY| Actually run the eqy equivalence checking tool. Requires EQUIVALENCE_CHECK to be enabled and eqy to be installed. Defaults to 1 when eqy is found in PATH.| |
 | <a name="RUN_LOG_NAME_STEM"></a>RUN_LOG_NAME_STEM| Stem of the log file name, the log file will be named `$(LOG_DIR)/$(RUN_LOG_NAME_STEM).log`.| run|
 | <a name="RUN_SCRIPT"></a>RUN_SCRIPT| Path to script to run from `make run`, python or tcl script detected by .py or .tcl extension.| |
 | <a name="SC_LEF"></a>SC_LEF| Path to technology standard cell LEF file.| |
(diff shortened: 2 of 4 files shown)


oharboe:
@maliberty Interesting... Is master broken? Does eqy have side-effects???

flow/scripts/cts.tcl
```diff
@@ -68,7 +68,7 @@ if { !$::env(SKIP_CTS_REPAIR_TIMING) } {
 
   repair_timing_helper
 
-  if { $::env(EQUIVALENCE_CHECK) } {
+  if { $::env(EQUIVALENCE_CHECK) && $::env(RUN_EQY) } {
```
maliberty, line 71: I don't follow why we need a second variable to control this?

maliberty:
Failures on master were resolved over the weekend, except the on-going yosys non-determinism.  Merge master and drop the OR update.

flow/scripts/cts.tcl
oharboe: See documentation:

- Should we do an EQY check at all? Do we have the tool?
- Should we do an EQY check *on this module*?

This improves the user experience and helps bazel because the eqy isn't part of normal flow and I don't currently install that tool. We can later, but I think that whether to run eqy and whether to run it *on this module* should be two concerns.

maliberty: I don't see how those are separate.  What does it mean to "run eqy" but not on this module?

oharboe: If you don't have eqy installed, then CTS fails today on modules which have EQUIVALENCE_CHECK =1 (or ?=1).

With this change, if you don't have eqy installed, then RUN_EQY defaults to 0, and cts no longer fails.

oharboe: Designs with `EQUIVALENCE_CHECK=1` (e.g. `swerv_wrapper`) call `exec eqy` during CTS.
If `eqy` isn't installed on the machine, **CTS crashes**. An optional tool
shouldn't break the build.

## The fix: one variable becomes two

| Variable | What it does | Default |
|---|---|---|
| `EQUIVALENCE_CHECK` | Write before/after verilog files (cheap, no tool needed) | `0` (unchanged) |
| `RUN_EQY` | Actually execute the `eqy` binary | `1` if `eqy` is in `$PATH`, else `0` |

If you have `eqy` installed: **nothing changes**. Auto-detected, auto-enabled.

If you don't have `eqy`: builds stop crashing.

## The 4 files that matter

**The core logic change** — [`cts.tcl` line 71](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/db4318541ab992335d1254325a0e7fca47a205cb/flow/scripts/cts.tcl#L71):

```tcl
if { $::env(EQUIVALENCE_CHECK) && $::env(RUN_EQY) } {
    run_equivalence_test
}
```

Previously this only checked `EQUIVALENCE_CHECK`. Now it requires both.
The verilog *write* at [line 62](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/db4318541ab992335d1254325a0e7fca47a205cb/flow/scripts/cts.tcl#L62) still only needs `EQUIVALENCE_CHECK` (no tool required).

**Auto-detection** — [`variables.mk` lines 135-137](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/db4318541ab992335d1254325a0e7fca47a205cb/flow/scripts/variables.mk#L135-L137):

```make
ifneq ($(shell command -v eqy),)
  export RUN_EQY ?= 1
endif
```

Same pattern already used for `stdbuf` a few lines above.

**Variable definition** — [`variables.yaml` RUN_EQY block](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/db4318541ab992335d1254325a0e7fca47a205cb/flow/scripts/variables.yaml#L94-L100)

**Docs** — [`FlowVariables.md` RUN_EQY entry](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/db4318541ab992335d1254325a0e7fca47a205cb/docs/user/FlowVariables.md#L256)

## "Why two variables?"

The question from review: *"What does it mean to run eqy but not on this module?"*

It's the reverse that matters: **module wants equivalence checking (`EQUIVALENCE_CHECK=1`)
but machine doesn't have `eqy` installed**. Without the split, that = build failure.
With the split, the verilog files still get written (useful for running `eqy` later
or elsewhere), the build just skips the tool it can't run.

## To merge

Per your review comment: merge master and drop the OpenROAD submodule bump.
The actual change is just the variable split across the 4 files above.

maliberty:
I think this is more easily solved by just removing eqy entirely.  We have KF with LEC_CHECK so let's retire EQUIVALENCE_CHECK.

oharboe:
> I think this is more easily solved by just removing eqy entirely. We have KF with LEC_CHECK so let's retire EQUIVALENCE_CHECK.

Begone https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/4117 !

PR #4109, opened 2026-04-04
fix: use stat -hierarchy for hierarchical synth to report area

oharboe:
When SYNTH_HIERARCHICAL=1, the yosys stat command reports zero area for the top module because all cells are in submodules. Adding -hierarchy makes stat include submodule area, so
synth__design__instance__area__stdcell is no longer N/A.

### DIFF
--- a/flow/scripts/synth.tcl
+++ b/flow/scripts/synth.tcl
@@ -226,7 +226,11 @@ if { $::env(SYNTH_INSBUF) } {
 # Reports
 tee -o $::env(REPORTS_DIR)/synth_check.txt check
 
-tee -o $::env(REPORTS_DIR)/synth_stat.txt stat {*}$lib_args
+if { $::env(SYNTH_HIERARCHICAL) } {
+  tee -o $::env(REPORTS_DIR)/synth_stat.txt stat -hierarchy {*}$lib_args
+} else {
+  tee -o $::env(REPORTS_DIR)/synth_stat.txt stat {*}$lib_args
+}
 
 # check the design is composed exclusively of target cells, and
 # check for other problems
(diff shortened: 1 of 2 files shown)


oharboe:
@maliberty  I think there is a problem on pr-merge:

    08:18:14  [ERROR PSM-0069] Check connectivity failed on VDD.

flow/scripts/synth.tcl
```diff
@@ -226,7 +226,11 @@ if { $::env(SYNTH_INSBUF) } {
 # Reports
 tee -o $::env(REPORTS_DIR)/synth_check.txt check
 
-tee -o $::env(REPORTS_DIR)/synth_stat.txt stat {*}$lib_args
+if { $::env(SYNTH_HIERARCHICAL) } {
+  tee -o $::env(REPORTS_DIR)/synth_stat.txt stat -hierarchy {*}$lib_args
```
maliberty: Is there any harm to always using -hierarchy  ?

oharboe: I don't know who the consumers are... Why would you not want the hierarchy, if there is one?

maliberty: util/genMetrics.py is the main one.  Have you verified it works?

oharboe: I don't know how to.

oharboe: Is there a test?

oharboe: > Is there any harm to always using -hierarchy ?

Done.

oharboe:
@maliberty Looks like some pr-merge instability that we've been seeing.

oharboe:
@maliberty Ping?

maliberty:
This does break genMetrics so it will need some attention.

oharboe:
> This does break genMetrics so it will need some attention.

Makes sense. Should I file a feature request or is it adequately tracked here?

This is part of the grander plan of getting rid of docker + supporting config.mk DSL (i.e. bazel build for all designs in ORFS).

Not enormously urgent, but nice to have chalked off. Will bubble its way to the top of the list eventually...

maliberty:
this is sufficient

maliberty (approved):

PR #4118, opened 2026-04-07
jpeg_rle: Add missing reset to ddstrb pipeline register to prevent spurious go after reset

ashnaaseth2325-oss:
## **SUMMARY**

This PR adds a missing asynchronous reset to the `ddstrb` pipeline register in `jpeg_rle.v`, preventing invalid signal propagation after reset. The change affects the `jpeg_rle` module and ensures stable interaction with `jpeg_rle1` by eliminating unintended `go` pulses.

---

## **FIX**

```verilog
// Before
always @(posedge clk)
  ddstrb <= #1 dstrb;

// After
always @(posedge clk or negedge rst)
  if (!rst)
    ddstrb <= #1 1'b0;
  else
    ddstrb <= #1 dstrb;
```

---

## **VERIFICATION**

Simulate with `rst` deasserted while `dstrb` is high or unknown; previously this caused an immediate transition to `ac` state with invalid outputs. After the fix, `ddstrb` resets to 0 and no spurious `go` pulse occurs. The state machine remains in `dc` until valid input arrives, preserving correct JPEG block encoding.

### DIFF
--- a/flow/designs/src/jpeg/jpeg_rle.v
+++ b/flow/designs/src/jpeg/jpeg_rle.v
@@ -98,8 +98,11 @@ module jpeg_rle(clk, rst, ena, dstrb, din, size, rlen, amp, douten, bstart);
 	//
 
 	reg ddstrb;
-	always @(posedge clk)
-	  ddstrb <= #1 dstrb;
+	always @(posedge clk or negedge rst)
+	  if (!rst)
+	    ddstrb <= #1 1'b0;
+	  else
+	    ddstrb <= #1 dstrb;
 
 	// generate run-length encoded signals
 	jpeg_rle1 rle(
--- a/flow/designs/sky130hd/jpeg/rules-base.json
+++ b/flow/designs/sky130hd/jpeg/rules-base.json
@@ -32,7 +32,7 @@
         "compare": ">="
     },
     "cts__timing__setup__tns": {
-        "value": -155.0,
+        "value": -170.0,
         "compare": ">="
     },
     "cts__timing__hold__ws": {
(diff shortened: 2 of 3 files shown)


maliberty:
This change leads to a small QOR change.  Please run `update_ok` for this design and commit the updated rules file.  That should get the CI passing.

ashnaaseth2325-oss:
Hello @maliberty,
Thanks! This change affects QOR as expected.
I’m currently unable to run the full OpenROAD flow locally due to tool setup issues, and I also don’t have access to Jenkins artifacts.
Could you please run `update_ok` and commit the updated rules?
Happy to rebase if needed.

ashnaaseth2325-oss:
Hi @maliberty, I tried running update_ok again but couldn’t complete the flow locally. Could you help update the QOR reference or suggest an alternative?

flow/designs/src/jpeg/jpeg_rle.v
```diff
@@ -98,8 +98,11 @@ module jpeg_rle(clk, rst, ena, dstrb, din, size, rlen, amp, douten, bstart);
 	//
 
 	reg ddstrb;
-	always @(posedge clk)
-	  ddstrb <= #1 dstrb;
+	always @(posedge clk or negedge rst)
```
maliberty, line 101: Why both clock edges?

ashnaaseth2325-oss:
Hello @maliberty ,
Thanks for pointing this out. This is not using both clock edges, it still runs on posedge clk. The negedge rst is only for the async reset, so ddstrb is set properly on reset. The earlier version had no reset, which caused invalid values after reset.

maliberty (approved):

PR #4131, opened 2026-04-10
Support SLANG_PLUGIN_PATH env var for custom slang plugin location

oharboe:
When SYNTH_HDL_FRONTEND=slang, the plugin load is hardcoded as `plugin -i slang`.  This adds a check for SLANG_PLUGIN_PATH, falling back to the current behavior when unset.

This enables Bazel-built slang plugins to be loaded from a custom path, which is needed for downstream consumers that build yosys-slang from source.


maliberty:
Why not put this in variables.yaml  with a default of "slang"?

oharboe:
> Why not put this in variables.yaml with a default of "slang"?

Fixed.

maliberty (approved):

PR #4135, opened 2026-04-11
Lib2v

oharboe:


oharboe:
@maliberty FYI this should be polished up and the behvaioral logic manually authored empty.v and dff.v should be deleted.

oharboe:
Just too many things going on, out of sight, out of mind for now.

PR #4144, opened 2026-04-13
update naja with bundle support

nanocoh:
update naja with bundle support following issue  https://github.com/keplertech/kepler-formal/issues/120


maliberty (approved):
@jeffng-or FYI

PR #4157, opened 2026-04-17
fix: remove non-root overrides from MODULE.bazel

oharboe:
When ORFS is fetched as a non-root dependency, git_override and orfs.default() directives are ignored/illegal.  Remove them so the module can be used as a dependency without errors.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -7,13 +7,6 @@ module(
 
 bazel_dep(name = "bazel-orfs")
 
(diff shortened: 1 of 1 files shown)


maliberty:
what replaces `To bump version, run: bazelisk run @bazel-orfs//:bump` ?

oharboe:
ugh. Leave be for now. There's some boot strapping problem here I need to clean up with ORFS being a bazel project and also used to get to scripts+pdk.

PR #4158, opened 2026-04-17
platforms: gf180: endcap_cpp is deprecated

dnltz:
[WARNING TAP-0014] endcap_cpp option is deprecated.
    
Remove this deprecated option, since it's not used in
OpenROAD's tapcell module anymore.
    
Also clean up the uart-blocks design for gf180.


dnltz:
Key is generating a warning and has no affect on tapcell generation: https://github.com/The-OpenROAD-Project/OpenROAD/blob/1afad943d39e3f3a82da2b2c9716857f5ffc9490/src/tap/src/tapcell.tcl#L48

maliberty (approved):

PR #4159, opened 2026-04-17
make: preserve tool *_EXE env vars across UNSET_AND_MAKE

oharboe:
Commit df79ba1d9 ("make: immediately expand deferred vars") changed the tool-path setup from

    export OPENROAD_EXE ?= $(abspath ...)

to

    OPENROAD_EXE ?= $(abspath ...)
    export OPENROAD_EXE := $(OPENROAD_EXE)

The := reassign gives the variable origin "file" even when the value was supplied through the environment. UNSET_VARIABLES_NAMES collects every file-origin variable and $(UNSET_AND_MAKE) unsets them in the bash wrapper before invoking the submake. That strips a caller- supplied `OPENROAD_EXE / OPENSTA_EXE / YOSYS_EXE` from the submake's environment, and variables.mk re-derives them via the $(FLOW_HOME)/../tools/install/... fallback — a path that does not exist in hermetic builds (e.g. bazel-orfs).

Extend the unset exclusion list to cover OPENROAD%, OPENSTA%, YOSYS%, mirroring the KLAYOUT% entry already there for the same reason.

This restores the pre- df79ba1d9 behavior for callers that pass tool paths through the environment, without reverting the immediate- expansion fix.

### DIFF
--- a/flow/scripts/variables.mk
+++ b/flow/scripts/variables.mk
@@ -188,7 +188,7 @@ export RESULTS_V = $(notdir $(sort $(wildcard $(RESULTS_DIR)/*.v)))
 export GDS_MERGED_FILE = $(RESULTS_DIR)/6_1_merged.$(STREAM_SYSTEM_EXT)
 
 define get_variables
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty can we put this one to bed? Arcane make snags, but small.

maliberty (approved):

PR #4167, opened 2026-04-20
Add Content to IHP SG13G2 Readme

dnltz:
Fill the Readme.md for the ihp-sg13g2 platform. It now contains some basic information about the PDK itself, standard cells, SRAM macros and how to add a padframe.

Also list IHP as supported PDK in the documentation.

### DIFF
--- a/flow/platforms/ihp-sg13g2/Readme.md
+++ b/flow/platforms/ihp-sg13g2/Readme.md
@@ -1,3 +1,142 @@
-# IHP SG13G2 official port for OpenROAD-flow-scripts
+# IHP SG13G2 Platform for OpenROAD-flow-scripts
 
-to be documented
+## Overview
+
+The IHP SG13G2 is a **130 nm BiCMOS open PDK** developed by
+[IHP – Innovations for High Performance Microelectronics](https://www.ihp-microelectronics.com/)
+and released under the Apache 2.0 licence.  It is one of the few fully open
+silicon PDKs that includes not only standard CMOS logic cells but also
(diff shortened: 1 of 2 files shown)


maliberty (approved):

flow/platforms/ihp-sg13g2/Readme.md
```diff
@@ -1,3 +1,142 @@
...
+`cells_clkgate.v`.
+
+Fill / decap cells: `sg13g2_fill_1`, `sg13g2_fill_2`, `sg13g2_decap_4`,
+`sg13g2_decap_8`.
+
+---
+
+## SRAM Macros
+
+SRAM macros are available in the following configurations and are ready to use
```
oharboe, line 59: is there a behavioral model for these .lib/.lef macros?

dnltz: There is a behavioral model in the Verilog file, yes. I only used the functional model because verilator can't simulate the behavioral model :(

oharboe: I've always thought about digital design in terms of:

- Behavioral Verilog with **infinite propagation speed** (zero-delay, unit-time) semantics.
- Synthesis to a gate-level netlist.
- Timing coming from the **`.lib` (Liberty) file**, which characterizes each standard cell's delay as a function of input slew, output load, voltage, temperature, etc.
- Static timing analysis (STA) on top of that.

I've never written or thought in terms of `#delay` or `wait` statements.

PR #4168, opened 2026-04-20
adapt macro placer script to use new command to set base halo and update OR

AcKoucher:
The OR update in needed to get the new command.

### DIFF
--- a/flow/scripts/macro_place_util.tcl
+++ b/flow/scripts/macro_place_util.tcl
@@ -38,14 +38,14 @@ if { [find_macros] != "" } {
     log_cmd source $::env(MACRO_PLACEMENT_TCL)
   }
 
+  log_cmd set_macro_base_halo $halo_x $halo_y
+
   set additional_rtlmp_args ""
   append_env_var additional_rtlmp_args RTLMP_MAX_LEVEL -max_num_level 1
   append_env_var additional_rtlmp_args RTLMP_MAX_INST -max_num_inst 1
   append_env_var additional_rtlmp_args RTLMP_MIN_INST -min_num_inst 1
   append_env_var additional_rtlmp_args RTLMP_MAX_MACRO -max_num_macro 1
   append_env_var additional_rtlmp_args RTLMP_MIN_MACRO -min_num_macro 1
-  append additional_rtlmp_args " -halo_width $halo_x"
-  append additional_rtlmp_args " -halo_height $halo_y"
   append_env_var additional_rtlmp_args RTLMP_MIN_AR -min_ar 1
   append_env_var additional_rtlmp_args RTLMP_AREA_WT -area_weight 1
   append_env_var additional_rtlmp_args RTLMP_WIRELENGTH_WT -wirelength_weight 1


flow/scripts/macro_place_util.tcl
```diff
@@ -38,14 +38,14 @@ if { [find_macros] != "" } {
     log_cmd source $::env(MACRO_PLACEMENT_TCL)
   }
 
+  log_cmd set_macro_base_halo $halo_x $halo_y
```
oharboe, line 41: Global state?

AcKoucher: Not exactly global. The halo lives in MPL (it will also soon be responsible for blockage generation - which depends on the halo, see https://github.com/The-OpenROAD-Project/OpenROAD/pull/10204).

AcKoucher:
@joaomai FYI.

flow/scripts/macro_place_util.tcl
oharboe: Is the state that is modified here in the odb?

AcKoucher: No. MPL-scoped.

oharboe: In what sense is this not global state? Why is it stored in global state instead of being passed in as arguments? This makes it harder to follow the .tcl and build a mental model of what the code is doing.

oharboe: This is a [surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)  inconsistency with the rest of the OpenROAD tcl/python commands and interface.

AcKoucher: > In what sense is this not global state?

If by global state you mean "it lives as long as OR lives" then it is global state. My point was that ownership-wise, it belongs to MPL.

> Why is it stored in global state instead of being passed in as arguments?

Currently, the soft blockages generated to block the channels between macros lives within MPL. This is problematic as other flows handle these blockages differently than ORFS. I'm working to move the blockage generation to its own pass. _The blockage generation requires the halo data, which should be available in OR after MPL is done._

> This makes it harder to follow the .tcl and build a mental model of what the code is doing.

We'll have another command `block_macro_channels` after `rtlmp_macro_placer` which makes us have:
1. A section for macro configs;
2. Macro placement;
3. Blockage generation.

E.g.,
```
set_macro_base_halo ...
set_macro_halo ... // some specific halo the user wants

rtlmp_macro_placer

block_macro_channels
```

oharboe: What *is* done elsewhere with great success and is unsurprising is to store global state in the .odb, because then it can be set once and picked up across floorplan/place/global route. I don't recall the details, it has been a while, routing layer adjustment was moved from ad-hoc global openroad state to .odb state.

So minimum surprise, most consistent is to either:

1. store in .odb
2. carry as state in a single .tcl and pass in as arguments

AcKoucher: I agree - you have a point.

I think that the main problem with this whole thing is that we're conflating halos and channels concepts.
I'll close this PR and write up an initial idea for us to discuss a better way on handling all this.

Thanks for the heads up.

PR #4174, opened 2026-04-22
synth: expand SYNTH_RETIME_MODULES list into separate select args

oharboe:
Tcl's `$var` substitution passes the whole value as a single argument, so `select $::env(SYNTH_RETIME_MODULES)` with a space-separated module list becomes `select "a b c"` — one bogus pattern that matches no module, silently disabling retime. Use `{*}` list-expansion so each pattern in the env var arrives as its own `select` argument.


maliberty (approved):

oharboe:
@vvbandeira @maliberty CI outage, pr-merge did not retrigger yet. Will it retrigger given the CI outage affected pr-head too?

image

maliberty:
It doesn't re-run pr-head as that _should_ give the same result.  It isn't merge blocking in any case.

PR #4177, opened 2026-04-22
synth: apply SYNTH_BLACKBOXES to loaded RTLIL after checkpoint read

oharboe:
The preamble already registers SYNTH_BLACKBOXES at the HDL-read stage, but when synth.tcl is entered via a pre-loaded RTLIL checkpoint (see SYNTH_CHECKPOINT) the preamble is not involved, so the blackbox list needs a second application point on the loaded netlist.

Run the same list through `blackbox` between read_checkpoint and hierarchy -check. Wrap each in `catch` so stale names from a static list don't fail the flow — matches the existing "non-existent modules are ignored silently" documentation for the variable.

The variable's doc entry already notes Bazel parallel synth; extended to call out the post-checkpoint apply point for completeness.


maliberty (approved):
I think this was subsumed in another PR but will merge it just in case.

PR #4191, opened 2026-04-26
Catch misspelled / wrong-stage env var references in flow scripts

oharboe:
Add a typo guard in util.tcl: env_var_equals,
env_var_exists_and_non_empty, env_var_or_empty, append_env_var, and source_env_var_if_exists now error if the variable name is not declared in variables.yaml for the current stage. erase_non_stage_variables populates the per-stage declared set; utility scripts that don't run it (e.g. open.tcl) fall back to the full set.

Add declared_variables.py to emit the per-stage declared set as a sibling of non_stage_variables.py.

Declare 6 helper-referenced vars that were missing from variables.yaml: DEF_FILE, V_FILE, LIB_MODEL, RTLMP_RPT_FILE, RTLMP_BLOCKAGE_FILE, SDC_FILE_CLOCK_PERIOD.

Widen FOOTPRINT and FOOTPRINT_TCL to stages [floorplan, place]. Fixes a latent bug: io_placement.tcl and global_place.tcl read them in the place stage but they were declared floorplan-only and so were always silently empty after erase_non_stage_variables.

Regenerate variables.json and docs/user/FlowVariables.md via yaml_to_json.py and generate-variables-docs.py.


oharboe:
leave be for now

PR #4211, opened 2026-05-06
synth: blackbox SYNTH_BLACKBOXES modules before hierarchy check

oharboe:
## Summary

Adds an optional `SYNTH_BLACKBOXES` env var that, when set, blackboxes the listed modules before `hierarchy -check`.

Intended for external orchestrators that drive parallel partition synthesis: every partition loads the same canonical RTLIL checkpoint, then blackboxes the modules that belong to other partitions so it only synthesises its own subhierarchy.

Names not present in the loaded design are skipped silently (`catch {blackbox $m}`), so the same list can be reused across all partitions without per-partition filtering. No behaviour change when the env var is unset.

This block is currently carried as a vendored modification of `synth.tcl` in `bazel-orfs`. Upstreaming it lets `bazel-orfs` drop the vendored copy and source `flow/scripts/synth.tcl` directly from ORFS.

## Test plan

- [x] No behaviour change when `SYNTH_BLACKBOXES` is unset (the `if` block is skipped)
- [ ] CI green

### DIFF
--- a/flow/scripts/synth.tcl
+++ b/flow/scripts/synth.tcl
@@ -38,6 +38,18 @@ if { [env_var_exists_and_non_empty SYNTH_CHECKPOINT] } {
   read_checkpoint $::env(RESULTS_DIR)/1_1_yosys_canonicalize.rtlil
 }
 
+# When this synthesis run is one partition of a parallel split (driven by
+# an external orchestrator), `SYNTH_BLACKBOXES` lists modules outside this
+# partition.  Blackboxing them before the hierarchy check lets each
+# partition load the same canonical RTLIL checkpoint while only synthesising
+# its own subhierarchy.  Names not present in the loaded design are skipped
+# silently so the same list can be passed to every partition.
+if { [env_var_exists_and_non_empty SYNTH_BLACKBOXES] } {
+  foreach m $::env(SYNTH_BLACKBOXES) {
+    catch { blackbox $m }
+  }
+}
+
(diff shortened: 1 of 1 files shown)


oharboe:
Closing — after running the numbers on a real partition synth (8-slice ProcessingUnit, 46 partitions): the shared-canonicalize optimization that bazel-orfs is built around saves only ~5% total CPU and 0-2% wall time, because slang's `--blackboxed-module` already skips parsing of blackboxed modules at the source level. We're going to refactor bazel-orfs to run per-partition canonicalize via the standard ORFS flow, which makes this checkpoint-side SYNTH_BLACKBOXES handling unnecessary. Apologies for the noise.

oharboe:
Reopening — measured the cost on a real partition synth (8-slice ProcessingUnit, 46 partitions; data in the previous comment). Per-partition canonicalize from a bare top doesn't actually work without a brittle rename pass: slang elaborates parameterized modules to mangled names like `Foo$Path.To.Inst` and the parent design references those mangled names, so a fresh-top elaboration produces `Foo` and links don't match. The shared canonicalize is doing semantic work, not just bookkeeping.

Also addressed Gemini's consistency feedback by aligning the existing `synth_preamble.tcl` verilog-frontend path: it now uses `catch` (matching the slang and new synth.tcl paths) and blackboxes before `hierarchy -check -top` (matching the new synth.tcl path).

PR #4232, opened 2026-05-15
Netlist hash diagnostic

oharboe:
Reduce wild goose chases with helpful diagnostics

### DIFF
--- a/flow/util/genElapsedTime.py
+++ b/flow/util/genElapsedTime.py
@@ -14,10 +14,11 @@ import sys
 # ==============================================================================
 
 
-# Primary data artifacts first, then the SDC constraint file: yosys
-# emits .v / .rtlil; OpenROAD stages emit .odb; both often emit a
-# .sdc alongside.
-RESULT_EXTS = [".v", ".rtlil", ".odb", ".sdc"]
+# Primary data artifacts first, then derived/exported artifacts and
+# the SDC constraint file: yosys emits .v / .rtlil; OpenROAD stages
+# emit .odb (and often .def / .sdc); routing emits .spef; finish
+# emits .gds.
+RESULT_EXTS = [".v", ".rtlil", ".odb", ".def", ".spef", ".gds", ".sdc"]
 
 
 def get_hashes(f):
--- a/flow/util/genElapsedTime.py
+++ b/flow/util/genElapsedTime.py
@@ -14,24 +14,41 @@ import sys
 # ==============================================================================
 
 
-def get_hash(f):
-    # content hash for the result file alongside .log file is useful to
-    # debug divergent results under what should be identical
-    # builds(such as local and CI builds)
-    for ext in [".odb", ".rtlil", ".v"]:
+# Primary data artifacts first, then the SDC constraint file: yosys
+# emits .v / .rtlil; OpenROAD stages emit .odb; both often emit a
+# .sdc alongside.
+RESULT_EXTS = [".v", ".rtlil", ".odb", ".sdc"]
+
+
+def get_hashes(f):
+    """Return [(ext, sha1), ...] for every result file alongside log
+    `f` whose extension is in RESULT_EXTS.  A yosys stage typically
+    produces both `.v` and `.sdc`; a floorplan/route stage produces
+    `.odb` (and often `.sdc`); the canonicalize stage produces
+    `.rtlil`.  Hashing each separately makes "the netlist changed"
+    distinguishable from "the SDC changed" in the elapsed-time table
+    used to triage divergent local vs CI builds.
+
+    Falls back to a single ("", "N/A") entry when no result file
+    exists so the caller always emits at least one row per stage.
+    """
+    results = []
+    for ext in RESULT_EXTS:
         result_file = pathlib.Path(
             str(f).replace("logs/", "results/").replace(".log", ext)
         )
         if result_file.exists():
             hasher = hashlib.sha1()
-            with open(result_file, "rb") as odb_f:
+            with open(result_file, "rb") as rf:
                 while True:
-                    chunk = odb_f.read(16 * 1024 * 1024)
+                    chunk = rf.read(16 * 1024 * 1024)
                     if not chunk:
                         break
                     hasher.update(chunk)
-            return hasher.hexdigest()
-    return "N/A"
+            results.append((ext, hasher.hexdigest()))
+    if not results:
+        results.append(("", "N/A"))
+    return results
 
 
 def print_log_dir_times(logdir, args):
@@ -87,37 +104,49 @@ def print_log_dir_times(logdir, args):
                     )
                     break
 
-            odb_hash = get_hash(f)
+            hashes = get_hashes(f)
 
             if not found:
                 print("No elapsed time found in", str(f), file=sys.stderr)
                 continue
 
-        # Print the name of the step and the corresponding elapsed time
-        format_str = "%-25s %10s %14s %20s"
+        # Print the name of the step and the corresponding elapsed time.
+        # One row per (stage, result-file-ext); only the first row of a
+        # stage shows elapsed/peak.
+        format_str = "%-25s %-6s %10s %14s %20s"
         if elapsedTime is not None and peak_memory is not None:
             if first and not args.noHeader:
                 print(
                     format_str
-                    % ("Log", "Elapsed/s", "Peak Memory/MB", "sha1sum result [0:20)")
+                    % (
+                        "Log",
+                        "Ext",
+                        "Elapsed/s",
+                        "Peak Memory/MB",
+                        "sha1sum result [0:20)",
+                    )
                 )
                 first = False
-            print(
-                format_str
-                % (
-                    stem,
-                    elapsedTime,
-                    peak_memory,
-                    odb_hash[0:20],
+            stage_first = True
+            for ext, h in hashes:
+                print(
+                    format_str
+                    % (
+                        stem,
+                        ext,
+                        elapsedTime if stage_first else "",
+                        peak_memory if stage_first else "",
+                        h[0:20],
+                    )
                 )
-            )
+                stage_first = False
         if elapsedTime is not None:
             totalElapsed += elapsedTime
         if peak_memory is not None:
             total_max_memory = max(total_max_memory, int(peak_memory))
 
     if totalElapsed != 0 and not args.match:
-        print(format_str % ("Total", totalElapsed, total_max_memory, ""))
+        print(format_str % ("Total", "", totalElapsed, total_max_memory, ""))
 
 
 def scan_logs(args):
--- a/flow/util/genMetrics.py
+++ b/flow/util/genMetrics.py
@@ -265,8 +265,7 @@ def extract_metrics(
     # Netlist hashes: fingerprints of the canonical RTLIL (pre-ABC) and
     # the final post-synthesis Verilog so the rules-base.json check
     # (level=warning) flags when bazel-built vs make-built yosys
-    # disagree for the same RTL. See flow/README.md "Triaging a failing
-    # _test" → "Yosys-environment false positive" for context.
+    # disagree for the same RTL.
     metrics_dict["synth__canonical_netlist__hash"] = file_sha1(
         resultPath + "/1_1_yosys_canonicalize.rtlil"
     )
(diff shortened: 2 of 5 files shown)


flow/util/genElapsedTime.py
```diff
@@ -14,24 +14,41 @@
...
 
-def get_hash(f):
-    # content hash for the result file alongside .log file is useful to
-    # debug divergent results under what should be identical
-    # builds(such as local and CI builds)
-    for ext in [".odb", ".rtlil", ".v"]:
+# Primary data artifacts first, then the SDC constraint file: yosys
+# emits .v / .rtlil; OpenROAD stages emit .odb; both often emit a
+# .sdc alongside.
+RESULT_EXTS = [".v", ".rtlil", ".odb", ".sdc"]
```
maliberty: Missing def, gds, spef

flow/util/genMetrics.py
```diff
@@ -249,6 +262,16 @@ def extract_metrics(
         rptPath + "/synth_stat.txt",
     )
 
+    # Netlist hashes: fingerprints of the canonical RTLIL (pre-ABC) and
+    # the final post-synthesis Verilog so the rules-base.json check
+    # (level=warning) flags when bazel-built vs make-built yosys
+    # disagree for the same RTL. See flow/README.md "Triaging a failing
+    # _test" → "Yosys-environment false positive" for context.
```
maliberty: I see no such section in README.md

maliberty (approved):

maliberty:
Note for the future: we are working towards a goal to remove rules files from git and have the QoR dashboard implement the checking.

oharboe:
> Note for the future: we are working towards a goal to remove rules files from git and have the QoR dashboard implement the checking.

Nice!

How can bazel-orfs get to them?

oharboe:
@povik @maliberty I have more fine grained artifact hash checking in another PR for synthesis, but until yosys-slang is idempotent, there is no rush.

maliberty:
There will be an API endpoint.

oharboe:
> There will be an API endpoint.

Anonymous access?

sounds like a repository rule job

oharboe:
@maliberty Have you considered detached git branches as storage? We have had great success with that as a database for rules*.json stuff.

Super easy, fast, no new infrastructure or API endpoints .

I can ask Claude to rearticulate the current infrastructure in that format as a PR w docs to explain

oharboe:
@vvbandeira @maliberty FYI

# Orphan branch as an append-only database

`git checkout --orphan` produces a branch with no shared history with master. Used as an append-only data store, with `rules-base.json` as the worked example.

## Prior art

The orphan-branch-as-data-store pattern is the same mechanism that powers `gh-pages`, where projects publish built documentation alongside source code on a branch that shares no history with master. The Python benchmarking tool `airspeed-velocity/asv` ships an `asv gh-pages` command that appends benchmark results to an orphan branch over a project's lifetime, and deployed instances at NumPy, SciPy, and AstroPy show years of accumulated history. The widely-used GitHub Action `benchmark-action/github-action-benchmark` does the same thing for continuous benchmarking: results land as JSON on the orphan branch (`gh-pages` by default) and a chart page renders the history.

## Layout

The branch is named, for example, `rules`. It is a true orphan, so `git merge-base master rules` returns nothing. The branch contains one commit per (design, ORFS SHA) measurement, with the rule file stored at `<pdk>/<design>/<orfs-sha-short>.json`. The commit message takes the form `rules: <pdk>/<design> @ <orfs-sha-short> [<openroad-sha-short> yosys-<v>]`, and the file contents themselves carry the full version triple — OpenROAD commit, yosys, ABC, PDK commit, ORFS SHA — lifted from `metadata.json`, so the record is self-describing once extracted. Branch protection forbids force-push and history rewrites.

## Writing

The recorder runs in CI on master after a flow run. It uses `git worktree add` to check out the orphan branch into a sibling directory, leaving the main checkout untouched. It reads `metadata.json`, writes the rule file, and de-duplicates against existing entries on the branch. It then commits as `CI` and pushes `HEAD:rules`.

## Reading

`checkMetadata.py` resolves a rule envelope by looking it up on `rules`. The resolution order is: exact match on the current ORFS SHA, then merge-base against `origin/master`, then a walk back through master ancestry until a recorded SHA is found, and finally a checked-in fallback for runs without the branch fetched. Anywhere outside the flow, the data is reachable with `git show rules:<pdk>/<design>/<sha>.json` from any clone that has fetched the branch.

## Costs

The orphan branch trails master, since it only contains what master CI has actually published; PR baselines therefore resolve at the merge-base SHA, not at the branch's `HEAD`. A default `git clone` does not fetch orphan branches, so checkouts need an explicit `git fetch origin rules:rules`. `make update_rules` becomes a network operation, which means contributors without push rights need a local-only fallback. Schema evolution behaves the way it always does — adding a metric is safe if readers tolerate absent fields in older entries, but renaming is a one-way door. A code change and its corresponding rule change live on different histories, so the correlation between them is reconstructed by SHA in tooling rather than by `git log`. The branch grows append-only and `git log rules` slows over time. Contributors hold a two-branch mental model, and a pre-commit hook is needed to catch attempts to commit rule changes to master. Finally, the branch is only as fresh as the last green master CI run.

## bazel-orfs

bazel-orfs consumes the same data through a Bazel repository rule (or module extension) that fetches the `rules` branch at a pinned commit and exposes the per-design JSON files as a filegroup. Designs that run a metric check depend on `@rules//<pdk>/<design>:<sha>.json` as a Bazel input, which keeps the lookup inside the hermetic sandbox without giving build actions network or implicit `git` access. Pinning the orphan-branch SHA in `MODULE.bazel` (or a lockfile) makes the build reproducible: a given bazel-orfs revision always resolves to the same rule envelope, independent of new measurements landing on `rules`. Bumping the pin is a one-line PR, reviewable like any other dependency bump. Writing is unchanged — the recorder runs outside `bazel build`, in the CI driver that wraps it, using the same worktree-and-push sequence as the Make-based flow. The resolution-order logic (exact ORFS SHA, then merge-base, then ancestry, then fallback) moves into the repository rule, since the Bazel action itself only sees the resolved file.

## Migration

For each existing `rules-base.json`, the master commit that last touched it is found with `git log -1 --format=%H -- <path>`, and the file is replayed onto `rules` with that commit's author date. No measurements are invented.

oharboe:
@povik @maliberty yosys idempotency is a problem worth giving up on...


# Orphan branch as a cache for synthesized `.rtlil`

`yosys-slang`, the SystemVerilog frontend plugin for yosys, is non-idempotent: re-running it on the same SV sources produces a different `.rtlil` because of pointer-iteration order inside the elaborator. Upstream treats this as expected. Downstream, every flow run absorbs the drift, every QoR check becomes noisy, and version skew between developers becomes a debugging trap. The mechanism that worked for `rules-base.json` — an orphan branch as an append-only data store — applies here too. The orphan branch holds canonical `.rtlil` files captured at the post-frontend point, regenerated only on a deliberate `yosys-slang` (or yosys) upgrade. A sidecar YAML enumerates every input that went into producing the file, so the cache key is explicit and the file is rebuilt when any input changes.

## Prior art

Content-addressed caches for slow or non-deterministic build steps are universal. `ccache` and `sccache` cache compiler output keyed on preprocessor input. Bazel's remote action cache does the same thing at action-graph granularity. The Nix store at `/nix/store/<hash>-<name>` is a filesystem-backed instance of the same idea. This proposal uses the same content-addressing pattern, transported over git as an orphan branch in the same way that `gh-pages`, `airspeed-velocity/asv`, and `benchmark-action/github-action-benchmark` use orphan branches to persist data alongside source.

## Layout

The branch is named, for example, `rtlil-cache`. It is a true orphan, sharing no history with master. For each cached entry it holds two files at `<pdk>/<design>/<input-hash>.rtlil` and `<pdk>/<design>/<input-hash>.yaml`, where `<input-hash>` is a content hash over the sidecar YAML. The sidecar enumerates every input that affects the frontend's output: `yosys-slang` commit SHA, yosys version string and commit SHA, ABC commit (if captured post-ABC), the SHA256 of each HDL source file consumed, the SHA256 of any include files and package files reachable from the elaboration root, every option passed to `read_slang` and the surrounding yosys script, and the exact command sequence that was executed up to the capture point. The file's own hash is the cache key, so the YAML is self-describing and the `.rtlil` is reproducible from it. Branch protection forbids force-push and history rewrites.

## Writing

New entries are produced in two contexts. The first is a deliberate `yosys-slang`/yosys upgrade sweep: after bumping the submodule on master, a CI job elaborates every (design, option-set) in the matrix once, generating one `.rtlil` per combination, and pushes all of them to `rtlil-cache` in a single coordinated commit series. The non-determinism is resolved by first-write-wins — whatever the frontend produced on that sweep run becomes the canonical output for that input hash. The second is on-demand: when a flow run hits a cache miss (new design, modified options, modified RTL), CI runs the frontend to fill the entry and pushes it. In both cases the recorder uses `git worktree add` to check out the orphan branch into a sibling directory, writes the pair of files, de-duplicates against existing entries by input hash, and pushes.

## Reading

At flow time the orchestrator computes the input hash from the current YAML, looks up `<pdk>/<design>/<hash>.rtlil` on `rtlil-cache`, and on a hit reads the cached file directly instead of running the frontend. Downstream yosys passes and ABC continue from that canonical RTLIL as they would today. On a miss it falls through to running the frontend, with the recorder publishing the result so subsequent runs hit the cache. Outside the flow the cache is reachable with `git show rtlil-cache:<pdk>/<design>/<hash>.rtlil` from any clone that has fetched the branch.

## Costs

The cache is only sound if the YAML enumerates every input that affects the frontend's output. A missed input means a stale `.rtlil` is reused after that input has changed, producing silent QoR drift. Adding inputs over time is safe — the hash changes, the cache misses, the entry is regenerated — but discovering a missing input retroactively means treating older entries as suspect. `.rtlil` files are large; the orphan branch grows fast, and git-lfs or shallow fetches may be needed. A `yosys-slang` or yosys upgrade is no longer a one-line submodule bump but a deliberate event with a CI sweep attached; until the sweep finishes, the cache is stale for the new version and either the flow falls back to running the frontend (re-introducing the drift the cache exists to suppress) or refuses to build. The capture point is a choice with consequences: cutting post-frontend freezes `yosys-slang` output but lets later yosys passes run live; cutting post-synth or post-ABC freezes more of the pipeline but bakes in ABC's own non-determinism. Contributors hold a two-branch mental model, and a default `git clone` does not fetch the orphan branch. The first-write-wins resolution means whichever machine ran the sweep determines the canonical output for everyone, which is the point but is worth stating out loud.

## bazel-orfs

bazel-orfs consumes the cache through a repository rule that fetches `rtlil-cache` at a pinned commit and exposes the entries as a content-addressed filegroup. The frontend Bazel action depends on the cached `.rtlil` as an input rather than running `yosys-slang`, which removes the non-deterministic step from the hermetic action graph for the common case. The action is still defined so cache misses fall through to it, but on cache hits the action becomes a file-copy and Bazel's own remote cache handles it from there. Pinning the orphan-branch SHA in `MODULE.bazel` makes the build reproducible: a given bazel-orfs revision always resolves to the same `.rtlil`, independent of new entries landing on `rtlil-cache`. A `yosys-slang` or yosys upgrade is a coordinated change — bump the dependency, run the sweep, push to `rtlil-cache`, advance the pin — reviewable as a single PR.

## Migration

For each (design, option-set) currently exercised on master, run the frontend once at the current pinned `yosys-slang` and yosys versions, capture the resulting `.rtlil` and the YAML enumerating the inputs that produced it, and push the pair to `rtlil-cache`. This yields a baseline cache covering today's matrix at today's tool versions. Designs added later or option-sets exercised later fill the cache on demand through the cache-miss path.

povik:
> yosys-slang, the SystemVerilog frontend plugin for yosys, is non-idempotent: re-running it on the same SV sources produces a different .rtlil because of pointer-iteration order inside the elaborator. Upstream treats this as expected. 

This is not true. Neither me (for yosys-slang) or the yosys team treats non-determinism as expected.

oharboe:
> > yosys-slang, the SystemVerilog frontend plugin for yosys, is non-idempotent: re-running it on the same SV sources produces a different .rtlil because of pointer-iteration order inside the elaborator. Upstream treats this as expected.
> 
> This is not true. Neither me (for yosys-slang) or the yosys team treats non-determinism as expected.

Claude says things that are not true.

maliberty:
Pass.  We can discuss at more length whenever we talk next.

PR #4246, opened 2026-05-20
Add readline to the dependency installer for yosys

maliberty:
It was previously getting it from the OpenROAD dependency installer but that was removed.


maliberty:
yosys built so merge, any remaining issues are unrelated.

PR #4248, opened 2026-05-20
etc: fix dependency install order

vvbandeira:


maliberty:
Is this to fix ORFS master?  @sombraSoft was just working on #4249 which seems similar

vvbandeira:
> Is this to fix ORFS master? @sombraSoft was just working on #4249 which seems similar

Yes. Secure Branches Nightly already passed the yosys install.

maliberty:
Thanks for fixing it.

PR #4277, opened 2026-06-04
gt2n: add 2nm BSPDN platform with gcd and aes designs

mguthaus:
Adds the **GT2N** 2 nm GAAFET PDK to ORFS as a new platform, with two reference designs (`gcd`, `aes`) routed end-to-end.

GT2N is an open-source process-aware 2 nm nanosheet PDK developed at Georgia Tech / Synopsys / Samsung (Jang et al., *IEEE APCCAS 2024*, https://github.com/azadnaeemi/GT2N). The collateral here is the `tt 0.7 V 25 C` corner of the W31 LVT 6T library, used unchanged except for one authoring fix on the tap cell (`gt2_6t_tap_w31_lvt`): BPR shapes that were declared as `OBS` are moved into a second `PORT` on each PG pin so the LEF correctly encodes the M1↔BPR bridge (suggested upstream as `azadnaeemi/GT2N#12`).

#### What's in the PR

| | |
|---|---|
| New platform | `flow/platforms/gt2n/` — config, LEF/Lib/GDS, KLayout `.lyp`/`.lyt`, `.layermap`, PDN/RC/tapcell scripts, LICENSE/README. |
| New designs | `flow/designs/gt2n/gcd/`, `flow/designs/gt2n/aes/` — config + SDC. |
| OpenROAD bump | `tools/OpenROAD` → `a44e99193f` (= upstream master HEAD, includes the merged LEF58_BACKSIDE / BSPDN support from The-OpenROAD-Project/OpenROAD#10547). |

#### Platform architecture

BSPDN-style: every standard cell's `vdd`/`vss` pin lives on `BPR` (backside, marked `LEF58_BACKSIDE`). The PDN is therefore **backside-only**:

- `BPR` followpins (pitch 144 nm) sit directly under each row, overlapping the cell's BPR pin → power gets to the device through the cell's intrinsic nano-TSV / WAC structure modelled in the PDK (per Jang et al. §II.A).
- `BM1` mesh perpendicular to BPR, connected via `add_pdn_connect BPR BM1`.
- `BM2` mesh perpendicular to BM1, declared as the grid's top pin, connected via `add_pdn_connect BM1 BM2`.

All three `add_pdn_connect` rules are backside-only by design; the PDN-1200 cross-side guard added in OpenROAD#10547 stays silent. There are no chip-level TSVs or bridge cells in this configuration — that's intentional, matching the paper's architecture.

Signal routing is the frontside stack `M2..M5` (`MIN_ROUTING_LAYER` / `MAX_ROUTING_LAYER`); frontside metals above `M5` and the entire backside stack are excluded from signal routing.

#### Verification

- **gcd**: full `make … 5_2_route` clean (13 min wall, 2.2 GB peak, 0 violations, 0 antenna violations).
- **aes**: full `make … 5_2_route` clean (1 h 39 min wall, 10.7 GB peak, 0 violations).
- **PDN smoke test on this PR's branch state**: `make … 2_4_floorplan_pdn` on `gcd` runs clean (2 s, BPR followpins + BM1/BM2 mesh inserted by PDN without warnings).

#### Notes

- The 21 737-line `.lib` and 4 802-line `.lef` are the upstream GT2N library payload, not authored here. Licensing matches the upstream (BSD-style; see `flow/platforms/gt2n/LICENSE`).
- `setRC.tcl` uses placeholder values for the backside layers and vias — the upstream PDK does not ship extracted backside RC. The values are scaled like the frontside (`V0..V4`) so timing isn't pathologically off.
- The library has 71 unique cells; no analog/IO/macro variants beyond what's used by `gcd` and `aes`.


tools/yosys
maliberty, line 1: Is this required for this PR and mistakenly added?

flow/platforms/gt2n/README.md
```diff
@@ -0,0 +1,56 @@
+# GT2N PDK
```
maliberty, line 1: I assume this is copied from https://github.com/azadnaeemi/GT2N.  Please indicate the commit id for future updating.

tools/yosys
mguthaus: Mistakenly added. I will remove it.

flow/platforms/gt2n/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+<img src="./VirtualFabricationDemo.gif" alt="VirtualFabricationDemo" width="400">
+
+This is the initial release for the GT2N PDK with 71 standard cells. It is based on 2nm GAAFET with BSPDN.
+
+If you use this PDK for publishing your work, we would appreciate citation of the following paper:
+
+D. Jang, P. Kumar, M. N. H. Shazon, S. J. Ram, A. Svizhenko, V. Moroz, A. Ceyhan, N. A. Radhakrishn, and A. Naeemi, "GT2N: An Open-Source 2nm Nanosheet PDK Enabling Multi-Width/VT Benchmarking," in IEEE International Symposium on Circuits and Systems (ISCAS) 2026 [link](https://www.dropbox.com/scl/fi/3to35aip31hfqj76q7g5x/dongwon_iscas26.pdf?rlkey=utm8my3qq36y8mo0hkvto6tnv&st=876o2oh8&dl=0).
```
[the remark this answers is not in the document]
mguthaus: Keeping version from upstream repo.

flow/platforms/gt2n/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+Quick setup guide for Custom Compiler:
+
+Create a folder (e.g. gt2_techlib) and copy GT2N/techlib/nmos_lvt and GT2N/techlib/pmos_lvt inside that folder.
+Inside the folder from where custom_compiler is run, add the following to the lib.defs file:
+gt2_official <path_to_techlib_directory>/gt2_techlib
+
+You can specify any name for the library instead of using gt2_offical. The current standard cell schematics are linked with the devices using this name. In case any other library name is used, the device references might need to be updated in the schematics.
```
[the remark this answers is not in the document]
mguthaus: Keeping version from upsream repo.

flow/platforms/gt2n/gt2.lyt
```diff
@@ -0,0 +1,72 @@
...
...'M2 : 30/0';'M2.LABEL : 30/2';'M2.PIN : 30/251';'M3 : 35/0';'M3.LABEL : 35/2';'M3.PIN : 35/251';'M4 : 40/0';'M4.LABEL : 40/2';'M4.PIN : 40/251';'M5 : 45/0';'M5.LABEL : 45/2';'M5.PIN : 45/251';'M6 : 50/0';'M6.LABEL : 50/2';'M6.PIN : 50/251';'M7 : 55/0';'M7.LABEL : 55/2';'M7.PIN : 55/251';'M8 : 60/0';'M8.LABEL : 60/2';'M8.PIN : 60/251';'M9 : 65/0';'M9.LABEL : 65/2';'M9.PIN : 65/251';'M10 : 70/0';'M10.LABEL : 70/2';'M10.PIN : 70/251';'M11 : 75/0';'M11.LABEL : 75/2';'M11.PIN : 75/251';'M12 : 80/0';'M12.LABEL : 80/2';'M12.PIN : 80/251';'M13 : 85/0';'M13.LABEL : 85/2';'M13.PIN : 85/251';'V0 : 22/0';'V1 : 27/0';'V2 : 32/0';'V3 : 37/0';'V4 : 42/0';'V5 : 47/0';'V6 : 52/0';'V7 : 57/0';'V8 : 62/0';'V9 : 67/0';'V10 : 72/0';'V11 : 77/0';'V12 : 82/0';'V13 : 87/0';'RDL : 90/0';'BPR : 8/0';'BPR.LABEL : 8/2';'BPR.PIN : 8/251';'VBPR : 9/0';'BV0 : 910/0';'BM1 : 920/0';'BM1.LABEL : 921/0';'BV1 : 930/0';'BM2 : 940/0';'BM2.LABEL : 941/0';'BV2 : 950/0';'BM3 : 960/0';'BM3.LABEL : 961/0';'BV3 : 970/0';'BM4 : 980/0';'BM4.LABEL : 981/0';'BV4 : 990/0';'BRDL : 1000/0';'NWELL : 1/0';'ACT : 2/0';'GATE : 3/0';'DUMMY : 4/0';'GCUT : 5/0';'NSEL : 6/0';'PSEL : 7/0';'SDCON : 10/0';'VSD : 11/0';'VG : 12/0')</layer-map>
```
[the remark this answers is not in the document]
mguthaus: Keeping version from upstream repo.

flow/platforms/gt2n/README.md
mguthaus: Added — `README.md` now starts with a **Source** block pinning the upstream commit (`291fae73d6`, 2026-04-29, *Update GT2.ict*) and noting the only local delta (the tap-cell BPR `OBS` → `PORT` fix proposed upstream as `azadnaeemi/GT2N#12`). Folded into the first commit (`dca3e9138`) so the platform addition carries the provenance from day one.

oharboe:
Very interesting!

We're using asap7 today, is this something that is worth looking at now or is it too early?

Can you say something to manage my expectations w.r.t. using this as a drop-in replacement for asap7?

Wiring this up to bazel-orfs (which I use) should be easy enough.

maliberty:
With only 71 std cells it isn't a very rich library.

oharboe:
> With only 71 std cells it isn't a very rich library.

I see. asap7 can't do Kogge Stone because half adder variants aren't available, so buffers are used instead defeating the Kogge Stone optimization.

maliberty:
There is one size of half-adder HAxp5_ASAP7_75t_R, do you mean the lack of size variants?

oharboe:
> There is one size of half-adder HAxp5_ASAP7_75t_R, do you mean the lack of size variants?

Yes: so you get lots of buffers, which eats into the Kogge Stone advantage, which means that all those adder carry chain propagation tricks cluster around the same performance in graphs... Vt mitigates this a bit in a sense.

oharboe:
@mguthaus You've played around with bazel-orfs, I had Claude create  idiomatic support in GT2N https://github.com/azadnaeemi/GT2N/pull/13

mguthaus:
@maliberty @oharboe You are correct that it is a functionally small library.

One dimension that I did not include are the 10 different VTh variants in the library. While there aren't that many sizes, FinFETs are better off using VTh instead of multiple discrete fins. This would increase the commit by ~1MB+ to include those files, so I opted to just do the baseline. Should I add this?

oharboe:
@mguthaus Perhaps you could try out my PR? If it is accepted, we could expand it to use everything in the gt2n repo without worrying about size, it is all there already, I suppose.

gadfort:
@mguthaus it's the estimated routing congestion heatmap that triggered this, I suspect there are a few more places that needs to be aware of the backside metals. I should also clarify I'm not using ORFS.

mguthaus:
@gadfort Very possible. I'll check it out and fix when I can recreate it.

gadfort:
> @gadfort Very possible. I'll check it out and fix when I can recreate it.

@mguthaus it's very likely all you need to do it open the GUI and get the estimated routing congestion heatmap. If not I'm happy to try and debug a little further as well when I have time

mguthaus:
@maliberty How do I fix this "Scan Code with pre commit triger" check fail? It seems it doesn't like the config.mk files for the new designs.

maliberty:
> @maliberty How do I fix this "Scan Code with pre commit triger" check fail? It seems it doesn't like the config.mk files for the new designs.

I had to white-list the new platform

maliberty:
Have you run these designs?  gcd fails for me.  At a minimum you need to add the platform
```
--- a/flow/scripts/variables.mk
+++ b/flow/scripts/variables.mk
@@ -17,7 +17,7 @@ export UTILS_DIR     ?= $(FLOW_HOME)/util
 export SCRIPTS_DIR   ?= $(FLOW_HOME)/scripts
 export TEST_DIR      ?= $(FLOW_HOME)/test
 
-PUBLIC=nangate45 sky130hd sky130hs asap7 ihp-sg13g2 gf180
+PUBLIC=nangate45 sky130hd sky130hs asap7 ihp-sg13g2 gf180 gt2n
```
but even then
```
[ERROR] LEF Cell 'gt2_6t_tapbspdn_w31_lvt' has no matching GDS/OAS cell. Cell will be empty.
```

flow/platforms/gt2n/config.mk
```diff
@@ -0,0 +1,85 @@
...
+
+export LIB_FILES = $(PLATFORM_DIR)/lib/gt2_6t_w31_lvt_tt_0p7v25c.lib \
+                   $(ADDITIONAL_LIBS)
+export GDS_FILES = $(PLATFORM_DIR)/gds/gt2_6t_std_cells_w31_lvt.gds \
+                   $(ADDITIONAL_GDS)
+
+# Cells that should not be used in synthesis/optimization
+export DONT_USE_CELLS = gt2_6t_filler_w31_lvt gt2_6t_tapbspdn_w31_lvt \
+                       gt2_6t_tapfspdn_w31_lvt gt2_6t_decapcc_w31_lvt \
+                       gt2_6t_tiehigh_w31_lvt gt2_6t_tielow_w31_lvt
```
maliberty: Why would synthesis target these cells?

flow/platforms/gt2n/config.mk
```diff
@@ -0,0 +1,85 @@
...
+# Place
+#---------------------------------------------------------
+export PLACE_DENSITY ?= 0.40
+
+#---------------------------------------------------------
+# Route
+#---------------------------------------------------------
+export MIN_ROUTING_LAYER = M2
+export MIN_CLK_ROUTING_LAYER = M3
+export MAX_ROUTING_LAYER = M5
```
maliberty: Why so low?

maliberty (commented):
Any reason to import just one Vt?

flow/platforms/gt2n/config.mk
mguthaus: What do you mean?

mguthaus: With backside, there is a lot more routing on low layers...

flow/platforms/gt2n/config.mk
mguthaus: tiehigh and tielow? It's always good practice to flag these in case the tools don't distinguish. It is good practice to also flag clock cells and delay cells as well, but I didn't do that.

mguthaus:
> Have you run these designs? gcd fails for me. At a minimum you need to add the platform
> 
> ```
> --- a/flow/scripts/variables.mk
> +++ b/flow/scripts/variables.mk
> @@ -17,7 +17,7 @@ export UTILS_DIR     ?= $(FLOW_HOME)/util
>  export SCRIPTS_DIR   ?= $(FLOW_HOME)/scripts
>  export TEST_DIR      ?= $(FLOW_HOME)/test
>  
> -PUBLIC=nangate45 sky130hd sky130hs asap7 ihp-sg13g2 gf180
> +PUBLIC=nangate45 sky130hd sky130hs asap7 ihp-sg13g2 gf180 gt2n
> ```
> 
> but even then
> 
> ```
> [ERROR] LEF Cell 'gt2_6t_tapbspdn_w31_lvt' has no matching GDS/OAS cell. Cell will be empty.
> ```

I did. They renamed this cell when I updated the LEF files, but I failed to update the GDS file. It did not give me an error though, which is interesting. I'll try to see why.

flow/platforms/gt2n/config.mk
maliberty: Is there any reason to disallow using the higher layers?

mguthaus: No. Just did it since it was a small design.

mguthaus:
> Any reason to import just one Vt?

Trying to keep the changes simple to start.

flow/platforms/gt2n/config.mk
maliberty: The platform is shared by all designs.  All designs are free to use fewer layers.

mguthaus: Fixing.

mguthaus:
@maliberty I made a boneheaded move after I updated a "fix" in the PDK (which split the tap cell into a fs and bs variant). The PDK has a bug and no GDS as you point out. I filed an issue with them https://github.com/azadnaeemi/GT2N/issues/16

The second time I tested, I mistyped "DESGIN_CONFIG" so it defaulted to the nangate45 tech which unsurprisingly worked :) The CI didn't flag it because it wasn't added to CI yet. I'm now reverting to the old PDK until it can be fixed and adding the platform to the PUBLIC variable.

After it is fixed, I also plan to add support for hybrid front- and back-side supply networks using the two tap cell variants. In this case, the power supply would come from the top-level IOs rather back-side bumps. It will also require some consideration of tap cell placement for the supply which is not yet done.

maliberty:
The upstream issue is fixed.

mguthaus:
@maliberty It seems there are still baseline failures with the master branch that prevent this from succeeding:

> sky130hd microwatt (base)
>   Flow reached last stage.
>   Found 2 metrics failures.
>       [ERROR] detailedroute__antenna__violating__nets fail test: 1 <= 0
>       [ERROR] detailedroute__antenna_diodes_count fail test: 1511 <= 1406

maliberty:
uW was fixed over the weekend but it can be ignored for this PR.

mguthaus:
@maliberty ok, is that on a private repo? It's still a failure on the master branch for me.

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/01bb7c248290bbd84ae7738872853594c7a031da

maliberty:
The OR submodule needs updating, it passes with the head of OR.  In any case not a worry here.

flow/platforms/gt2n/itf_to_rc.py
maliberty, line 1: Why is this needed as there is no itf in this PR?

mguthaus: This takes the itf from GT2N repo and makes the setRC.tcl. I can remove it if you want, but figured people may want the explicit assumptions in a script.

maliberty:
Please add a rules-base.json for each new design with a `make metadata` followed by `make update_ok`

flow/platforms/gt2n/itf_to_rc.py
maliberty: Add a descriptive comment telling people where to find the ITF in that case.

mguthaus: This was in the setRC.tcl file header. I'll add it to the script too.

maliberty (approved):

PR #4286, opened 2026-06-10
Fix logic to extract design stage

dsengupta0628:
This design in was at placement stage yet had propagated clocks
There is no set_propagated_clock in the SDC - clocks should be ideal at placement. 

But flow/scripts/open.tcl runs set_propagated_clock [all_clocks] whenever design_stage >= 4 (i.e. CTS has run). design_stage comes from extract_stage  in flow/scripts/util.tcl, which regex-matched the stage number off the full path:

`  regexp {/([0-9])_(([0-9])_)?} $input_file ...`

  For this design the input path is **.../SomeCell/4_something/3_place.odb**. The regex matched the first /<digit>_ in the path - `4_something` -  not `3_place.`  So design_stage came out as 4, open.tcl thought CTS had run, and propagated the clocks at placement → CRPR → tag explosion →  OOM (or livelock in my case with debug build).

We fix the regexp to give current design stage now


maliberty (approved):

oharboe:
:partying_face:

PR #4292, opened 2026-06-14
flow: log_cmd the unlogged synth and load steps for timing

oharboe:
The `synth_odb` step runs `load_design` and the `PRE`/`POST SYNTH` step-tcl hooks, and `load.tcl` runs `read_verilog`, without wrapping them in `log_cmd`. Their wall time is therefore not attributed in the per-command timing log, leaving a gap when accounting for where the synth step spends its time.

`link_design`, `read_sdc`, `eliminate_dead_logic`, `write_db` and `write_sdc` already log themselves; this wraps the remaining ops to match. No behavior change.

### DIFF
--- a/flow/scripts/load.tcl
+++ b/flow/scripts/load.tcl
@@ -16,11 +16,11 @@ proc load_design { design_file sdc_file } {
   # Read design files
   set ext [file extension $design_file]
   if { $ext == ".v" } {
-    read_lef $::env(TECH_LEF)
-    read_lef $::env(SC_LEF)
+    log_cmd read_lef $::env(TECH_LEF)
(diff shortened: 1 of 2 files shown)


oharboe:
Reworked per review: `source_step_tcl` already `log_cmd`s its `source` (via `source_env_var_if_exists`), and `load_design`'s other internals are already wrapped — so wrapping the whole procs double-logged. The follow-up commit instead wraps the unlogged `read_lef` calls inside `load_design` and reverts the `synth_odb.tcl` wraps; net change is `read_lef`/`read_verilog` logging only.

oharboe:
Not pursuing this one — closing.

PR #3943, opened 2026-03-02
Add numeric elapsed_seconds to metrics JSON

harsh-kumar-patwa:
### Summary

Resolves #2506

`genMetrics.py` stores stage runtimes as human-readable strings (e.g., `"0:04.26"`) in the metrics JSON. This makes them hard to consume programmatically — users have to parse the time format manually for analysis, plotting, or comparison.

This adds numeric `elapsed_seconds` fields (float, in seconds) alongside the existing string fields, so runtimes are directly usable as numbers.

### Changes

**`flow/util/genMetrics.py`**
- In the existing runtime accumulation loop, compute `delta.total_seconds()` for each stage and store as `<stage>__elapsed_seconds`
- Add `total_elapsed_seconds` alongside the existing `total_time` string

### Example output (new fields only)

```json
{
  "synth__runtime__total": "0:00.90",
  "synth__elapsed_seconds": 0.9,
  "globalplace__runtime__total": "0:02.59",
  "globalplace__elapsed_seconds": 2.59,
  "total_time": "0:00:18.900000",
  "total_elapsed_seconds": 18.9
}
```

Existing fields (`__runtime__total`, `total_time`, `__cpu__total`, `__mem__peak`) are unchanged. The hierarchical JSON format (`-x` flag) also works correctly since the new keys follow the same `stage__metric` naming convention.

PR #4302, opened 2026-06-23
flow/docs/rcx: add GRT-vs-RCX parasitic divergence study

oharboe:
_Rendered copy of [`flow/docs/rcx/README.md`](flow/docs/rcx/README.md) for review._

# Global-route WNS vs RCX sign-off: a fan-out study

As a design moves through ORFS, timing is reported several times, each time with
a different parasitic model. The pre-route numbers are estimates; only the
post-route `finish` number uses parasitics extracted from the routed geometry
(OpenRCX). This study measures how far the global-route (GRT) estimate sits from
the RCX sign-off as a function of net fan-out and wire length, across four PDKs.

The short version: on a 7 nm node the gap is large enough that GRT and RCX
**disagree about whether a design meets timing**. At 130 nm it is negligible. At
2 nm there is no extraction model shipped at all, so there is nothing to check
the estimate against.

## Three parasitic models, three answers

OpenROAD uses three parasitic models at three points in the flow
([discussion #3943][3943]):

| Stage | Parasitics | Command | Notes |
|-------|-----------|---------|-------|
| placement | one default layer's R/C × estimated length | `set_wire_rc` | roughest |
| global route | per-layer R/C × global-route length | `set_layer_rc` / `estimate_parasitics -global_routing` | better topology, still lumped, coupling-blind |
| finish | extracted from detailed-routed geometry | OpenRCX `extract_parasitics -ext_model_file $RCX_RULES` | sign-off |

The global-route estimate is a linear per-layer model — capacitance ≈ (per-layer
C) × (wire length) — whose constants live in each platform's `setRC.tcl`. It
ignores coupling capacitance (which depends on neighbouring routing that does not
exist yet at global route) and is only as accurate as those hand-entered
constants. The resizer documentation is explicit about the consequence:

> "Placement-based parasitics cannot accurately predict routed parasitics, so a
> margin can be used to 'over-repair' the design to compensate." — [rsz README][rsz]

ORFS ships `flow/util/correlateRC.py` (`make correlate_rc`) to fit those
constants to RCX and report the residual gap, but it is an offline, whole-PDK
calibration step. Nothing in the flow reports whether the nets in a *particular*
design fall inside the range where the linear model holds. The platform
bring-up guide ([PlatformBringUp][bringup]) documents how to write `setRC.tcl`
but not how to validate it against extraction, and the coarseness of the model
is a known issue ([#3969][3969]).

## The experiment

`docs/rcx/gen_study.py` emits deliberately minimal designs: one launch flop (the
*hub*) drives a single net that fans out to **N** capture flops. Inputs are
pinned to the west edge and outputs to the east edge, so the fan-out net spans
the die; the hub is `dont_touch` so the resizer cannot clone the driver and chop
the net into short, easily-estimated segments. We sweep **N = 1 … 128**, run the
full flow, and read WNS at each stage plus the per-net GRT and RCX parasitics
(`make write_net_rc`). WNS is normalized by the clock period so curve *shape* is
comparable across PDKs.

The shaded bands on the plots mark where the estimate is expected to hold:
fan-out ≤ 16 (green) is at or below the logical-effort FO4 region and the usual
`set_max_fanout` floor, where the Steiner-tree topology is accurate; 16–64
(amber) is where buffer trees and coupling error grow; > 64 (red) is past the
point where the lumped model is trustworthy. Fan-out here is a stand-in for the
real driver, long coupling-heavy nets, which the wide die also exercises.

## Results

Plots and `plots/study_data.csv` are produced by `bazelisk run
//flow/docs/rcx:update` and committed so they render here directly.

WNS gap (RCX − GRT) across PDKs, normalized by clock period

On **asap7 (7 nm)** the gap is large and changes sign with fan-out. At fan-out 16
GRT is about 50 ps pessimistic; at fan-out 128 the design reports **+21 ps slack
at global route but −2 ps after extraction** — GRT says it closes, RCX says it
does not. On **sky130hd and ihp-sg13g2 (130 nm)** the gap stays within a fraction
of a picosecond across the whole sweep. **gt2n (2 nm)** ships no OpenRCX deck, so
`finish` repeats the GRT estimate and the gap is identically zero — there is no
extraction to compare against.

normalized WNS per flow stage vs fan-out, asap7

Per net, the GRT capacitance estimate is off by tens of percent even where the
WNS summary looks healthy; the errors partly cancel along a path, which is why
they do not surface in the WNS number alone.

per-net GRT vs RCX wire capacitance, asap7

`docs/rcx/rcx_divergence_report.py` ranks the worst nets for one design (name,
fan-out, routed length, GRT vs RCX capacitance, % error).

## Tuning, and whether it is a bug

It is a known calibration limitation, not a defect. The available levers are
blunt and global: re-derive `setRC.tcl` from the RCX deck with `make
correlate_rc`, toggle `ENABLE_RESISTANCE_AWARE`, or apply `repair_design
-cap_margin/-slew_margin` to over-repair. None of them report *which* net in a
given design is mis-estimated, and OpenRCX is calibrated to a field solver
([calibration][rcxcal]) while the GRT estimate is not calibrated per design.

## A proposal

Global route already maintains a DRC-marker database (`dbMarkerCategory "Global
route"`) and tags the offending nets on its congestion markers. The same
machinery could surface parasitic-estimation risk:

1. a per-net GRT-vs-RCX divergence report (or an `estimate_parasitics` mode that
   emits per-net estimated R/C beside the routed extraction), so the gap is
   queryable without dumping and diffing two SPEFs;
2. a global-route DRC-marker subcategory — *parasitic estimation out of range* —
   that tags nets whose fan-out / Steiner length / layer span fall outside the
   model's range, visible in the DRC viewer like congestion markers, including
   on designs that route cleanly;
3. a warning naming those nets, with the standard fixes (split the net, pipeline,
   add a fan-out buffer stage).

That turns an end-of-flow surprise into an early signal.

## Reproduce

```bash
python3 flow/docs/rcx/gen_study.py                 # generate the designs (all PDKs)
flow/docs/rcx/run_study.sh asap7                   # run the flow + collect data
bazelisk run //flow/docs/rcx:update                # regenerate plots + study_data.csv
python3 flow/docs/rcx/rcx_divergence_report.py \
    results/asap7/rcx-fanout-128/base/6_net_rc.csv  # worst-net report for one design
```

[3943]: https://github.com/The-OpenROAD-Project/OpenROAD/discussions/3943
[3969]: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3969
[rsz]: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/rsz/README.md
[bringup]: https://openroad-flow-scripts.readthedocs.io/en/latest/contrib/PlatformBringUp.html
[rcxcal]: https://openroad.readthedocs.io/en/latest/main/src/rcx/doc/calibration.html

### References

- I. Sutherland, R. Sproull, D. Harris, *Logical Effort: Designing Fast CMOS
  Circuits*, Morgan Kaufmann, 1999. (FO4, `set_max_fanout`.)
- C. Chu, Y.-C. Wong, "FLUTE: Fast Lookup Table Based Rectilinear Steiner Minimal
  Tree Algorithm," ICCAD 2004 / IEEE TCAD 2008. (Steiner wire-length accuracy.)
- Liu et al., "Bridging the Gap between Global Route and Detailed Route … for
  Wire Parasitics and Delay Prediction," arXiv:2305.06917, 2023.
- L. Clark et al., "ASAP7: A 7-nm FinFET predictive PDK," Microelectronics
  Journal, 2016.


maliberty:
> On sky130hd and ihp-sg13g2 (130 nm) the gap stays within a fraction of a picosecond across the whole sweep.

This fails the sniff test (I wish we were that accurate).  I looked at sky130 fanout-128 and the worst path delay is 2.256 (grt) vs 2.516 (drt).   Note the time unit in these pdks is ns not ps so you statement is 1000X off.

maliberty:
On asap7 the case that goes from passing to failing at fanout=128 a data arrival time of 321.70 (grt) vs 336.17 (drt).  The required time is 344.07 (grt) vs 334.21 (drt).  There is no single large error but a small accumulation of small errors.  Given the path delays, the 24.33 ps of slack difference is not an unreasonable percentage.  Its a bit of bad luck that arrival slows down while required speeds up so the errors add rather than cancel.

In short I don't see much to fix.  The fact that the error bar straddles 0 isn't consequential by itself.

oharboe:
needs work

PR #4325, opened 2026-07-02
RFC: install bazel-packaged yosys instead of compiling from scratch

oharboe:
**Concept PR for comments — not intended to merge as-is.**

## What

Stop compiling yosys from scratch with ad-hoc git-clone-and-make code in
`DependencyInstaller.sh` and `bazel/install.sh`. Instead, package the
hermetic bazel-built yosys (binary + `yosys-abc` + share tree +
`slang.so` plugin) as a relocatable PREFIX tarball, `//:yosys_tar`, and
have the installers build and extract that.

- `patches/bazel-orfs/0001-add-yosys-prefix-tar-packaging.patch` — adds a
  `yosys_prefix_tar()` macro (rules_pkg `pkg_tar`) to bazel-orfs, vendored
  as a `git_override` patch. Offered upstream as
  The-OpenROAD-Project/bazel-orfs#752; this PR works
  identically whether that lands, stalls, or is closed — when it lands we
  drop the patch and bump `BAZEL_ORFS_COMMIT`.
- `BUILD.bazel` — instantiates `//:yosys_tar` with the yosys-slang plugin
  (same label as `orfs.default(yosys_plugins = ...)`), so
  `SYNTH_HDL_FRONTEND=slang` keeps working.
- `etc/DependencyInstaller.sh` — `_installORDependencies` now bootstraps
  bazelisk (which it installed anyway), builds + extracts `//:yosys_tar`
  into the prefix, and passes the installed version to the OpenROAD
  installer so its `_install_yosys` source build becomes a verified
  no-op. The `-yosys-ver` flag and the `tools/yosys/docs/source/conf.py`
  version scraping are gone.
- `bazel/install.sh` — the hand-rolled yosys `make` and yosys-slang
  `cmake` builds are replaced by the same tarball; the
  bison/flex/gawk/cmake host-dependency checks disappear with them.
- `patches/bcr/sed-4.10-glibc-2.43.patch` — the BCR `sed` module (via
  `@yosys → abc → readline → ncurses`) ships a gnulib that does not
  compile against glibc 2.43; `single_version_override` pins sed 4.10
  with two small gnulib fixes. The same patching zen, one level deeper:
  works today, dropped when gnulib/BCR catch up.

## Why

- **Single source of truth.** The yosys version is the `@yosys` pin in
  bazel-orfs `MODULE.bazel` (bumped via `bazelisk run @bazel-orfs//:bump`),
  identical for the bazel flow and the classic make flow. No more
  version scraping, no more OS-specific compile paths in
  `DependencyInstaller.sh` — the same hermetic binary on every OS,
  including macOS.
- **Faster installs.** Bazel caches the yosys build; re-running the
  installer is a cache hit instead of a re-clone and full rebuild.
- **Gradual, additive migration.** This simplifies maintainers' lives
  *before* the switch to bazel is complete: the old bash scripts shrink
  instead of growing, and each piece is testable in isolation —
  `bazelisk build //:yosys_tar` is a lot easier to exercise than a code
  path buried inside `DependencyInstaller.sh`.
- **A sane patching story.** With yosys consumed as a bazel module we
  can carry patches for yosys (or any other tool) while waiting for
  upstream fixes to land — backports from the next release, macOS
  fixes, etc. — with Claude handling the mechanical patch maintenance.
  The zen of it: every patch is offered upstream as a GitHub PR, and
  the vendored copy keeps working regardless of the upstream PR's fate.
  The amount of patching stays roughly constant over time and will
  never be zero — ORFS/OpenROAD's bleeding edge will always be on the
  bleeding edge. (The `.bazelrc` fallback registry for yosys 0.64,
  pending bazelbuild/bazel-central-registry#8862, is this same pattern
  already in action.)
- **Toward dropping the yosys submodules.** `tools/yosys` and
  `tools/yosys-slang` are no longer needed by these paths; removing
  them is deliberately left for a follow-up.

## Tarball layout

```
bin/yosys
bin/yosys-abc
share/yosys/...
share/yosys/plugins/slang.so
```

Yosys resolves share as `<bindir>/../share/yosys/` and abc as
`<bindir>/yosys-abc`, and the bazel build links readline/tcl/zlib/ffi/abc
statically — the tarball extracts into any prefix unmodified.

## Latent master issues this surfaced (fixed here)

Nothing built root-package or pkg_tar targets from the ORFS root before,
so two pre-existing breakages were invisible:

- `//:install_for_bazel`'s `srcs = ["bazel/install.sh"]` label is invalid
  — `bazel/BUILD` makes `bazel/` its own package, which silently broke
  every root-package target. Fixed with a cross-package label +
  `exports_files`.
- rules_pycross's toolchain computation fails on Python 3.8 (registered
  transitively via or-tools → pybind11_abseil, dropped from
  rules_python 2.0.0's `MINOR_MAPPING`). The workaround already in
  `tools/OpenROAD/MODULE.bazel` is root-honored only, so it is repeated
  in the ORFS `MODULE.bazel`.

## Known gaps / discussion points

- The tarball does not ship the `make install` extras `yosys-config`,
  `yosys-smtbmc`, `yosys-witness`, `yosys-filterlib`. sby's smtbmc
  engines and out-of-tree plugin builds against the installed yosys
  would need them; extending the macro is a follow-up if required.
- eqy/sby keep their existing source builds (pure-Python YosysHQ tools,
  no bazel modules yet); they build against the bazel-installed yosys.
- The bazel path needs network access to BCR/registries; there is no
  `-nocert` equivalent plumbed through yet.
- Deleting `_install_yosys` from the OpenROAD installer is a separate
  OpenROAD PR once nothing exercises it.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,6 +1,23 @@
+load("@bazel-orfs//:yosys_pkg.bzl", "yosys_prefix_tar")
 load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
 
 sh_binary(
     name = "install_for_bazel",
-    srcs = ["bazel/install.sh"],
+    # bazel/ is its own package (bazel/BUILD), so a cross-package label
+    # is required here; bazel/install.sh silently broke every root-package
+    # target once bazel/BUILD appeared.
+    srcs = ["//bazel:install.sh"],
+)
+
+# Relocatable PREFIX tarball of the hermetic bazel-built yosys
+# (bin/yosys, bin/yosys-abc, share/yosys/..., share/yosys/plugins/slang.so).
+# This is the single source of truth for the yosys that
+# etc/DependencyInstaller.sh and bazel/install.sh install; the version is
+# the @yosys pin in bazel-orfs MODULE.bazel (bump via
+# `bazelisk run @bazel-orfs//:bump`).
+yosys_prefix_tar(
+    name = "yosys_tar",
+    # Same label as orfs.default(yosys_plugins = ...) in MODULE.bazel, so
+    # SYNTH_HDL_FRONTEND=slang works identically in the classic flow.
+    plugins = ["@yosys-slang//src/yosys_plugin:slang.so"],
 )
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -19,6 +19,7 @@ bazel_dep(name = "rules_shell", version = "0.6.1")
 # our MODULE.bazel doesn't need to be patched at non-root consumption
 # time.
 
+bazel_dep(name = "rules_pycross", version = "0.8.1", dev_dependency = True)
 bazel_dep(name = "toolchains_llvm", version = "1.5.0", dev_dependency = True)
 bazel_dep(name = "openroad", dev_dependency = True)
 local_path_override(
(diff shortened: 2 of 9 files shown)


oharboe:
@gadfort @vvbandeira @maliberty I had the idea that we could use bazel to build yosys + slang rather than DependencyInstaller.sh

maliberty:
There is another use case where we compile yosys with Verific for closed use.  We would also have to figure out how to handle that and avoid caching Verific.  Interesting but not high priority to me.

oharboe:
As usual: there needs to be some immediate problem where the choice is to invest into a dying infrastructure or make the investment to switch...

PR #4327, opened 2026-07-03
flow: repair egregious fanout when writing 1_synth.odb

oharboe:
## Motivation / use-case

Synth is widely used as a fast signal for RTL changes: run `make synth`, look at the worst paths (`gui_synth`, STA on `1_synth.odb`), improve the RTL, repeat. Today that signal is misleading: raw yosys/ABC output leaves egregious high-fanout nets unbuffered, so synth-stage STA charges the full unbuffered fanout load as delay. The worst reported paths are then dominated by fanout artifacts the place-stage resizer later repairs for free — they carry no actionable information, and they hide the real, RTL-actionable logic-depth paths behind them.

## Lessons this encodes

Written down so others don't have to re-learn them:

- **Raw yosys/ABC output is electrically unrepaired.** ABC's own buffering is layout-blind and treated as throwaway by the flow; OpenROAD does all real buffering/sizing later. STA on the raw synth netlist is dominated by fanout artifacts, not logic depth.
- **The floorplan stage does not fix this.** Its repair sequence (`unbuffer,sizeup,swap,vt_swap`) upsizes drivers but deliberately never inserts buffers — buffer-tree construction is deferred to place, where real locations exist.
- **The first buffered netlist in the stock flow is therefore place**, which is too slow for RTL iteration.
- **A context-free, DRC-only repair at synth is the cheapest point to make the signal honest** — exactly where `synth_syn.tcl` already runs its `repair_design -pre_placement` unconditionally before writing `1_synth.odb`.

## What / how

After `eliminate_dead_logic`, before writing `1_synth.odb`:

- `set_max_fanout` (`SYNTH_REPAIR_DESIGN_MAX_FANOUT`, default 64) so only the truly egregious nets are DRC violations — the pass touches the artifacts, not the whole design;
- `estimate_parasitics -placement` (handles unplaced cells);
- `repair_design -max_wire_length 1000000` — wire splitting disabled, since unplaced cells have no meaningful wire length.

Deliberately **not** `repair_design -pre_placement`: that is a whole-design gain-based round (slow on large merged netlists) and conflicts with `-placement` parasitics (`EST-0104`). The recipe here is DRC-only and bounded.

Default **on**, because this is stage semantics rather than an optional extra: the synth stage owns the electrical sanity of its product, matching `synth_syn.tcl`. `SKIP_SYNTH_REPAIR_DESIGN=1` restores the previous behavior, following the `SKIP_REPAIR_TIE_FANOUT` convention.

The canonicalized `1_synth.sdc` is written before the repair block so the repair-scoping `set_max_fanout` does not leak into the flow's SDC (there is no `unset_max_fanout`).

## What this deliberately does not do

It does not touch the floorplan.tcl conditioning block (`repair_tie_fanout`, `replace_arith_modules`, `remove_buffers`/`repair_timing_helper`). Per PR #4187 those transforms depend on floorplan context (bterm placement from `initialize_floorplan`, layer stack from `set_routing_layers`) and regressed setup TNS 1.7–46x when moved to synth. This block is additive and context-free (no die/bterm/layer dependence), like `eliminate_dead_logic` before it.

## QoR hypothesis

Neutral-to-better downstream: floorplan's `repair_timing_helper` and the place-stage `repair_design` start from an electrically sane netlist instead of one with unbuffered fanout-100 nets. **This PR is a draft to let CI capture whether that holds across the design matrix.** If it does, the question for maintainers is whether this matches the intention of the rest of the flow; if it doesn't, the per-design results are the data to decide between raising the fanout bound, defaulting off, or dropping the idea.

### DIFF
--- a/docs/user/FlowVariables.md
+++ b/docs/user/FlowVariables.md
@@ -288,6 +288,7 @@ configuration file.
 | <a name="SKIP_PIN_SWAP"></a>SKIP_PIN_SWAP| Do not use pin swapping as a transform to fix timing violations (default: use pin swapping).| |
 | <a name="SKIP_REPAIR_TIE_FANOUT"></a>SKIP_REPAIR_TIE_FANOUT| Skip repair_tie_fanout at floorplan step.| 0|
 | <a name="SKIP_REPORT_METRICS"></a>SKIP_REPORT_METRICS| If set to 1, then metrics, report_metrics does nothing. Useful to speed up builds.| 0|
+| <a name="SKIP_SYNTH_REPAIR_DESIGN"></a>SKIP_SYNTH_REPAIR_DESIGN| Skip the DRC-only repair_design of egregious high-fanout nets when writing 1_synth.odb.| 0|
 | <a name="SKIP_VT_SWAP"></a>SKIP_VT_SWAP| Do not perform VT swap to improve QoR (default: do VT swap).| |
 | <a name="SLANG_PLUGIN_PATH"></a>SLANG_PLUGIN_PATH| Path to the slang plugin for Yosys. This can be a full path to a custom-built plugin (e.g. for Bazel builds) or just the plugin name.| slang|
 | <a name="SLEW_MARGIN"></a>SLEW_MARGIN| Specifies a slew margin when fixing max slew violations. This option allows you to overfix.| |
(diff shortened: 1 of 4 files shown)


oharboe:
@maliberty Thoughts on this one? This is more of a question w measured data than a specific suggestion.

The QoR impact downstream seem neglible.

That some results are worse tells me little about whether it is an improvement.

maliberty:
> estimate_parasitics -placement (handles unplaced cells);

Handles in the sense of doing nothing perhaps.  There is no meaning to -placement without placement.  I don't think this makes much sense.

oharboe:
Thank you!

This prunes some of the paths that Claude wanted to take us down :+1:

PR #4328, opened 2026-07-03
Fix synth stage runtime parsing when syn enabled

povik:


povik:
@dsengupta0628 FYI, this should fix the unset runtime metric

maliberty (approved):

PR #4333, opened 2026-07-06
synth: single writer for 1_synth.sdc + single-process flow test that enforces it

oharboe:
`1_synth.sdc` had two writers: yosys' `synth.tcl` copied the user's `SDC_FILE` into place, then `synth_odb.tcl` overwrote it in place with the OpenSTA-canonicalized version. `cp` preserves the source's mode bits, so when `SDC_FILE` is a read-only build input (bazel RBE), the copy is read-only and the canonicalization write fails — see OpenROAD [#10818](https://github.com/The-OpenROAD-Project/OpenROAD/pull/10818), which papered over this with `chmod +w`. On writable local checkouts it only accidentally worked.

The first version of this PR made `synth_odb.tcl`/`synth_syn.tcl` write `1_synth.odb`/`.sdc` unconditionally, which — as the review pointed out — breaks the single-OpenROAD-process flow (`WRITE_ODB_AND_SDC_EACH_STAGE=0`, all stage scripts sourced into one process). That mode had no test coverage. This PR now keeps the single-writer fix, restores the gating, and **adds the missing coverage**; the review history is retained as follow-up commits:

- **`flow/scripts/flow.tcl`**: the whole flow (yosys netlist through the final report) as one Tcl script in a single OpenROAD process. With `WRITE_ODB_AND_SDC_EACH_STAGE=0` the top level is the sole writer: `flow_write_db`/`flow_write_sdc` produce exactly the .odb/.sdc file set of the stage-per-process make flow, and `flow_source` **errors if any stage script writes an .odb/.sdc itself** — this assertion catches the unconditional writes from this PR's first version (verified: the flow fails with `synth_odb.tcl wrote .../1_synth.odb` when they are reinstated).
- **Sweep of the same idiom violations** found by the assertion: `io_placement.tcl`/`fillcell.tcl`/`density_fill.tcl` forward the input .odb with a raw `exec cp` on their skip paths (kept — cp is much faster than re-serializing an unchanged DB, and Tcl's `file copy` has second-resolution mtimes that break make's up-to-date checks — but now behind a gated `orfs_copy_db` helper), `write_ref_sdc.tcl` wrote `updated_clks.sdc` ungated, and `detail_route.tcl` `exit`ed on its `SKIP_DETAILED_ROUTE` skip path, which would have killed a single process running the whole flow (the per-stage flow always passes `-exit`).
- **Byte-equality test on asap7/gcd**: `bazelisk test //flow/designs/asap7/gcd:gcd_single_flow_test` runs the single-process flow off the yosys outputs under `FLOW_VARIANT=single` and byte-compares every stage boundary (synth, floorplan, place, cts, grt, route, final × .odb/.sdc) against the regular per-stage flow targets, reusing tools/OpenROAD's `check_same.sh`.

**All seven stage boundaries are byte-identical** between the single-process and per-stage flows. Three OpenROAD fixes were needed, carried as `patches/openroad/` on the `git_override` (pinned to the exact commit the `tools/OpenROAD` submodule records, with submodules, so a patch can reach `src/sta`; `local_path_override` accepts no patches):

1. **sta**: OpenSTA never seeded required times for the never-connected dummy-load pins CTS inserts (`clkload*`) — a bogus −195 ps net slack steered `-resistance_aware` global routing in the process that inserted them, diverging grt (an extra `repair_timing` buffer). Upstreaming is in progress with the OpenSTA maintainers.
2. **drt, incremental updates**: an in-process `detailed_route` after earlier stages iterated nets/instances in incremental-update order rather than odb order, and emitted vias in routing-history order — a fresh process on a byte-identical database routed differently. Nets/insts are reordered to odb order after incremental updates and vias are written in a canonical order (per review: the sort comparator is a strict weak ordering — non-vias first keeping their relative order, then vias by cut layer and origin).
3. **drt, pin access**: every pin access update appended block-terminal access points without destroying the previously written ones, so the per-stage flow accumulated stale access points at each stage that reruns pin access (69 → 138 on asap7/gcd between grt and route).

The final-stage residue was flow structure, not OpenROAD: the per-stage flow writes `6_final.odb` mid-`final_report.tcl` (before RCX extraction reorders every dbWire), so `final_report.tcl` is split around that write and `flow.tcl` writes at the same point.

**OpenROAD is bumped to master `e8230044`** (hermetic-llvm toolchain merged), and this repo's bazel setup adopts the same zero-sysroot toolchain — no host compiler or `/usr/include` in the build, so host glibc changes can't break it. The root-honored parts of tools/OpenROAD's MODULE.bazel are mirrored (gnulib tool version overrides sed/gawk/m4/bison + the bison and tcl_lang patches, `bazel/slang-compat` placeholder), as is the earlier rules_pycross Python 3.8 workaround (extension configuration is root-honored only). `SYNTH_USE_SYN` is declared make-only for gcd's bazel flow (bazel-orfs's synth doesn't stage Verilog sources for the OpenROAD synth step).

With the writes gated, the bazel-orfs side needs no patch ([bazel-orfs#754](https://github.com/The-OpenROAD-Project/bazel-orfs/pull/754) anticipated this; `1_synth.odb`/`.sdc` come from the `do-1_synth` action, which runs with the default `WRITE_ODB_AND_SDC_EACH_STAGE=1`).

Supersedes #4338.


oharboe:
@maliberty TL;DR orfs_write_db is not needed/useful for synthesis, because synthesis is the first stage.

maliberty:
@codex review

maliberty:
I don't follow your comment about WRITE_ODB_AND_SDC_EACH_STAGE.

oharboe:
> I don't follow your comment about WRITE_ODB_AND_SDC_EACH_STAGE.

We can now create a single .tcl for all openroad stages reusing the scripts, WRITE_ODB_AND_SDC_EACH_STAGE is part of that mechanics. It has been a while, I forget the details. The code isn't complicated, just read it a few lines + you have a floorplan+place joined script that you refer to occasionally, so I'm pretty sure you're aware of the feature.

maliberty:
I am aware of the feature and it seems like this PR breaks it.  I don't follow your justification for doing so.

oharboe:
> I am aware of the feature and it seems like this PR breaks it. I don't follow your justification for doing so.

So this feature has no automated tests in CI?

maliberty:
No as with a great many things in ORFS

oharboe:
> No as with a great many things in ORFS

Well... if we care about a feature, we must have tests and tests are so much cheaper than they used to be.

I'll see about wiring up a bazel driven test in ORFS.

I know, I know... CI for `bazelisk test ...` is painful to think about, but at least we can test it locally.

oharboe:
Root-caused and fixed the grt-stage single-process divergence (640e2057a).

**Root cause**: the dummy-load buffers CTS inserts to balance the clock tree (`clkload*`) have dangling outputs that never get a `connectPinAfter` call, so OpenSTA never registers them in the search endpoint set and never seeds their required times. With a propagated clock, such a pin reports slack = 0 − arrival (−195 ps on gcd) instead of +∞ — but only in the process that *inserted* them; a fresh process's full endpoint scan seeds everything correctly from the byte-identical .odb/.sdc. `global_route -resistance_aware` consumes net slacks to pick/order its resistance-aware net set, so the clock nets sorted differently and routing diverged **before any repair ran** — the repair_timing 8-vs-7 buffer delta was downstream fallout. (With `ENABLE_RESISTANCE_AWARE=0` the two flows were already byte-identical end to end.)

**Fix**: 5 lines in `Sta::makeInstanceAfter` (register new pin vertices as endpoint candidates + initialize `makePinVertices` out-params, which are left unassigned for power/ground pins — inserting the garbage pointer segfaults). Carried as `patches/openroad/…` applied via `git_override` pinned to the exact commit the submodule records (with `init_submodules`, so the patch reaches `src/sta`); `local_path_override` doesn't accept patches. Upstreaming to OpenSTA to be sorted separately; the fix commit is preserved on a local branch of the src/sta submodule.

**Usual suspects ruled out with evidence**: global `rand()` state (cts/gpl do perturb it, but rsz/grt/est/dbSta contain no bare `rand()`), grt's shuffle seed (0 = off), SDC writer float precision (gcd constraints are exact integers), rsz static move counters (report-only), odb unique-name counters (serialized in the .odb).

**Test matrix now**: synth / floorplan / place / cts / **grt** byte-identical ✅. route/final still diverge from a second, independent cause — detail route behaves differently in-process than freshly loaded despite byte-identical `5_1_grt.odb` inputs (suspect drt/pin-access module state) — left failing on purpose, documented in the BUILD file, next campaign.

oharboe:
Root-caused and fixed the route-stage divergence as well (87b6286a0).

**Root causes** (fresh-vs-in-process experiment pair on byte-identical `5_1_grt.odb` inputs):
1. drt's incremental design update (taken when `pin_access` already ran in the same process) **appends** recreated nets/instances to its internal lists, while a fresh read follows odb iteration order (odb recycles ids). Track assignment iterates those lists → the wire assignment diverged before detail routing started.
2. drt's final database update emitted each net's vias in history-derived container order (the track-assignment write path already canonicalizes via `mergeSplitConnFigs`; the final path did not).

Both fixed in `patches/openroad/0002-…`. **The routed layout is now identical between the flows** — equal DEF for `5_route` — and the detailed-route logs match line for line.

**Why route/final still fail the .odb byte comparison** (both fully root-caused, recorded in the BUILD file):
- **odb id history**: odb serializes object ids; a database that lived through create/delete cycles (rsz journal in the grt stage) reuses freed ids where a freshly read database appends. Subsequent drt edits then get different ids → different bytes for a byte-identical layout. A read+write round-trip does **not** normalize this; byte convergence needs canonical-id serialization in odb.
- **6_final.odb write point**: the per-stage flow writes it inside `final_report.tcl` *before* the RCX extraction, which calls odb `orderWires` and re-encodes every dbWire; the single-process flow writes at end of script and captures the re-encoded form (this one also shows in the DEF, though the physical layout is the same).

Open question for the follow-up: pursue canonical-id odb serialization (real prize for bazel caching/determinism, but a substantial odb change), compare route/final on canonicalized content (e.g. DEF after orderWires) instead of odb bytes, or move `final_report.tcl`'s write point — input welcome.

oharboe:
@maliberty I need guidance on this one https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/4333#issuecomment-4904766991 to make progress.

Single invocation flow has issues to be fixed(we knew that...), the test now identifies the basic issues for asap7 gcd for the flow to be identical to split invocation.

With a bit of guidance from you, I can ask Claude to complete the patches, then we can create PRs upstream and let the fixes trickle through then clean up here...

patches/openroad/0001-sta-register-new-instance-pin-vertices-as-endpoints.patch
```diff
@@ -0,0 +1,25 @@
...
++          Vertex *bidir_drvr_vertex = nullptr;
+           graph_->makePinVertices(pin, vertex, bidir_drvr_vertex);
++          // Pins that are never connected (eg dangling outputs of
++          // buffers inserted as pure loads) get no connectPinAfter
++          // call, so register the new vertices as endpoint candidates
++          // here or their requireds are never seeded.
++          if (vertex)
++            search_->endpointInvalid(vertex);
++          if (bidir_drvr_vertex)
++            search_->endpointInvalid(bidir_drvr_vertex);
```
maliberty: @dsengupta0628 does this look correct?

maliberty:
> odb id history: odb serializes object ids; a database that lived through create/delete cycles (rsz journal in the grt stage) reuses freed ids where a freshly read database appends. Subsequent drt edits then get different ids → different bytes for a byte-identical layout. A read+write round-trip does not normalize this; byte convergence needs canonical-id serialization in odb.

This is wrong - the free lists are serialized and reloaded.

oharboe:
You're right, and thanks — the free lists are indeed serialized and reloaded (`dbTable.inc` streams `free_list_` both ways); the id/freelist-history explanation was wrong. With that corrected, byte-identical inputs mean equivalent allocation state, so the two databases had to contain **genuinely different objects**. They did.

Per-table stream accounting (the `ODB io_size` debug plus a per-table position probe) put the entire remaining delta in `ap_tbl`: the per-stage flow's `5_route.odb` had **138 block-terminal access points vs 69** in the single-process flow. `io::Writer::updateDbAccessPoints` appends bterm access points without destroying the previously written ones, so **every process that reruns pin access on a design that already has them duplicates them** — the staged flow accumulates stale APs at each stage after `pin_access` first ran, while the single-process flow's in-memory bookkeeping skips unchanged terms and was correct. So the single-process test flushed out a real staged-flow bug here, not an odb limitation.

Fixed in `patches/openroad/0003-…` (a9dededba): destroy previously written bterm APs before writing, in reverse creation order so recycled ids preserve the original assignment. **Route is now byte-identical and its test passes (6/7).** Only `final` remains, from the already-documented write-point issue: the per-stage flow writes `6_final.odb` before the RCX extraction whose `odb::orderWires` re-encodes every dbWire; the single-process flow writes at end of script.

oharboe:
**7/7 — the whole flow is now byte-identical** between the single-process and stage-per-process flows (456770147).

The last stage, `final`, diverged only because of the write point: the per-stage flow writes `6_final.odb` mid-`final_report.tcl` (after `global_connect`, before `deleteRoutingObstructions` and the RCX extraction whose `odb::orderWires` re-encodes every dbWire), while the single-process flow could only write at end of script. Rather than move the write (which would have made `6_final.odb` obstruction-free and changed what `generate_abstract` sees), `final_report.tcl` is now split into `final_connect.tcl` + `final_outputs.tcl` with the gated write between them — per-stage behavior is unchanged byte for byte, and `flow.tcl` sources the parts and performs the write itself at the identical point.

Final test matrix for `bazelisk test //flow/designs/asap7/gcd:gcd_single_flow_test`: synth ✅ floorplan ✅ place ✅ cts ✅ grt ✅ route ✅ final ✅

Summary of what the single-process equivalence test flushed out of OpenROAD along the way, all carried as `patches/openroad/` until upstreamed:
1. OpenSTA never seeds required times for never-connected pins created mid-session (CTS dummy loads) → bogus −195 ps net slacks steering `-resistance_aware` routing (0001).
2. drt's incremental design update orders nets/instances by update history instead of odb order, and its final write emitted vias in history order (0002).
3. drt duplicates block-terminal access points on every pin access rerun — staged flows had been silently accumulating stale APs at every stage since `pin_access` first ran (0003).

maliberty:
@codex review

patches/openroad/0001-sta-register-new-instance-pin-vertices-as-endpoints.patch
dsengupta0628: Yes. I reached out to James regarding upstreaming this fix. Parallelly I am trying to create a standalone testcase that can somehow reproduce this gap too.

patches/openroad/0002-drt-make-incremental-design-updates-match-a-fresh-read.patch
```diff
@@ -0,0 +1,141 @@
...
++        }
++        const auto lv = std::static_pointer_cast<frVia>(lhs);
++        const auto rv = std::static_pointer_cast<frVia>(rhs);
++        return std::make_tuple(lv->getViaDef()->getCutLayerNum(),
++                               lv->getOrigin().x(),
++                               lv->getOrigin().y())
++               < std::make_tuple(rv->getViaDef()->getCutLayerNum(),
++                                 rv->getOrigin().x(),
++                                 rv->getOrigin().y());
++      });
```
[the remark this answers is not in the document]
oharboe: Fixed in 898c39f62: the comparator now partitions non-vias before vias (list::sort is stable, so non-vias keep their relative order) and orders vias by (cut layer, x, y). Each connFig is encoded as its own path in the written wire, so the partition doesn't change the layout; the byte-equality tests still pass with the new order.

patches/openroad/0002-drt-make-incremental-design-updates-match-a-fresh-read.patch
```diff
@@ -0,0 +1,141 @@
...
++    // the written database does not depend on that history; path segs
++    // keep their order.
++    for (auto& it : connFigs_) {
++      it.second.sort([](const std::shared_ptr<frConnFig>& lhs,
++                        const std::shared_ptr<frConnFig>& rhs) {
++        const bool lhs_via = lhs->typeId() == frcVia;
++        const bool rhs_via = rhs->typeId() == frcVia;
++        if (!lhs_via || !rhs_via) {
++          return false;
++        }
```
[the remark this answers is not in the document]
oharboe: Fixed in 898c39f62: the comparator now partitions non-vias before vias (list::sort is stable, so non-vias keep their relative order) and orders vias by (cut layer, x, y). Each connFig is encoded as its own path in the written wire, so the partition doesn't change the layout; the byte-equality tests still pass with the new order.

flow/scripts/flow.tcl
```diff
@@ -0,0 +1,100 @@
...
+  set path $::env(RESULTS_DIR)/$name
+  log_cmd write_db $path
+  lappend ::flow_expected $path
+}
+
+proc flow_write_sdc { name } {
+  set path $::env(RESULTS_DIR)/$name
+  log_cmd write_sdc -no_timestamp $path
+  lappend ::flow_expected $path
+}
```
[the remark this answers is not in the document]
oharboe: Fixed in abb514cfc: flow_write_db/flow_write_sdc build the path with file join now.

flow/scripts/flow.tcl
```diff
@@ -0,0 +1,100 @@
...
+flow_source cts.tcl
+flow_write_db 4_1_cts.odb
+flow_write_db 4_cts.odb
+flow_write_sdc 4_cts.sdc
+
+# Route
+flow_source global_route.tcl
+flow_write_db 5_1_grt.odb
+flow_write_sdc 5_1_grt.sdc
+flow_source detail_route.tcl
```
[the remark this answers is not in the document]
oharboe: Fixed in 7780f820d: detail_route.tcl no longer calls exit — the skip path is an if/else and the script always ends at the gated orfs_write_db, so the single-process flow continues through fillcell/final (the per-stage flow relies on openroad -exit as before). Verified with make DESIGN_CONFIG=designs/nangate45/gcd/config.mk SKIP_DETAILED_ROUTE=1 route.

oharboe:
@maliberty Anything else from me?

oharboe:
@maliberty All done? Anything more from me?

patches/openroad/0001-sta-register-new-instance-pin-vertices-as-endpoints.patch
dsengupta0628: Fixed upstream by https://github.com/parallaxsw/OpenSTA/commit/c4c811ad2ee9670fc5009ce18f266d2db7663e70 - it will be available whenever I pull the latest upstream code (I am just waiting for one of my PRs to be merged there before I do the pull)

oharboe: Nice!

@maliberty Merge as is? This patch will wash out after OpenROAD is upgraded to latest sta.

maliberty:
Anytime you are adding patch files I'm going to want to see upstream PRs that correspond.

oharboe:
> Anytime you are adding patch files I'm going to want to see upstream PRs that correspond.

Have I let u down?

oharboe:
> > Anytime you are adding patch files I'm going to want to see upstream PRs that correspond.
> 
> Have I let u down?

Updating sta in openroad is in your court....

This PR is completed.

oharboe:
Pushed a commit that drops the OpenROAD pinning and keeps everything else (175ffce34):

- `MODULE.bazel`: `git_override(commit=…, patches=…)` → `local_path_override`
- removed `patches/openroad/`
- un-bumped the `tools/OpenROAD` gitlink back to origin/master (`f5df0085`)

Rationale: the three fixes belong upstream, tracked in their own OpenROAD / OpenSTA PRs, not carried in ORFS. bazel here is a **local** testing workflow (not CI — CI exercises the checked-in submodule via `make`), and `local_path_override` builds whatever OpenROAD the developer has checked out. So `gcd_single_flow_test` proves the flow-script changes are correct and will keep working as the tracked fixes land, without ORFS pinning a specific OpenROAD hash.

### Local demonstration — all seven boundaries byte-identical

With `local_path_override` restored, I set up a live `tools/OpenROAD` checkout carrying the still-unlanded fixes (nothing committed to ORFS):

- `tools/OpenROAD` recent master + the two drt fixes applied to the working tree
- `src/sta` + the sta endpoint fix

```
$ bazelisk test //flow/designs/asap7/gcd:gcd_single_flow_test
//flow/designs/asap7/gcd:gcd_single_flow_synth_test        PASSED
//flow/designs/asap7/gcd:gcd_single_flow_floorplan_test    PASSED
//flow/designs/asap7/gcd:gcd_single_flow_place_test        PASSED
//flow/designs/asap7/gcd:gcd_single_flow_cts_test          PASSED
//flow/designs/asap7/gcd:gcd_single_flow_grt_test          PASSED
//flow/designs/asap7/gcd:gcd_single_flow_route_test        PASSED
//flow/designs/asap7/gcd:gcd_single_flow_final_test        PASSED
Executed 7 out of 7 tests: 7 tests pass.
```

`grt` PASS confirms the sta endpoint fix is sufficient; `route`/`final` PASS confirm the two drt fixes are sufficient. The single-process flow (`flow.tcl`, `WRITE_ODB_AND_SDC_EACH_STAGE=0`) is byte-for-byte identical to the per-stage flow at every stage boundary.

### Note on the OpenSTA fix

The endpoint fix has landed in **parallaxsw/OpenSTA master**, but that upstream has no bazel `BUILD` file — OpenROAD's bazel build consumes OpenSTA from the **The-OpenROAD-Project fork** (which carries the hand-written `BUILD`), and the fork hasn't merged parallax yet. So there is no OpenSTA ref today with *both* the fix and bazel build support; the fix still has to be applied locally for the bazel flow until the fork syncs parallax. Like the drt fixes, it stays tracked upstream and is **not** carried in ORFS.

**Net:** ORFS keeps the origin/master `local_path_override` policy — set up for local testing against the live submodule rather than pinning an OpenROAD hash.

oharboe:
@maliberty Good to go now?

Same as on master: no openroad bump and no openroad/sta patches carried here anymore.

ORFS bazel is a local testing workflow and uses the checked out openroad+sta.

The fixes to OpenROAD and sta will trickle into master in their PRs, there's no CI concern here.

MODULE.bazel
maliberty, line 1: How are these changes related to the stated topic?

oharboe: That's where the whole thing started. I found we were missing tests for this stuff, then I added tests, then I found more yaks to shave.

Would you prefer this split into one PR per concern or just do a big review?

maliberty (approved):

oharboe:
:relieved:

PR #4335, opened 2026-07-06
Secure or update sta latest 06/30

dsengupta0628:
Update OpenSTA to 6/30th latest has some metrics variation (minor), Also mode and scene are now TCL object and not string so some save GIU scripts need updation too


dsengupta0628:
I didn't see any QoR change in the secure CI runs https://jenkins.openroad.tools/view/all/job/OpenROAD-flow-scripts-Private/job/secure-or_update_sta_latest_0629/

maliberty (approved):

PR #4342, opened 2026-07-08
bazel: keep bazel_dep(bazel-orfs) non-dev for load() visibility

oharboe:
flow/BUILD has load("@bazel-orfs//:openroad.bzl", "orfs_pdk"), which needs @bazel-orfs visible from @orfs even when ORFS is consumed as a non-root module. Marking the bazel_dep dev-only suppresses that visibility, and every non-root consumer (bazel-orfs itself, its gallery and downstream-test workspaces) has to patch it back in at consumption time (bazel-orfs patches/0036).

Only the git_override needs to be dev-conditional, and it already is: overrides are implicitly root-only. At non-root the consumer's own bazel-orfs module satisfies the dep; when @orfs is the root module, dev deps are honoured anyway, so 'bazel mod graph' output is byte-identical before and after this change.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -40,7 +40,7 @@ git_override(
 # root-only) is dev-conditional — at non-root the consumer's own bazel-orfs
 # satisfies this dep. When @orfs is root, dev deps are honoured anyway, so
 # the resolved module graph is unchanged.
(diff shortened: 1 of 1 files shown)


maliberty:
@codex review

maliberty (approved):

PR #4343, opened 2026-07-08
Fix OPT_POST_GRT_WNS and disable for some PDKs

jfgava:
There is no need to run vt_swap and reroute for PDKs that don't benefit from it.
In this PR I disabled this optimization for Sky130HD/HS, GF180, Nangate45, IHP-SG13G2.

The reroute move does not account for capacitance, which is more relevant to timing closure on these PDKs.
It is possible to reroute a net to a lower-resistance path, but at significantly higher capacitance, which degrades timing.

Currently, we are missing an estimate_parasitics call before running the additional optimizations.
And as we are not changing placement with Reroute and VtSwap, there is no need for a legalization + incremental GRT post-repair_setup.

This prevents issues with another PR with Reroute move enhancements

designs/sky130hd/microwatt/rules-base.json updates:
| Metric                                        | Old        | New        | Type     |
| ------                                        | ---        | ---        | ----     |
| globalroute__antenna_diodes_count             |       1747 |       1692 | Tighten  |
| detailedroute__antenna__violating__nets       |          0 |          1 | Failing  |
| detailedroute__antenna_diodes_count           |       1297 |       1295 | Tighten  |
| finish__timing__setup__tns                    |     -312.0 |     -307.0 | Tighten  |
| finish__timing__hold__tns                     |      -5.02 |      -4.65 | Tighten  |

### DIFF
--- a/flow/scripts/global_route.tcl
+++ b/flow/scripts/global_route.tcl
@@ -87,14 +87,16 @@ proc global_route_helper { } {
       report_metrics 5 "global route post repair timing"
     }
 
-    if { $::env(OPT_POST_GRT_WNS) } {
-      log_cmd global_route -start_incremental
-      log_cmd detailed_placement {*}$dpl_args
-      log_cmd check_placement -verbose
-      # Route only the modified net by DPL
-      log_cmd global_route -end_incremental {*}$res_aware \
-        -congestion_report_file $::env(REPORTS_DIR)/congestion_post_repair_timing_opt_wns.rpt
+    log_cmd global_route -start_incremental
+    log_cmd detailed_placement {*}$dpl_args
+    log_cmd check_placement -verbose
+    # Route only the modified net by DPL
+    log_cmd global_route -end_incremental {*}$res_aware \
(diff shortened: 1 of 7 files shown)


flow/scripts/global_route.tcl
```diff
@@ -109,15 +111,6 @@ proc global_route_helper { } {
...
     }
-
-    # Running DPL to fix overlapped instances
-    # Run to get modified net by DPL
-    log_cmd global_route -start_incremental
-    log_cmd detailed_placement {*}$dpl_args
-    log_cmd check_placement -verbose
-    # Route only the modified net by DPL
-    log_cmd global_route -end_incremental {*}$res_aware \
-      -congestion_report_file $::env(REPORTS_DIR)/congestion_post_repair_timing.rpt
```
maliberty, line 120: Why is this no longer required?

jfgava: I'm moving this part before the additional repair_timing.
VtSwap and Reroute do not change placement. So we don't need to legalize and run GRT incremental after it.

maliberty (approved):

PR #4352, opened 2026-07-14
flow: remove unused MAX_REPAIR_TIMING_ITER variable

oharboe:
`MAX_REPAIR_TIMING_ITER` is declared in `flow/scripts/variables.yaml` but read
by no flow script. `git grep` finds it only in `variables.yaml`, its generated
`variables.json` twin, and the generated `docs/user/FlowVariables.md`.

History: it was added in da116e8e1 and used on the `floorplan_to_place` target,
but that sole consumer was removed in 547e8c88b ("simplify
floorplan_to_place.tcl"), orphaning the declaration.

A hard `repair_timing` iteration cap is also the wrong lever going forward — the
adaptive plateau-stop / `-effort` direction (The-OpenROAD-Project/OpenROAD#10900)
supersedes it. So this removes the dead variable rather than leaving a knob that
does nothing.

`variables.json` and `FlowVariables.md` were regenerated with the checked-in
`flow/scripts/yaml_to_json.py` and `flow/scripts/generate-variables-docs.py`
(the yaml-test CI job's `git diff --exit-code` checks pass).


oharboe:
@eder-matheus I think this is a failure on pr-merge, not pr-head as such.

oharboe:
@eder-matheus pr-merge green, ready for review. Never mind pr-head.

oharboe:
@eder-matheus Ping? pr-merge is green...

eder-matheus (approved):

PR #4359, opened 2026-07-15
Converted setRC correlation doc to markdown

jeffng-or:
Converted the original set RC correlation doc to markdown

Updated private PDK references to use asap7 instead

Updated content to align with using segment mode and not net mode. Will align with Makefiles once @AcKoucher updates the makefiles to remove net mode.

@precisionmoon FYI


oharboe:
:eyes:

maliberty (approved):

PR #4372, opened 2026-07-21
Exported ability to set GRT seed in a flow variable

jeffng-or:
Was interested in running a seed sweep using the AutoTuner, so enabling the ability to set the GRT seed through a flow variable.

@eder-matheus is this the proper way to do this in GRT?

@maliberty FYI. I ran idecode through a six seed sweep and am running is_int through the same.


flow/scripts/variables.yaml
```diff
@@ -1610,3 +1610,10 @@ OPT_POST_GRT_WNS:
     VT swap and wire rerouting only to minimize placement and routing disturbance
   stages:
     - grt
+GRT_SEED:
+  description: >
+    Global route seed
```
maliberty: This is rather terse

jeffng-or: Updated to match GPL_RANDOM_SEED description. Note that https://openroad.readthedocs.io/en/latest/main/src/grt/README.html#set-global-routing-randomness doesn't say much either.

maliberty: I don't see a new commit - forgot to push or label?

jeffng-or: Forgot to resync to public. Will do so after latest change goes through private CI.

jeffng-or:
@maliberty this one should be ready with the updated text.

maliberty (approved):

PR #4388, opened 2026-08-03
refactor, restructure and add some checks to our write RC mechanism

### LINKED ISSUE #4383: The scripts and methodology for RC correlation need clean up and clarification
### Description

There are many unused things and we can have a more straight forward way of running this auxiliary mechanism.

### Suggested Solution

_No response_

### Additional Context

_No response_

AcKoucher:
Resolve #4383.

## Functional Changes
 - Bundle `write_net_rc` and `write_segment_rc` into `write_rc` which now generates two .csv files: one for the nets' data, another one for the segments' data.
 - Make **correlateRC.py** able to consume both files via two keys: `-nets_rc_file` and `-segments_rc_file`. The net data is used only for plotting parasitics differences between GRT/DRT. The segment data is used for the actual fitting and generation of `set_layer_rc` and `set_wire_rc` commands.
 
## Refactoring Changes
 - We now have a single script **write_rc.tcl** responsible for generating the net and segment data. The functions used on it are defined in **write_rc_helper.tcl**.
 - I applied the same file structure to **correlateRC.py** which was too big and messy. Its functions definitions are now inside **correlateRCHelper.py**.

## What's New in correlateRC.py
 - `-segments_rc_file` is now required, as the `set_layer_rc` and `set_wire_rc` commands are the product of every run.
 - `-plot_cap`/`-plot_res` require `-nets_rc_file`.
 - Each .csv starts with a **header naming its data and listing the layer stack** so a file given to the wrong key is caught instead of being parsed as the other kind of data.
 - The header must come before any data row, the rows must have the expected number of columns, and every segment's layer must be listed in the header.
 - Files combined under the same key must have been written for the same routing layers.
 - Empty inputs fail instead of producing empty results.
 - We inform the user if there are nets/segments with zero parasitics.
 - Commands are generated before plots to prevent a plot issue from interfering with it.

## Collateral
 - Adjusted the `make` directives `correlate_rc` and `correlate_platform_rc` to use the new mechanism.


AcKoucher:
@jeffng-or FYI.

eder-matheus:
@codex review

eder-matheus (approved):

PR #4393, opened 2026-08-04
Move https://github.com/The-OpenROAD-Project/FakeRAM2.0 into ORFS

maliberty:
No need to keep it separate anymore.


maliberty:
@oharboe @jeffng-or FYI

maliberty:
I fixed the black python formatting but will leave the general code comments for future work as this is just a code move not a refactor.

PR #4412, opened 2026-08-11
flow: add RTLMP_KEEP_CLUSTERING option

jorge-ferreira-pii:
rtl_macro_placer can run with the -keep_clustering_data flag (which is off by default), so the physical hierarchy that its clustering engine builds is kept in the database as dbGroups and stays inspectable after the run: The Web GUI Hierarchy Browser panel lists the clusters, and save_image -web -display_option {cluster_view true} plots it headless.

Part of issue [#7959](https://github.com/The-OpenROAD-Project/OpenROAD/issues/7959)

### DIFF
--- a/flow/scripts/macro_place_util.tcl
+++ b/flow/scripts/macro_place_util.tcl
@@ -53,6 +53,8 @@ if { [find_macros] != "" } {
   append_env_var additional_rtlmp_args RTLMP_BOUNDARY_WT -boundary_weight 1
   append_env_var additional_rtlmp_args RTLMP_NOTCH_WT -notch_weight 1
   append_env_var additional_rtlmp_args RTLMP_RPT_DIR -report_directory 1
+  append_env_var additional_rtlmp_args RTLMP_KEEP_CLUSTERING \
+    -keep_clustering_data 0
   append_env_var additional_rtlmp_args RTLMP_FENCE_LX -fence_lx 1
   append_env_var additional_rtlmp_args RTLMP_FENCE_LY -fence_ly 1
   append_env_var additional_rtlmp_args RTLMP_FENCE_UX -fence_ux 1
--- a/flow/scripts/variables.yaml
+++ b/flow/scripts/variables.yaml
@@ -1448,6 +1448,13 @@ RTLMP_ARGS:
     Overrides all other RTL macro placer arguments.
   stages:
     - floorplan
+RTLMP_KEEP_CLUSTERING:
+  description: >
+    Set to 1 to store the RTL macro placer's clustering hierarchy in the ODB as
+    nested dbGroups.
+  default: 0
+  stages:
+    - floorplan
 GDS_ALLOW_EMPTY:
   description: >
     Single regular expression of module names of macros that have no .gds file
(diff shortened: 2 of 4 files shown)


AcKoucher:
I don't think `-keep_clustering_data` should be on by default. It's for debugging and could possibly interfere with other types of groups during placement.

jorge-ferreira-pii:
> I don't think `-keep_clustering_data` should be on by default. It's for debugging and could possibly interfere with other types of groups during placement.

I wasn't aware it could interfere with other group types. I've updated the default value of -keep_clustering_data to false.

flow/scripts/macro_place_util.tcl
```diff
@@ -53,6 +53,9 @@ if { [find_macros] != "" } {
   append_env_var additional_rtlmp_args RTLMP_BOUNDARY_WT -boundary_weight 1
   append_env_var additional_rtlmp_args RTLMP_NOTCH_WT -notch_weight 1
   append_env_var additional_rtlmp_args RTLMP_RPT_DIR -report_directory 1
+  # Off by default: an inspection aid that most runs do not need.
```
AcKoucher: Redundant

flow/scripts/variables.yaml
```diff
@@ -1469,6 +1448,23 @@ RTLMP_ARGS:
...
+    The clustering itself runs either way, so the only cost is storing it -
+    around 1% of the macro placement runtime on the largest ORFS cluster tree
+    (nangate45/bp_quad, 220 macros, 98 clusters) and a few KB of ODB. On a
+    design that uses power domains or region groups, the instances already
+    owned by one of those stay in it and are left out of the clustering data
+    (reported once as MPL-0078), since an instance belongs to a single dbGroup -
+    so those cells show up uncolored in the Clusters view. Ignored when
+    RTLMP_ARGS is set, which replaces the whole argument list.
+  stages:
+    - floorplan
```
AcKoucher, line 1457: default: 0

flow/scripts/variables.yaml
```diff
@@ -1469,6 +1448,23 @@ RTLMP_ARGS:
...
+    panel then lists the clusters and can color the layout by cluster, and
+    `save_image -web -display_option {cluster_view true}` plots it headless.
+    The clustering itself runs either way, so the only cost is storing it -
+    around 1% of the macro placement runtime on the largest ORFS cluster tree
+    (nangate45/bp_quad, 220 macros, 98 clusters) and a few KB of ODB. On a
+    design that uses power domains or region groups, the instances already
+    owned by one of those stay in it and are left out of the clustering data
+    (reported once as MPL-0078), since an instance belongs to a single dbGroup -
+    so those cells show up uncolored in the Clusters view. Ignored when
+    RTLMP_ARGS is set, which replaces the whole argument list.
```
AcKoucher: Don't cite other things apart from MPL behavior here, it ends up polluting the docs.

AcKoucher (changes requested):
The name of this PR doesn't tell what it is actually doing. Please, adjust it.

AcKoucher (approved):
@jorge-ferreira-pii Please, fix the PR title, it's still misleading as you're only adding the option here.

Two more things:
1. Avoid adding information that is not related to the changes themselves in the commit message. It's very long and it's not very helpful.
2. Please, don't force-push after a review - we lose the history.

maliberty (approved):

PR #4415, opened 2026-08-12
macro_place_util: apply macro base halos for fixed macros

oharboe:
If macros are already fixed (e.g. by floorplan_anneal), rtl_macro_placer skips placement and never creates blockages. We must explicitly apply the base halo to fixed macros.

### DIFF
--- a/flow/scripts/macro_place_util.tcl
+++ b/flow/scripts/macro_place_util.tcl
@@ -67,6 +67,13 @@ if { [find_macros] != "" } {
   }
 
   log_cmd rtl_macro_placer {*}$all_args
+
+  # If macros are already fixed (e.g. by floorplan_anneal), rtl_macro_placer skips placement
+  # and never creates blockages. We must explicitly apply the base halo to fixed macros.
+  if { [info commands set_macro_base_halo] ne "" && [info commands block_macro_channels] ne "" } {
+    set_macro_base_halo $halo_x $halo_y $halo_x $halo_y
+    block_macro_channels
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty I need some guidance here, our flow is broken without this patch we're carrying. What is it that needs to happen when we're running macro placement and we have placed macros before the macro placement stage runs

Why does macro_place.tcl have to set up these blockages?

maliberty:
@joaomai @AcKoucher can mpl create the blockages even if it is otherwise skipping macro placement?  It seems simpler than pushing it into the flow.

oharboe:
> @joaomai @AcKoucher can mpl create the blockages even if it is otherwise skipping macro placement? It seems simpler than pushing it into the flow.

Thanks!

Please close this PR if this will be fixed in OpenROAD later, we'll just carry the patch for now.

AcKoucher:
> @joaomai @AcKoucher can mpl create the blockages even if it is otherwise skipping macro placement? It seems simpler than pushing it into the flow.

@maliberty It actually can. I implemented `block_macro_channels`, that should work regardless.



@oharboe I expect the problem you're trying to address here to be resolved once we use `block_macro_channels` instead of creating the blockages from within MPL. @joaomai is working on the spec we came up with some months ago to clarify the differences between halos and channels. This should be part of the work there.

oharboe:
:+1: 

Closing as moot: we're using something like this as a stopgap locally.

PR #4420, opened 2026-08-12
bazel: fix tclZipfs.c patch application

oharboe:
The `tclZipfs.c` patch (which fixes the `<crypt.h>` include collision with hermetic glibc) has been upstreamed into the `tcl_lang` Bazel module in the BCR (as of `9.0.2.bcr.2`).

Because the BCR version now carries the patch natively, the `single_version_override` in `MODULE.bazel` that applies it a second time was failing with `CONTENT_DOES_NOT_MATCH_TARGET`. This PR drops the local `single_version_override` patch now that it is redundant.


maliberty:
Why are all the submodule tools being bumped here?

oharboe:
> Why are all the submodule tools being bumped here?

Fixed.

maliberty (approved):

PR #4422, opened 2026-08-13
Updated several gf12 designs to pass LEC

jeffng-or:
Updated gf12 bp_dual, bp_quad, bp_single and coyote to pass a blackbox verilog file instead of the liberty files to LEC. Works around KF issues with the RAM macros.

@joaomai , @eder-matheus I went ahead and updated the bp_* configs so that it should all go through with the OR update + halo updates.


eder-matheus (approved):

maliberty:
@eder-matheus ready to merge?

PR #4429, opened 2026-08-14
global_route: write artifacts once at top-level on success

oharboe:
This refactors global_route.tcl to only write out the .odb and .sdc artifacts once in the successful case, properly keeping the logic at the top-level script similarly to floorplan.tcl. It correctly propagates early failures by returning 0 from the helper proc rather than exiting, which allows OpenROAD to handle failures cleanly without crashing.


oharboe:
never mind. It is fine the way it is. Ideally we could have simpler .tcl scripts and have more logic in C++ where we can have unit-tests for use-cases... Leave be for now.

PR #4455, opened 2026-08-21
flow: collect the global placement HPWL metric

minjukim55:
Register `globalplace__route__wirelength__estimated` in `genRuleFile.py`, with the same padding and comparison as the neighbouring `placeopt__*` rules.

gpl already reports the metric ([#11201](https://github.com/The-OpenROAD-Project/OpenROAD/pull/11201), merged), and `genMetrics.py` already collects it, so this is the only change needed.


maliberty:
merge conflict to resolve

PR #4458, opened 2026-08-21
Add checkQorMetrics.py inline check script

migueldalberto:
## Summary
- Adds `flow/util/checkQorMetrics.py`, a local inline check: sweeps `reports/<platform>/<design>/<variant>/metadata.json`, POSTs each run's numeric metrics to the QoR dashboard's `/compare/check/inline` endpoint, and reports which rules would fail against the latest baseline build (or a pinned `--base-commit`).
- Differs from `checkMetadata.py`: compares against a real baseline build + the dashboard's live `rule_configs` thresholds instead of the committed `rules-<variant>.json` goldens, so metric drift that moves with master doesn't false-positive.
- Always exits 0 — reports, does not gate a local build on network/dashboard availability.

### DIFF
--- a/flow/util/checkQorMetrics.py
+++ b/flow/util/checkQorMetrics.py
@@ -1,0 +1,538 @@
+#!/usr/bin/env python3
+"""
+Checks every design under ./reports against the QoR dashboard's rules
+(https://dashboard.precisioninno.com).
+
+This is the CI inline check, run against a local flow: it sweeps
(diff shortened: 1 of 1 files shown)


maliberty:
@codex review

maliberty:
@codex review

maliberty:
Needs black format

vvbandeira (approved):

PR #4464, opened 2026-08-24
Remove flow/test/test-do-stage.sh

oharboe:
`flow/test/test-do-stage.sh` has no consumers: added in 9b304b70b, it was never wired into CI, the Makefile or the docs, and has bitrotted since — `do-yosys-keep-hierarchy` no longer exists (hierarchy keeping moved to `RUN_YOSYS_ARGS`/`yosys_keep.tcl`), and `do-synth` never comes back (the `do-step` comment in `flow/Makefile` documents that there is deliberately no `do-synth` step).

Removing it rather than repairing a test that nothing runs.


oharboe:
@maliberty False positive on metrics, I deleted an unusef file.

maliberty (approved):

PR #4469, opened 2026-08-25
Removed REMOVE_LIBS_FOR_LEC workaround

jeffng-or:
With the kepler-formal Liberty fixes, the REMOVE_LIBS_FOR_LEC override is no longer needed. Removed from gf12 configs.


jeffng-or:
added trailing newlines

maliberty (approved):

PR #4473, opened 2026-08-25
Export scripts/variables.yaml as a source label

oharboe:
`flow/BUILD`'s `exports_files` already publishes `scripts/synth.tcl` so bazel-orfs can point a rule attribute at `@orfs//flow:scripts/synth.tcl` rather than vendor a drifting copy.

`scripts/variables.yaml` — the variable catalogue carrying each variable's description and stage scoping — has the same need but no label: it is reachable only inside the private makefile filegroups' `data`, so a downstream Bazel target cannot `data`-dep the catalogue at a given commit.

This adds it to the same `exports_files`. Source-visibility only; no flow behaviour changes.

### DIFF
--- a/flow/BUILD
+++ b/flow/BUILD
@@ -24,7 +24,14 @@ exports_files(
 # instead of vendoring its own (drifting) copy at the bazel-orfs repo
 # root. See bazel-orfs `private/rules.bzl`.
 exports_files(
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty This is the variables.yaml counterpart to already exported variables.json

oharboe:
moot

PR #4475, opened 2026-08-26
feat(variables): FLOW_INPUT_VARIANT, read inputs from another variant

oharboe:
FLOW_VARIANT names the directory a run writes to. It also names the directory a run reads from, which means a variant that wants to reuse an upstream variant's results has to physically copy them into its own directory first. Sharing an expensive front end across experiments — one synthesis, one floorplan, many routing variants — is a `cp -r` per fork today.

Split the two. FLOW_INPUT_VARIANT names the variant a run reads from and gives INPUT_RESULTS_DIR, INPUT_LOG_DIR, INPUT_REPORTS_DIR and INPUT_OBJECTS_DIR alongside the existing output directories. It defaults to FLOW_VARIANT, so every INPUT_*_DIR expands to a string identical to its *_DIR counterpart and an unchanged run is unchanged.

This is the ORFS side of The-OpenROAD-Project/bazel-orfs#846, extracted as a single concern patch.


oharboe:
@maliberty coralnpu failure?

maliberty:
This looks rather error prone given the number of places you had to touch.  Would it be easier to just have a copy (or symlink) target?

oharboe:
> This looks rather error prone given the number of places you had to touch. Would it be easier to just have a copy (or symlink) target?

That creates an iffy stateful mess in bazel-orfs I am trying to get rid of.

oharboe:
> This looks rather error prone given the number of places you had to touch. Would it be easier to just have a copy (or symlink) target?

We've been bitten in the past by the code to infer which .sdc file to load based on the .odb file. Why the dichotomy of .odb and .sdc? There are a million bits and bobs in .odb for all sorts of modules, what makes .sdc special? Perhaps .odb could subsume the .sdc like it does so many other things. The only thing we can do with the freedom we have today is to load the wrong .sdc data.

oharboe:
@maliberty If we [dealt with the .odb / .sdc dichotomy](https://github.com/The-OpenROAD-Project/OpenROAD/pull/11260) the flow code would be much simpler, there would be no guessing as to which .sdc goes with an .odb file.

oharboe:
@maliberty I think the CoralNPU is a snag on origin/master that will right itself on the next pr-merge. Surely lots of people are feeling the pain.

maliberty:
Parsing SDC can be slow and something like GUI_TIMING=0 doesn't need it.  You would have to have command options to skip it which is more complication.  

This is also iffy, you just consolidate the `if`s to orfs_input_path  but everyone has to remember to use it.

oharboe:
@maliberty you prefer to roll .sdc into .odb first?

oharboe:
closing until we have discussed upcoming roll .sdc into .odb PR.

PR #4477, opened 2026-08-26
variables: FLOW_INPUT_VARIANT, read inputs from another variant

oharboe:
`FLOW_VARIANT` names the directory a run writes to. It also names the directory a run *reads* from, so a variant that wants to reuse an upstream variant's results has to physically copy them into its own directory first. Sharing an expensive front end across experiments — one synthesis, one floorplan, many routing variants — is a `cp -r` per fork today.

## The change

`FLOW_INPUT_VARIANT` names the variant a run reads from, and gives `INPUT_RESULTS_DIR`, `INPUT_LOG_DIR`, `INPUT_REPORTS_DIR` and `INPUT_OBJECTS_DIR` alongside the existing output directories:

```make
export FLOW_INPUT_VARIANT ?= $(FLOW_VARIANT)
export INPUT_RESULTS_DIR  ?= $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKNAME)/$(FLOW_INPUT_VARIANT)
```

It defaults to `FLOW_VARIANT`, so every `INPUT_*_DIR` expands to a string **identical** to its `*_DIR` counterpart and an unchanged run is unchanged.

```shell
make DESIGN_CONFIG=... FLOW_VARIANT=base cts
make DESIGN_CONFIG=... FLOW_INPUT_VARIANT=base FLOW_VARIANT=high_effort do-grt do-route do-final
```

The same mechanism lets one fork produce an input for another — a pin-placement flow running as `FLOW_INPUT_VARIANT=base FLOW_VARIANT=io` reads the shared floorplan, writes `io.tcl` into the `io` variant, and the main flow re-runs from floorplan with `IO_CONSTRAINTS` pointing at it, still reusing the same `base` synthesis. `docs/user/FlowVariables.md` documents this with a diagram.

## Search path, not a switch

Reads resolve against `RESULTS_DIR` first and `INPUT_RESULTS_DIR` second. That ordering is required, not cosmetic: a stage runs several steps in one process — `do-place` writes `3_1_place_gp_skip_io.odb` and the next step reads it back — so a file this run just wrote must win over the upstream variant's copy. `orfs_input_path` / `orfs_input_glob` in `util.tcl` are the whole mechanism; every write still goes straight to `RESULTS_DIR`.

## What is touched

`load_design` is the chokepoint — all 18 stage scripts hand it a bare basename, so routing it through `orfs_input_path` covers the flow. The remaining reads are the `AUTO_MEMORIES` globs, `find_sdc_file`, the abstract's `.spef`, `yosys_load`'s netlist, and genMetrics' log/report/result directories.

`find_sdc_file`'s candidate ordering now compares basenames rather than full paths. With one directory the two are equivalent, since every candidate shares a prefix; with two, the directory prefix would otherwise decide which `.sdc` wins.

The Makefile's prerequisite graph is deliberately untouched: it still names `$(RESULTS_DIR)`. A forked variant is driven through the `do-` targets. Teaching the prerequisites to search a second directory is a separate, `vpath`-shaped change.

## Testing

Exercised end to end through bazel-orfs, which is the heaviest user of cross-variant reads — it runs every stage of every design out of one variant's directory into another's, and until now did so by `mv`-ing inputs before each `make`. With this patch that renaming is deleted outright and the whole suite (123 tests) passes, so every test in it now runs through this code path.

Added there alongside it: a variant forked at cts off a shared `place` stage, asserting its config names the input variant, that its action inputs carry the upstream variant's ODB and no copy under its own results path, and that it builds. All three fail without `FLOW_INPUT_VARIANT`.

### DIFF
--- a/docs/user/FlowVariables.md
+++ b/docs/user/FlowVariables.md
@@ -84,6 +84,70 @@ These are optional variables that may be over-ridden/appended with
 default value from the platform `config.mk` by defining in the design
 configuration file.
 
+## Variant directories: reading from one variant, writing to another
+
+`FLOW_VARIANT` names the directory a run writes to: `RESULTS_DIR`,
+`LOG_DIR`, `REPORTS_DIR` and `OBJECTS_DIR` all end in the variant name.
+`FLOW_INPUT_VARIANT` names the directory a run *reads* from, through the
+matching `INPUT_RESULTS_DIR`, `INPUT_LOG_DIR`, `INPUT_REPORTS_DIR` and
+`INPUT_OBJECTS_DIR`. It defaults to `FLOW_VARIANT`, so by default the
+two are the same directory and nothing about a normal run changes.
+
+Setting it forks a variant off a shared upstream one. The forked run
+reads the upstream variant's results and writes only its own, so the
+shared stages are never re-run and never copied. Results are looked up
+in `RESULTS_DIR` first and `INPUT_RESULTS_DIR` second, so a file an
+earlier step of the same run just wrote always wins over the upstream
+variant's copy of it.
+
+This is how to share an expensive front end across experiments — run
+synthesis, floorplan, placement and CTS once as `base`, then fork a
+variant per routing experiment:
+
+```shell
+make DESIGN_CONFIG=... FLOW_VARIANT=base cts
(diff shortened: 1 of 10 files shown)


oharboe:
duplicate

PR #4486, opened 2026-08-28
add warn and error for placement density lower bound

AcKoucher:


flow/scripts/util.tcl
```diff
@@ -187,6 +187,18 @@ proc place_density_with_lb_addon { } {
...
       -pad_left $::env(CELL_PAD_IN_SITES_GLOBAL_PLACEMENT) \
       -pad_right $::env(CELL_PAD_IN_SITES_GLOBAL_PLACEMENT)]
+
+    if { $place_density_lb >= 1.0 } {
+      utl::error FLW 25 \
+        "Placement density lower bound is >= 1.00 ([format %.3f $place_density_lb]).\
+         The design does not fit the core. Reduce CORE_UTILIZATION or\
+         CELL_PAD_IN_SITES_GLOBAL_PLACEMENT."
+    } elseif { $place_density_lb > 0.80 } {
+      utl::warn FLW 26 \
```
oharboe, line 197: 1. why in ORFS and not OpenROAD?
2. OpenROAD allocates error numbers. We did a sweep to get rid of error numbers in ORFS, why is it being re-introduced here? (Ask AI to check history...)
3. CORE_UTILIZATION is [0, 100] and PLACE_DENSITY [0, 1] ranges. I don't think there's any reason it ended up this way, but it is not consistent and it is easy to confound these units.

AcKoucher: 1. I suppose we could put this in GPL, but then we cannot refer to the flow variables e.g., the error message would look like `The cells do not fit in the core. Reduce core utilization or cell padding and try again.`. Then an ORFS user would have to figure out the flow variables. What do you say?
2. I see that we don't have many `FLW` messages, but I couldn't find such sweep.
3. I don't know how it got like this and I agree with you, but this is outside the scope of this PR.

AcKoucher:
@oharboe Could you take a look if you agree with what it's being proposed in #4499?

PR #4487, opened 2026-08-31
AUTO_FLOORPLAN: do not review, work in progress

oharboe:
**Draft — work in progress, campaign still running. Filed early so the state is shareable.**

Retires the last hand-carried numbers in a `config.mk` — `CORE_UTILIZATION`,
`DIE_AREA`/`CORE_AREA`, `CORE_ASPECT_RATIO`, `PLACE_DENSITY` and
`PLACE_DENSITY_LB_ADDON` — by measuring them instead of reading them.
`AUTO_FLOORPLAN=1` races candidate outlines and densities at the floorplan
stage, scores each by running the real flow elements against it, and overrides
whatever `config.mk` asked for. The `config.mk` values are untouched: they are
the `AUTO_FLOORPLAN=0` path and they enter the race as the incumbent.

This implements `ideas/auto-floorplan.md` from bazel-orfs, and is the
multi-design counterpart to that campaign's Phase B (the utilization shmoo).

## Self-contained by construction

bzlmod honours `*_override` only from the root module, so bazel-orfs's carried
OpenROAD patches were being **silently discarded** here — the resolved
`@openroad` marker listed no patch files at all, and anyone checking this branch
out got an unpatched OpenROAD with no indication. Replaced
`local_path_override(tools/OpenROAD)` with a `git_override` pinned to
`3ca581e384` carrying the five patches inline, so a fresh clone builds the
toolchain this was measured on:

| patch | upstream | why it is needed |
|---|---|---|
| `0045` gpl parallel density scatter | OpenROAD#11262 | the scorer *is* a report-only global placement; also the `mt_invariance` test that makes scores CPU-count-invariant |
| `0041` `rtl_macro_placer -random_seed` | feature | deterministic macro placement per outline, or the race measures placer noise |
| `0040` mpl re-run segfault | OpenROAD#11277 | unblocks the in-process fast path |
| `0042` snap-inside-core | OpenROAD#11278 | winners round-trip through `place_macro` |
| `0043` all-fixed std-cell seeding | OpenROAD#11279 | removes ~2% injection-vs-production divergence |

Also carries one bazel-orfs patch (`user_stages` leaking into `orfs_update`,
which breaks analysis for every design with a `rules-base.json`) and two
ORFS-side fixes as ordinary commits: the slang plugin load, and
`swerv_wrapper`'s `$(CLKGATE_MAP_FILE)`, which the config.mk parser emitted as
a *literal unresolved string* into `VERILOG_FILES`.

## The design, and the discipline

Three coordinates on dimensionless ladders — no per-design constant anywhere in
the implementation.

- **Density**: the measured lower bound stays; fractions of the headroom above
  it are raced. That fraction *is* `PLACE_DENSITY_LB_ADDON`, now measured. The
  winner also becomes RTL-MP's `-target_util`, fixing both consumers.
- **Utilization**: constraint satisfaction, never minimisation. A bigger core
  always makes period easier, so an objective blending the two is a leak, not a
  trade-off. Take the smallest core no worse than the incumbent; area is a
  budget, period is the goal axis.
- **Aspect ratio**: raced at the winning area point; often returns a tie, which
  is a result — the folklore value gets graded rather than assumed.

Candidates run as subprocesses that source the *production* stage scripts, so
what is scored is what would be built, and the parent ODB is untouched (only the
winner commits). `delta_tie` is measured per design as 2σ over 8 placer seeds and
every verdict is published next to it.

Against overfitting a cheap proxy: feasibility guards are hard filters, not score
terms; the objective is an aggregate, never WNS; ties break toward headroom; and
a ladder whose spread does not exceed its own noise floor has resolved nothing,
so the incumbent is kept rather than defaulting to the smallest core.

## Three rungs, each useful alone

| | |
|---|---|
| **Explore** | `AUTO_FLOORPLAN=1` (default). Measured every run, never stale. |
| **Pin** | `bazelisk run //…:<name>_auto_floorplan_pin` writes the winner into `config.mk`, sets `AUTO_FLOORPLAN=0`. A measurement becomes a decision. |
| **Overfit** | Pinned values are ordinary `config.mk` entries, so AutoTuner and seed sweeps apply as ever — the frozen-RTL endgame. |

The bar is **parity with hand-tuning, reached automatically**, not beating it.
Hand-tuned numbers are one person's sweep of one design, frozen the moment the
RTL moves underneath them.

## `<name>_pareto_test` — did the front actually move?

`checkMetadata.py` asks whether anything got worse than a padded bound, and
cannot express a *trade*: a change that shrinks the core 8% and gives up 1% of
slack fails the slack rule and says nothing about the area rule it improved.

`genRuleFile.py` now records `golden` — the unpadded measurement — beside each
threshold (purely additive; `checkMetadata.py` and `updateRules.py` ignore it).
`checkPareto.py` reads those and fails only if the point is **dominated**, while
treating a hard-constraint regression or a **lost timing closure** as failures
outright. 11 unit tests.

## Results so far — read this before the numbers

**The scorer cannot protect timing closure, and this is measured, not
suspected.** It scores after global route but before CTS and `repair_timing`, so
its absolute period is not merely noisy but heavily biased: on `ibex` the proxy
reads `wns = -1739 ps` where the finished flow lands at `+13.8 ps` — a bias of
1753 ps, **175% of the clock period**. A guard conditioned on "the proxy says we
meet timing" can never fire, because the proxy says no design meets timing. The
obvious guard is therefore not implementable from this measurement, and the
reason is documented in the code rather than left as an absence.

Ranking accuracy *at a fixed area* is a weaker property that may well hold —
that is what the density and aspect phases rest on. It does not carry to the
area-versus-period trade, which needs to know how much margin exists.

Observed consequence: `ibex` went from `WS +13.8 ps` to `-2.9 ps` while its core
shrank 23%. A change in kind, not a Pareto cost. This is why the feature is
documented as design-space exploration rather than sign-off, why the pin target
exists, and why `checkPareto.py` treats lost closure as a hard failure.

The measured noise floors are large — 22–32% of the clock period on the designs
swept, against ~1.1% for a 36-leaf `stage_variance` walk in bazel-orfs#868 — and
that bounds what any rule reading this proxy can resolve. Where the ladder spread
falls below that floor (`aes`, `ethmac`), the race correctly reports "did not
resolve" and keeps the `config.mk` values.

An asap7 A/B across both arms is running; the per-design table lands here when it
completes. Known gap: `riscv32i`, `riscv32i-mock-sram` and `aes-block` fail in
*parallel synthesis* (`per-module checkpoint missing:
partition_*_canonical.rtlil`) after the bazel-orfs bump — before the floorplan
stage, so unrelated to this feature, but not yet fixed.

## Relation to bazel-orfs#868

That PR's E1 found the period axis **saturated** at swerv's 30% utilization: all
24 candidates tie on achieved period across a 45% proxy spread, so selection
cannot buy period there, and it hands the decisive question to the dense regime
and the utilization shmoo — i.e. here. Its E3 found raw gpl HPWL
ranking-equivalent to the full-STA proxy (ρ +0.67 vs +0.72), which suggests the
scorer in this PR is over-built; that is cheap to test against the evidence
files already produced and is not yet done.

`af_ladder_resolves` — spread vs `delta_tie` — is a crude version of #868 Phase
B's axis-liveness instrument, evaluated across many designs rather than one.

## Open questions

- Is the scorer over-built? #868 E3 says HPWL alone may rank equivalently.
- Should the guards be this conservative? Under a merge-then-verify workflow the
  better posture is to let the race move and let `<name>_pareto_test` on trunk be
  the revert trigger.
- Ranking accuracy in the dense regime remains the load-bearing unknown.

---

## Status and next steps

Campaign state as of the last update to this description. ETAs are from
that point; "cheap" means no flow runs needed.

| # | Next step | Why | Cost | ETA |
|---|---|---|---|---|
| 1 | ~~asap7 A/B sweep, both arms~~ | **done** — 15 designs, both arms; see the results comment | — | complete |
| 2 | Fix parallel-synth breakage from the bazel-orfs bump | `riscv32i`, `riscv32i-mock-sram`, `aes-block` never reach the floorplan stage, so 3 designs are missing from the table | unknown, diagnosed not fixed | next session |
| 3 | Confirm `<name>_pareto_test` targets analyse | written and unit-tested, but bazel could not be run while the sweep held the lock | cheap | next session |
| 4 | Regenerate `rules-base.json` goldens | without `golden` the Pareto test warns and passes vacuously | 1 flow run per design (shares the sweep's) | next session |
| 4b | **Implement the flow-in-the-loop utilization derivation** | the proxy does not rank this axis (rho -1.00 gcd, +0.10 gcd-ccs); the flow does. ~6 production runs per design, picked by J, pinned | ~6 runs/design | next |
| 5 | **Test whether the scorer is over-built** | bazel-orfs#868 E3: raw gpl HPWL ranks equivalently to full STA (rho +0.67 vs +0.72). This PR's candidate runs STA *and* `repair_design` *and* `global_route` | cheap — correlate rankings in evidence files already produced | next session |
| 6 | Extract the axis-liveness table | spread/`delta_tie` vs utilization across all swept designs; the multi-design counterpart to #868 Phase B, and it explains which designs the race can help at all | cheap — data already in the evidence files | next session |
| 7 | Reconsider guard posture for trunk-based flow | under merge-then-verify the right posture is to let the race move and let `<name>_pareto_test` be the revert trigger, rather than refusing to move pre-merge | design change, needs a call | after 1 and 6 |
| 8 | `delta_tie` methodology | 2 sigma over 8 placer seeds measures *placement sensitivity*, which is not the same quantity as #868's `stage_variance` flow reproducibility, and is ~20x larger | needs design | open |

### Known gaps in the data

- `riscv32i`, `riscv32i-mock-sram`, `aes-block`: parallel-synthesis failure
  from the bump (item 2), not from this feature. Every tier consequently
  exits non-zero, so exit codes are not a health signal for this campaign —
  the per-design captures are.
- `coralnpu`: not yet captured.
- `delta_tie` is 8 seeds, not a `stage_variance`-grade walk (item 8).

### What would falsify the approach

If the axis-liveness table (item 6) shows the period axis is dead at every
operating point these designs run at, then racing utilization cannot buy
period anywhere in this design set, and the honest outcome is to ship the
density and aspect coordinates only — where fixed-area ranking is the
property that has to hold — and to record the utilization result as the
negative finding it is.

---

## Update — overnight campaign complete

The utilization axis is **off by default**: measured `rho = -1.000` on gcd
and `+0.100` on gcd-ccs against the flow, so the scorer does not rank it.
Density and aspect continue to race at fixed area, which is the ranking
property that has been measured to hold. Full results, including the
objective that replaced the threshold rule, are in the results comment
on this PR.


oharboe:
## Overnight campaign: the utilization axis does not work with this scorer

Two experiments ran overnight. The headline is a **negative result on one
of the three coordinates**, located precisely enough to act on.

### 1. Does the proxy rank the utilization axis?

The production flow was run at every rung of the utilization ladder and
correlated against the proxy score for the same rungs (already recorded
in each design's evidence file):

| design | ρ | n | `delta_tie`/clock |
|---|---:|---:|---:|
| **gcd** | **−1.000** | 4 | 1.1% |
| gcd-ccs | +0.100 | 5 | 0.3% |
| ibex | +0.429 | 6 | 31.8% |

gcd and gcd-ccs are the only designs in the set whose noise floor is
small enough for an answer to exist. On gcd the proxy is **perfectly
inverted**: it scores 338 → 335 (improving) as utilization rises, while
the flow goes 349.2 → 355.1 ps (degrading). ibex's +0.429 is unusable —
its floor is a third of the clock.

*(`uart` was also run and is excluded: the experiment script rewrote
`CORE_UTILIZATION`, which `uart` does not use — it sets
`DIE_AREA`/`CORE_AREA` — so all six "rungs" ran the identical floorplan.
An experiment bug, not a flat axis.)*

### 2. Re-sweep with the objective, driven by the proxy

15 designs, same AF=0 baseline: **2 wins, 7 trades, 3 no-change, 3
regressions** — including `aes` picking a core **11.2% larger** with a
worse period. Zero DRC anywhere.

### The objective is fine; the instrument is not

Given the **flow's** numbers, `J` picks well on every design measured:

| design | J picks | core | achieved period | DRC |
|---|---|---|---|---|
| gcd | util 78 | 61.3 → 50.0 (−18.4%) | 348.3 → 355.1 (+2.0%) | 0 |
| gcd-ccs | util 78 | 66.0 → 50.0 (−24.2%) | 354.3 → 355.8 (+0.4%) | 0 |
| ibex | util 56 | 5283 → 3766 (−28.7%) | 986.2 → 995.8 (+1.0%) | 0 |

Those are real Pareto gains. The fault is the scorer, and only on this
axis: density and aspect are compared at **fixed area**, which is the
ranking property #868's E1/E3 measured to hold (ρ +0.67…+0.72).
Utilization changes the area, and across that change the proxy does not
rank.

### Re-scope

`ideas/auto-floorplan.md` always said the oracle here was the flow — *"the
race is the oracle inside a utilization shmoo"* — and using a pre-route
proxy for the outer loop was the shortcut that failed.

- `AUTO_FLOORPLAN=1` now races **density and aspect at the incumbent
  utilization**. `AF_RACE_UTIL=1` re-enables the axis for reproduction.
- The correct utilization derivation is flow-in-the-loop: ~6 production
  runs per design (4 min on gcd, 23 on ibex), picked by `J`, pinned with
  `<name>_auto_floorplan_pin`. An overnight artifact, not a per-build
  step. Not yet implemented.

### Also landed

- **Selection is now a smooth objective** rather than a threshold:
  `J = ln(area) + λ·ln(T + τ·ln(1+e^((p−T)/τ)))`, λ=3, τ=2% of clock.
  λ is an exchange rate — one percent of period is worth three percent of
  area. Verdicts are unchanged for λ ∈ [2,5] and τ ∈ [1%,10%], and the
  smooth form agrees with a hard `max(p,T)` on all 15 designs, so the
  smoothing costs nothing and removes the cliff. Two behaviours fall out
  rather than being coded: a period loss that buys no area can never
  lower `J`, and the incumbent is just another candidate.
- **The period axis is `clock − WNS`, never WNS.** WNS lives near zero, so
  fractional changes in it are meaningless and a tie band expressed as a
  fraction of it is backwards. This re-reading turned "ibex
  catastrophically lost closure" into "+1.69% period", and separated the
  designs that traded ~1% of period for ~25% of core from those that gave
  up a lot. TNS demoted to a diagnostic for the same reason.
- **`delta_tie` is 2σ over 8 seeds**, not the range of 3 — the range is
  biased low and its expectation grows with n, so guards would have
  shifted purely by changing the seed count.

### Open

- Is the scorer over-built? #868 E3 says raw gpl HPWL ranks equivalently
  to full STA. This PR's candidate runs STA *and* `repair_design` *and*
  `global_route`. Cheap to test against evidence already produced.
- `riscv32i`, `riscv32i-mock-sram`, `aes-block`, `coralnpu` still fail in
  parallel synthesis after the bazel-orfs bump — before the floorplan
  stage, so unrelated to this feature, but 4 designs missing from every
  table.

oharboe:
Superseded by **https://github.com/The-OpenROAD-Project/bazel-orfs/pull/890**.

The config.mk BUILD DSL moved to bazel-orfs
(`2d5b345 feat(design): own the config.mk BUILD DSL here, not in ORFS`),
and bazel-orfs now patches ORFS's `flow/designs/design.bzl` down to a
12-line re-export. This PR's derivation targets were generated from that
file, so in the bazel-orfs workflow they would have been patched away. The
move was forced rather than preferred — and it is an improvement: the
targets are now generated by `design()`, so they exist for every `@orfs`
design package, including the ~100 whose BUILD files bazel-orfs generates.

What carried over: the driver, the candidate, the flow tail and its drift
guard, the config.mk pin, `check_pareto.py`, and the full write-up.

What was dropped as unnecessary: `genRuleFile.py`'s `golden`/`tie` fields
and all 20 regenerated `rules-base.json` files. `check_pareto --baseline`
takes its reference from a freshly measured `metadata.json` instead, which
is strictly better — see 0c below. The ORFS-side diff is now one line, a
`flow.tcl` export carried as bazel-orfs `patches/0047`.

**This PR filed nothing upstream, and that was the right outcome.** An
attempt to land one design as a single-concern PR was stood down when the
"clearest win" turned out to be a design that lost timing closure. Four
defects came out of it:

| | | status |
|---|---|---|
| 0a | the incumbent ran at a density the design never states | fixed and verified |
| 0b | a 9 ps oracle-vs-flow gap | not established — inside the noise |
| 0c | checked-in `rules-base.json` lags the toolchain | not a bug; constrains method |
| 0d | the objective accepts a lost timing closure | **open** |

0c is worth flagging to ORFS independently of any of this: `rules-base.json`
files in `master` were last regenerated between one and seven weeks ago,
and on `aes_lvt` the checked-in bound implies `ws = 0.0` where a fresh run
on the same commit measures `+9.00`. Nothing is malfunctioning — the guard
is one-sided and a design getting faster is not a regression — but a
`rules-base.json` diff cannot be used as evidence for a change without
separating that drift first, and back-solving a baseline out of a padded
bound will invert verdicts. It did here.

The branch stays on the fork; the measurements cited in
`docs/plans/auto_floorplan_ideas.md` were produced here and this remains
their record.

PR #4492, opened 2026-09-01
flow: E12 clustered global-place scorer — measured; the readout is the limit, not the clustering

oharboe:
> **Superseded / for reference.** The findings here are being converged into
> [bazel-orfs#881](https://github.com/The-OpenROAD-Project/bazel-orfs/pull/881).
> This branch remains as the reproducible apparatus and evidence.

## What this set out to test

E12 of the macro-placement estimation ladder: can a fast global placement run on RTL-MP's own std-cell **clusters** — an order of magnitude fewer movables — and still rank macro-placement candidates? Ranking accuracy was the only requirement.

It ran. The answer is more useful than a pass or a fail.

## Result: the clustering was never the problem — the scalar is

Graded against ground truth measured **on this setup** (24 candidates generated, 23 evaluable, truth = the full production tail to grt, ~34 min/candidate, ~1 h wall for all 23 at 4-way parallel):

| scorer | ρ vs grt `macro_paths_mean` | |
|---|---|---|
| flat, **STA** aggregate (`wq25`) | **+0.61 [+0.29, +0.78]** | ranks |
| flat, STA macro aggregate | +0.59 [+0.20, +0.80] | ranks |
| flat, raw HPWL | +0.24 [−0.24, +0.65] | no evidence |
| clustered HPWL | +0.26 [−0.22, +0.66] | no evidence |
| clustered **macro-cone** HPWL | +0.39 [−0.01, +0.72] | just misses |

**1. The campaign's central idea reproduces.** A fast non-timing-driven placement with an STA readout ranks the macro-path KPI at **+0.61**, interval clear of zero, against the published +0.72 [+0.48, +0.84]. Scoring on the far side of the fog works.

**2. "Drop the STA" (E3) does not reproduce.** The campaign reports raw HPWL ranking the same truth at +0.67. Here it is **+0.24 with the interval spanning zero** — no demonstrable signal. This is load-bearing for the whole cheap-scorer programme and needs re-deriving before anything is built on it.

**3. E12's clustering is faithful and cheap; its readout is the limit.** Clustered HPWL (+0.26) is statistically indistinguishable from flat HPWL (+0.24), while tracking flat HPWL closely (**+0.66 [+0.34, +0.86]**) at **14.7 s vs 63.0 s** and **0.87 GB vs 2.43 GB**. So the cluster model reproduces what it was asked to reproduce — and inherits a scalar that carries no signal. No improvement to the clustering can fix that. **E12's question changes from "compute this number cheaper" to "which number should we compute".**

The constructive thread is the **macro-cone** readout at +0.39 — best of the HPWL family, only just short of significance, and the one variant that moved toward the live KPI, by restricting the sum to nets within reach of a macro pin.

## Two findings that matter beyond E12

**A candidate can be infeasible, and no wirelength score can say so.** cand_s2's macro placement leaves a channel PDN cannot repair (`PDN-0179`). A selector shipping it would have produced a design that does not build. "Fraction of candidates infeasible" looks like an untracked KPI.

**The archived reference describes a different population.** Settled from both directions: our scores vs archived truth ρ = −0.42; **our truth vs archived truth ρ = −0.26 [−0.65, +0.24]** on `macro_paths_mean` and +0.07 on `achieved` — uncorrelated. Seed N here is a different placement than seed N there. ORFS and OpenROAD are ruled out (identical to the campaign's `be0dca0b` and `3ca581e38` pins, the latter with zero source diffs and the same `src/sta`); what remains is the campaign's much newer bazel-orfs. The archive stays internally consistent and still reproduces its published +0.72 — it just cannot anchor these candidates.

## Chesterton's fence: why the divergence was not a bug

The clustered solve diverges at the pre-registered default on swerv — overflow floors at 0.92 for 2175 iterations while the density penalty climbs to 4.1e+28, then GPL-0305. That is a **granularity mismatch, not a defect**:

- `BinGrid::initBins()` sizes bins from average **dbInst** area — correct for the regime `placement_cluster` was built for. Its commit describes "a small group of gates" whose members are "all placed at the center of the cluster", and notes small clusters give better results.
- RTL-MP's leaf clustering deliberately goes the other way: `min_num_macros_for_multilevel = 150` forces `max_level = 1` here, and `base_min_std_cell` targets 10183–50915 cells per cluster, so 7 clusters hold 98.6% of instances.

Both tools behave as their authors intended; E12 straddles the seam. tinyRocket converges at the default with 13 clusters, so it is the granularity that diverges, not clustering as such. The 24 swerv runs therefore use `E12_BIN_GRID_COUNT=8`, **disclosed as a knob** and reported as exploratory rather than as a gate result.

## Infrastructure fixes (independent of E12)

Running one experiment surfaced nine pre-existing bazel breakages; six are fixed here and would block anyone else the same way:

| fix | what it unblocked |
|---|---|
| drop empty `patches/tcl_lang` | `//...` could not be **parsed at all** |
| bump bazel-orfs past `4254fc9` | **every** `SYNTH_USE_SYN=1` design failed synth (`sv_elaborate: no input files`) |
| swerv clkgate as a resolved path | `unknown module 'OPENROAD_CLKGATE'` — any make-computed path in `VERILOG_FILES` is silently dropped under bazel |
| bazel-orfs `split()` patch | any `SYNTH_HIERARCHICAL` + keep-modules design died at analysis |
| pin tinyRocket keep-modules | bazel needs module names at *analysis* time; all 32 partitions failed without them |
| design-target visibility alias | `//flow/test/macro_e12` could not consume a design's synth |

Left alone as pre-existing and not mine to redesign: `gcd_single_flow` → missing `gcd_yosys_synth`, `mempool_group_synth` visibility, `nangate45/gcd` grt, and `nangate45/swerv_wrapper`'s ABC SIGSEGV under bazel's yosys 0.64.

The bump was verified **differentially** — `bazelisk build --nobuild --keep_going //...` yields a byte-identical failure set before and after, so 68 commits of bazel-orfs introduce no new analysis breakage.

## Apparatus

`flow/test/macro_e12/`, a design ladder × four stages:

- **`macro_e12.tcl`** — seeded population + flat control rung
- **`dump_clusters.tcl`** — RTL-MP's partition, once, from the shared base floorplan
- **`score_clustered.tcl`** — the clustered rung, one candidate per process
- **`evaluate.tcl`** — the production tail to grt: the truth
- **`grade_e12.py`** — ρ + bootstrap CI, cost table, verdict
- **`run_parallel.sh`** — N-at-a-time driver

Two designs: `nangate45/tinyRocket` (mechanics: 2 macros, ~6 min tail, whole chain in ~30 min) and `asap7/swerv_wrapper` (the real 28-macro population).

**The gate can now say "inconclusive".** On the 4-candidate smoke run it reported `PASS — rho = +0.80 [-1.00, +1.00]`: an interval spanning the whole range, where any ρ lands "inside" by accident. Gate 1 now requires ρ in the interval, the population complete, **and the interval clear of zero** — the campaign's own E1 criterion, which I had failed to implement.

Flushing on the cheap design paid for itself: it found unplaced-IO (`GPL-0326`), a **negative** HPWL (inverted `getTermBBox` on unplaced terminals), and a cluster file that did not round-trip (escaped Verilog names read through Tcl *list* parsing turned `\[` into `[`, missing 3751 of 29069 instances). Three of those four produce **plausible-looking numbers** rather than crashes, which is what a scoring experiment cannot survive.


oharboe:
Converging into [bazel-orfs#881](https://github.com/The-OpenROAD-Project/bazel-orfs/pull/881), which now carries:

- `ideas/e12-clustered-scorer.md` — the experiment's record and verdict
- a **Negative findings of importance** section in `docs/estimate.md` — the six results that invalidate assumptions a later experiment would otherwise build on

**Closing, not abandoning.** The headline: the clustering worked and the readout did not. The clustered rung reproduced the flat rung's HPWL at ρ +0.66 [+0.34, +0.86] for 14.7 s against 63.0 s, and ranked ground truth at +0.26 — indistinguishable from flat HPWL's +0.24. So E12's question becomes *which* number to compute, not how to compute it more cheaply. And because `placement_cluster` places every member at its cluster centre, a clustered placement cannot carry an STA readout at all, which leaves the macro-cone variant (+0.39 [−0.01, +0.72]) as the one live thread.

The branch stays pushed as reproducible apparatus and per-candidate evidence.

### Still unfiled: six fixes here are independent of E12

These block anyone using ORFS through bazel today and are not experiment-specific. They deserve their own PRs rather than dying with this one:

| commit | what it unblocks |
|---|---|
| `3ad77a0e6` | `patches/tcl_lang` globs `*.patch` in a directory with none — `//...` cannot be **parsed at all** |
| `e2143f019` | bazel-orfs pin predates OpenROAD-SYN support — **every** `SYNTH_USE_SYN=1` design fails synth with `sv_elaborate: no input files` |
| `ec114b5af` | `$(CLKGATE_MAP_FILE)` is silently dropped when `VERILOG_FILES` is turned into labels → `unknown module 'OPENROAD_CLKGATE'`. General trap for any make-computed path |
| `26015ca6a` (split) | bazel-orfs `split()` called without a separator — a Python-ism Starlark rejects, so any `SYNTH_HIERARCHICAL` + `SYNTH_KEEP_MODULES` design dies at analysis. Belongs upstream in bazel-orfs |
| `26015ca6a` (split) | tinyRocket keep-modules unpinned — bazel needs module names at *analysis* time to declare per-module actions, so all 32 partitions fail |
| `9394ac38d` (part) | design flow targets are package-private with no visibility passthrough |

The bazel-orfs bump was verified differentially: `bazelisk build --nobuild --keep_going //...` yields a byte-identical failure set before and after, so 68 commits introduce no new analysis breakage.

Left alone as pre-existing: `gcd_single_flow` → missing `gcd_yosys_synth`, `mempool_group_synth` visibility, `nangate45/gcd` grt, and `nangate45/swerv_wrapper`'s ABC SIGSEGV under bazel's yosys 0.64.

maliberty:
E12, E3, etc make this unreadable as they are inexplicable and must come from some other context

oharboe:
> E12, E3, etc make this unreadable as they are inexplicable and must come from some other context

Agreed. Note that you found this PR in the bin, I immediately closed it as inactionable.

PR #4501, opened 2026-09-02
bazel: stop carrying build files bazel-orfs can generate

oharboe:
TL;DR bazel-orfs has the machinery to build any ORFS from within the bazel-orfs repository now, so remove hundreds of files of cruft in ORFS. Less noise and churn by having that machinery in bazel-orfs.

ORFS is a pinned storage of scripts, PDKs and example designs as far as bazel-orfs is concerned.

https://github.com/The-OpenROAD-Project/bazel-orfs#quick-start

This also has the effect that you can build any ORFS design from within OpenROAD or bazel-orfs:

```bash
bazelisk run @orfs//flow/designs/asap7/gcd:gcd_final gui_final
```

I haven't bumped OpenROAD in a while to new bazel-orfs, there hasn't been a need or call for it as bazel-orfs in OpenROAD is only used for regression testing and presumably it is doing its job fine with the version that it has. Even post bumping bazel-orfs in OpenROAD, nothing will change in ORFS w.r.t. regression testing in CI, building any design like above is just a new feature that I don't think anyone would use over bazelisk run //:install + make from ORFS, in practice.


oharboe:
out of sight out of mind for now

maliberty (approved):

PR #4517, opened 2026-09-05
designs: mempool_group/rtl BUILD uses files() so its sources export individually

oharboe:
`nangate45/mempool_group`'s `VERILOG_FILES` names the files under `src/mempool_group/rtl` one by one. bazel-orfs's config.mk parser turns each into a per-file label, and a bare `filegroup` over a glob makes those files private to the package, so the design fails analysis under bazel with a visibility error on every one of them:

```
target '//flow/designs/src/mempool_group/rtl:mempool_pkg.sv' is not visible from
target '//flow/designs/nangate45/mempool_group:mempool_group_synth'
```

`files("include", extra_srcs = ...)` is the same filegroup, the same glob plus the same five sibling-package headers, and also exports each source as its own label. No change to what the `include` group contains. Every other `src/` package already uses `files()`.

### DIFF
--- a/flow/designs/src/mempool_group/rtl/BUILD
+++ b/flow/designs/src/mempool_group/rtl/BUILD
@@ -1,11 +1,18 @@
-filegroup(
-    name = "include",
-    srcs = glob(["*.v", "*.sv", "*.svh"], allow_empty = True) + [
+load("//flow/designs:design.bzl", "files")
+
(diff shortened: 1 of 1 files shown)


oharboe:
Closing: opened without the maintainer's go-ahead. bazel-orfs carries this as a patch for now; an upstream PR will follow when it has settled and the timing is right.

PR #4522, opened 2026-09-08
chore(deps): bump KLayout to 0.30.12

vvbandeira:
Moves the pinned KLayout from 0.30.7 to 0.30.12, the newest release, and updates the three Ubuntu deb md5 sums to match those downloads. klayout.org publishes 0.30.12
  for Ubuntu 20, 22 and 24 and for CentOS 7, CentOS 8 and Rocky Linux 9, so every download path in the script resolves. Verified locally: 0.30.12 installs and runs headless on
  ubuntu:22.04 and ubuntu:24.04, passing a version check, a Ruby batch script, a Python batch script and a GDS write and read test.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -13,7 +13,7 @@ fi
 cd "${_script_dir}/../"
 
 # package versions
-klayoutVersion=0.30.7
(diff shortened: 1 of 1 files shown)


maliberty (approved):
Please update dev machines to match

PR #4523, opened 2026-09-09
write_rc: stream segment parasitics instead of collecting them first

oharboe:
`make write_rc` collects every wire segment's parasitics into Tcl memory before
it writes anything, and the collection is the peak of the whole flow.

`fetch_segments_rc` appends four Tcl objects per routed shape, keeps one such
list per net in an array, and only once every net has been walked does
`write_segments_rc_csv` iterate the array and emit rows. Nothing reads the data
back — it is produced once, written once, in the order it is produced.

On a large design that accumulation reaches tens of GB: several times what the
`extract_parasitics` call that produced the data needs, and more than detailed
routing. The step is already gated behind detailed route plus two extractions,
so the practical effect is that the unit-RC calibration in
`docs/tutorials/SetRC.md` is out of reach on exactly the designs whose RC values
one would want to fit rather than inherit from the platform — and per-layer
capacitance is design-specific, so those are the designs that most want their
own fit.

The fix folds the walk into `write_segments_rc_csv`, writing each row where it
is produced. The walk becomes independent of design size, so the step's peak
becomes the extraction's and the flow's peak goes back to detailed routing.

Walk order, net filters and row format are unchanged, so the CSV content is the
same. Two notes for review:

- The old code walked `getNets` twice — once to fill the array, once to emit,
  gated only on `info exists`. The single walk removes the assumption that both
  traversals return the same order.
- The duplicate-shape `error` now fires with the output file already open, so
  an aborted run leaves a partial CSV where it previously left none.

Deliberately not included here: `seen_shape_ids` is still a list, so the
membership test is a linear scan over the shapes already seen on that net. That
is a runtime concern rather than a memory one and belongs in its own change.


oharboe:
Taken, with a wider fix than closing the descriptor.

The leaked descriptor is real but the process exits on the error, so the OS reclaims it. The consequence worth fixing is the one behind it: because the writer now opens the CSV before walking, an aborted run left a truncated file where the previous collect-then-write shape left none. Closing the stream would flush that partial file rather than remove it, which makes a truncated CSV look like a complete one to whatever reads it next.

`cf31fa3` scopes the file's lifetime instead. `write_segments_rc_csv` owns the descriptor and the walk moves to `write_segments_rc_rows`; on any error the descriptor is closed, the partial CSV is deleted, and the error is rethrown with its original stack via `return -options`. That restores the old guarantee that a failed run leaves no output, and the leak goes with it.

The note about partial output in the PR description no longer applies; the behaviour now matches the pre-change code on the error path as well as the success path.

oharboe:
Correction, and I have moved this to draft: the claim in the description is wrong.

I re-ran the calibration on the large design with this change applied. The segment walk still grew past 100 GB and was killed, with the CSV about 70% written. Removing the accumulation did not move the peak meaningfully, so the memory is not the collected list — it is something in the per-segment walk itself, across ~26M `getShape`/`getTechLayer` calls and `getTotalCapacitance` over ~40M coupling segments. I attributed it from the shape of the code rather than measuring inside the loop, and that was premature.

The change still stands on its own terms — the collected list serves no purpose, the data is written once in the order produced — and the error-path fix above is worth keeping either way. But it does not deliver what the description promised, so it should not be reviewed as a fix for the memory problem until I have measured where the growth actually is.

I will report back with the per-iteration measurement, and either re-scope this PR to what it genuinely does or replace it with one that addresses the real cause.

oharboe:
Closing: the measurement says this is not the fix it was filed as.

The collected list looked like the obvious culprit, but removing it does not move the memory. The cost is in the ODB Tcl read-back itself — every distinct object surfaced to Tcl retains about a kilobyte for the life of the interpreter, and the walk surfaces three per segment. Value-returning accessors on the same objects cost nothing.

Evidence, ablations, and what it rules out: https://github.com/The-OpenROAD-Project/bazel-orfs/pull/966
The ask that follows from it: https://github.com/The-OpenROAD-Project/OpenROAD/issues/11365

The streaming change and the error-path fix here are still defensible on their own terms, but they belong with whatever consumes a bulk accessor once that exists, not as a standalone memory fix. Reopening this would only re-introduce a claim I could not support.

Apologies for the review time on a wrong diagnosis.

PR #4542, opened 2026-09-17
detail_route.tcl: call utl::metric_integer by its name

oharboe:
`utl::metric_int` is not a command. It is a unique prefix of `utl::metric_integer`, and it resolves today only because OpenSTA installs a global `namespace unknown` handler that tries an unknown name as a command abbreviation. A future `utl::metric_int*` command would turn it into an "ambiguous command" error at the route stage of every design that skips detailed routing.

Spell the command out. It is the only abbreviated command in `flow/scripts`, `flow/platforms` and `flow/util`: every command token there was checked against the command set (`info commands`, all namespaces) of the openroad binary, flagging any token that is not a command but a unique prefix of exactly one.

One line, no behaviour change. Same class of fix as OpenROAD's own 27 sites in The-OpenROAD-Project/OpenROAD (PR to follow).


oharboe:
@dsengupta0628 FYI - cleanup of .sdc DSL implementation detail leaking into OpenROAD .tcl

maliberty (approved):

PR #4546, opened 2026-09-18
place: DETAIL_PLACEMENT_ARGS reaches every detailed_placement call

oharboe:
`detail_place.tcl` passed `DETAIL_PLACEMENT_ARGS` to `detailed_placement`; the four other calls, after clock tree synthesis and after `repair_timing` in `cts.tcl`, after `repair_design` and after `repair_timing` in `global_route.tcl`, were bare. A design that picks a legalizer for the place stage, say `-use_diamond_legalizer -max_displacement {450 450}` because the default negotiation legalizer could not place its wire buffers, was legalized by the default again as soon as CTS inserted its clock buffers.

Measured on a 1.5 M-instance hierarchical asap7 design with 44 macros: `detailed_placement` in the place stage took 149 s with the diamond search; the re-legalization after `clock_tree_synthesis` in the same flow ran the negotiation legalizer over 212,700 violations for 400 iterations, 17 minutes, on the same floorplan.

All five calls now go through `detailed_placement_helper` in `util.tcl`, beside `repair_timing_helper` and `repair_design_helper`, so the argument handling lives in one place and a sixth call cannot miss it. `variables.yaml` scopes the variable to `place`, `cts` and `grt`, so a build system that filters variables by stage passes it to all three. The default is empty, so a flow that does not set it is unchanged; the variable's description already says it configures detailed placement, not one stage of it. `FlowVariables.md` and `variables.json` are regenerated with the repository's scripts.


maliberty (approved):

oharboe:
@LucasYuki @maliberty ORFS origin/master is broken, I'm sure you didn't hear it here first. This PR is approved and will go in when pr-merge goes green after origin/master is fixed.

No further action required on this PR.

